PR #21, opened 2023-09-28
Add a navigator.identity proposal

samuelgoto:
Adding another proposal that occurred to us in the past that is worth investigating


RByers (approved):
The note on how to factor in FedCM is the key question to me, but happy to have this as another alternative to discuss.

PR #37, opened 2023-10-29
Add custom schemes alternative analysis

RByers:
We've had some requests to publicly expand on the concerns we have with the use of custom URL schemes for wallet invocation. @timcappalli and @marcoscaceres, does this match your understanding as well?

@Sakurann and @msporny does this seem accurate and fairly represented to you? Improvements are appreciated, you both are more of experts in this space than I am.


custom-schemes.md
```diff
@@ -0,0 +1,41 @@
...
+
+   - Browsers generally silently ignore navigations to URL schemes with no handlers installed. This likely means a website cannot distinguish between the time it takes for a wallet to process a request and nothing happening due to the lack of any wallet app being installed. 
+   - Additionally browsers on Android and iOS often offer the user the choice of whether or not to invoke an application for a custom URL scheme navigation (even if there is only one such app). The user canceling at this stage is also indistinguishable from no valid application being installed.
+   - A browser-based API could show the user some meaningful error message when no relevant wallets are installed.
+
+5. What is the user experience on a desktop operating system?
+
+   - Most wallet applications exist only as mobile applications, and users are generally reluctant to install native applications on desktop operating systems. What fallback will flows relying on custom URL schemes provide for users of desktop operating systems?
+   - A browser-based API can potentially provide low-friction integration between a desktop browser and a wallet application running on a mobile phone.
```
msporny, line 30: ... or a web-based wallet, which provide a more seamless experience in many use cases that don't require HSMs, device-lockdown, etc. https://github.com/WICG/identity-credential/issues/33#issuecomment-1770955104

I will also note that there are now governments pointing out the need/desire for web-based wallets/solutions: https://github.com/WICG/identity-credential/issues/33#issuecomment-1784157457

custom-schemes.md
```diff
@@ -0,0 +1,41 @@
...
+
+6. What are the privacy implications of a wallet accepting custom schemes?
+
+   - In general Android works to keep the list of applications installed private from other applications. However applications registering to handle custom URL schemes effectively make their package name visible to other applications. By registering to support custom URL schemes a wallet application may be leaking PII (such as the user's likely citizenship status or state of residency) to any other application on the device.
```
msporny, line 34: The mDL wallets that implement  these schemes and that are in production today, already leak this sort of information (individual lives in a particular state).

msporny (approved):
Good initial analysis, +1 to merge; some additional commentary provided inline.

custom-schemes.md
```diff
@@ -0,0 +1,41 @@
...
+
+As a result of these concerns, Google Chrome has [proposed](https://groups.google.com/a/chromium.org/g/blink-dev/c/wcCrcMTELS0/m/ZSPxAT0LAgAJ) blocking website access to specific high-risk custom URL schemes in favor of wallets being invoked via a browser-mediated API. Anyone interested in contributing to the design of such an API is encouraged to join the W3C's [Web Incubator Community Group](https://www.w3.org/community/wicg/) and contribute to the [Identity Credential proposal](https://github.com/WICG/identity-credential/). While websites will likely always be able to communicate directly with wallet applications in some fashion, it is the group's hope that a browser-based API can be designed which is both a better option for wallets and verifiers, and which provides users with confidence in having their sensitive identity information potentially accessible to websites they visit.
```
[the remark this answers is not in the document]
RByers: Weird, works for me in a Firefox private browsing window. Maybe a temporary/intermittent outage? FWIW here's [another mirror](https://www.mail-archive.com/blink-dev@chromium.org/msg05549.html).

custom-schemes.md
RByers: But do they otherwise leak it without getting any consent from (or even indication to) the user?

custom-schemes.md
RByers: Yes, definitely! So you suggest expanding this point (or adding an additional point) to ask about integration with web-based wallets? I know I need to dig into this more, but I haven't gone over the implications too much yet. Does the browser API offer similar advantages over, say, `registerProtocolHandler` working with web wallets in response to custom scheme navigations?

timcappalli (approved):
looks great and reflects similar concerns echoed over the past few related workstreams.

custom-schemes.md
```diff
@@ -0,0 +1,41 @@
...
+   - On iOS when multiple applications are installed that register for the same custom URL scheme, the behavior is [undefined](https://stackoverflow.com/questions/13130442/multiple-apps-with-the-same-url-scheme-ios). One wallet will get invoked without any indication that other wallets were available.
+
+4. What is the user experience when no wallet app is installed or when the user cancels wallet invocation?
+
+   - Browsers generally silently ignore navigations to URL schemes with no handlers installed. This likely means a website cannot distinguish between the time it takes for a wallet to process a request and nothing happening due to the lack of any wallet app being installed. 
+   - Additionally browsers on Android and iOS often offer the user the choice of whether or not to invoke an application for a custom URL scheme navigation (even if there is only one such app). The user canceling at this stage is also indistinguishable from no valid application being installed.
+   - A browser-based API could show the user some meaningful error message when no relevant wallets are installed.
```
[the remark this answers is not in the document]
RByers: Thanks! Since I'm creating one alternatives analysis here, I can also add some links for CHAPI. I do think we have a lot to learn from it. Maybe we should have a (separate, outside this PR) discussion on the design tradeoffs in CHAPI relative to what we've been discussing for IdentityCredential?

custom-schemes.md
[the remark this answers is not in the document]
msporny: Woops... our corporate apps restrictions don't allow the use of Google Groups for some reason... it works if opened in an incognito window. Original link works as advertised, sorry for the noise.

README.md
```diff
@@ -1,13 +1,18 @@
...
 * [Identity Credential](identity-credential-proposal.md)
 * [Digital Credential](digital-credential-proposal.md)
 * [navigator.identity.get()](navigator-identity-proposal.md)
 * [Mobile Document Request API](mobile-document-request-api-proposal.md)
 
+## Alternatives being developed elsewhere
+
+* Analysis of [custom URL schemes](custom-schemes.md)
+* [CHAPI](https://chapi.io/) - a browser API [spec](https://w3c-ccg.github.io/credential-handler-api/) and polyfill focusing on wallet selection (rather than credential selection) and enabling both storage and presentation of credentials from both native and web wallets.
+
```
RByers, line 15: @msporny is this a good summary? Also happy to add (or point to) a dedicated page for CHAPI discussing it's advantages / disadvantages relative to the other options we've been discussing here.

custom-schemes.md
msporny: > Yes, definitely! So you suggest expanding this point (or adding an additional point) to ask about integration with web-based wallets? 

Yes, please. A separate bullet point would probably be best, something along the lines of:

- A browser-based API can also potentially provide a low-friction integration between a desktop browser and a website that provides an in-browser wallet experience on the same device.

> Does the browser API offer similar advantages over, say, `registerProtocolHandler` working with web wallets in response to custom scheme navigations?

Yes. We've gone to great lengths in CHAPI to provide wallet selection between both native and web-based wallets in the same UX. Excluding native wallets, or excluding web-based wallets always resulted in gnashing of teeth among the wallet providers that were left out.

Another thing to consider is requiring/forcing one to always switch devices between desktop and mobile when using the Web is a pretty annoying UX pattern. I expect always requiring mobile phone interaction to result in: "I'll do this later" or "Ugh, my phone is upstairs/downstairs/charging." issues when a web-based wallet is a more convenient experience for most of the common interactions we expect to be associated with the Verifiable Credentials ecosystem (read: not high assurance use cases).

Also keep in mind that mobile devices aren't the only secure ways of signing information -- external authenticators exist and need to be taken into consideration.

custom-schemes.md
msporny: > But do they otherwise leak it without getting any consent from (or even indication to) the user?

I was going off of this statement:

> applications registering to handle custom URL schemes effectively make their package name visible to other applications.

If that's true, then the current state-by-state mDL apps that register to handle the custom URL schemes (like `openid4vp://` as [spec'd here](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-10.1.2)) would fall into that category. I believe the attack that's being discussed here is:

1. US state mDL app registers to handle `openid4vp://`.
2. Attacker app gets all package names capable of handling `openid4vp://`.
3. Attacker now knows that the victim is most likely a resident of a specific US state.

That said, I don't what avenue one would use to do that on Android (I'm not a mobile phone developer), so it may be that understanding which app registered doesn't happen until the individual clicks on the app in the protocol scheme handler app selection sheet?

What I was trying to provide here was a concrete example of this happening today. The text makes it sound like registering to handle a custom URL scheme on Android bypasses the "keep list of installed applications private from other applications" protection. If so, it would help to provide a concrete example, if not, you might want to reword the above as it makes it seem like it's possible to bypass that protection today.

README.md
msporny: Yes, good summary, thank you. https://chapi.io/ is the right site to point to.

I'm not sure we need to speak to it's benefits vs. other options as CHAPI's natural progression is (ideally) the Identity Credential API (IC API). If IC API becomes broadly deployed, then CHAPI (eventually) goes away because we got to where we wanted to go -- a native browser API to store/transmit digital credentials). IOW, the ultimate benefits of CHAPI implemented in the browser should align with the benefits of the Identity Credential API implemented in the browser (if we're able to address the remaining sticking points, of which there are few). These "CHAPI benefits" include:

* Direct communication between website and wallet, reducing phishing risk (IC API provides this)
* Reduction in protocol complexity over OID4 (no need for OpenID CSRF protections, PKCE, PIN-approval, HTTP-backend APIs, etc.)
* Query language and protocol agnostic (hopefully IC API goes this route)
* Supports both web-based and native digital wallets (hopefully IC API provides web-based wallet support)
* Keeps UX "in the flow" (if you're on your desktop browser, you don't need to switch to your phone to complete all flows) (IC API provides this)

We are strongly considering upgrading CHAPI to do credential selection if the Identity Credential API goes down this path. Most likely by using the same mechanism as provided by the Identity Credential API (if possible given the polyfill constraints). We want to make sure we get convergence on the same UX and devX -- after considering if there is going to be a backlash from the community over wallet selection vs. credential selection. We might provide credential selection as an (experimental) option that individuals can set: "Don't automatically find credentials for me across wallets [stronger privacy]" vs. "Automatically find credentials for me in my wallet [weaker privacy]" ... and this would be "stronger/weaker privacy" in relation to the polyfill website, which would be able to run code (just like the browser would) to see if a certain wallet had credentials in it.

custom-schemes.md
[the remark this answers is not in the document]
msporny: See comments in https://github.com/WICG/identity-credential/pull/37#discussion_r1375504345 -- CHAPI and the Identity Credentials API are far more aligned in philosophy and direction than most of the other solutions, in a way that they're already roughly 75%-ish converged, AFAICT.

@samuelgoto has done an excellent job deep-diving on CHAPI and I believe understands it at enough depth to provide better input on IC API's direction than any text could provide. That said, it might not be that difficult to identify the gaps between IC API and CHAPI (I expect that they are few in number).

README.md
```diff
@@ -1,13 +1,18 @@
 # Proposals related to identity credentials
 
-This repo is being used to work on, and hopefully converge, proposals
-related to identity credentials on the web:
+This repo is being used to work on, and hopefully converge on, proposals
+related to the innvocation of digital identity wallets from the web:
```
dlongley, line 4:
```suggestion
related to the invocation of digital identity wallets from the web:
```

custom-schemes.md
dlongley: Another point perhaps worth mentioning:

A browser API could potentially keep private (from the relying verifier / issuer) the specific wallet the user selects to complete a request in ways that may be impossible or challenging when using other mechanisms (e.g., a custom URL scheme that includes an invitation to use a protocol that requires a wallet to communicate as a client with the relying verifier / issuer).

custom-schemes.md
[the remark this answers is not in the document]
TallTed:
```suggestion
Due to these concerns, Google Chrome has [proposed](https://groups.google.com/a/chromium.org/g/blink-dev/c/wcCrcMTELS0/m/ZSPxAT0LAgAJ) blocking website access to specific high-risk custom URL schemes, and instead invoking wallets via a browser-mediated API. Anyone interested in contributing to the design of such an API is encouraged to join the W3C's [Web Incubator Community Group](https://www.w3.org/community/wicg/) and contribute to the [Identity Credential proposal](https://github.com/WICG/identity-credential/). While websites will likely always be able to communicate directly with wallet applications in some fashion, it is the group's hope that a browser-based API can be designed which is a better option for wallets and verifiers, and which also provides users with confidence in having their sensitive identity information potentially exposed to websites they visit.
```

TallTed (changes requested):
(editorial) Some language tweaks to improve clarity.

custom-schemes.md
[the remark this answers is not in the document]
RByers: Thanks Manu! Yes I've heard a bunch about CHAPI from Sam and I know it's shaped his thinking on design. I'm personally comfortable that we've done enough here to ensure CHAPI is being leveraged, but I was worried that we perhaps hadn't communicated that well enough to folks who had invested a lot of their time and energy into CHAPI 😊. So perhaps just citing CHAPI in the alternatives section of the README is sufficient?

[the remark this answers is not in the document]
msporny: Thanks for the consideration, yes, citing CHAPI (and it's broad alignment w/ Identity Credentials API) in the alternatives section would be sufficient.

Not for this PR, but we might also consider having a matrix comparing the various approaches/protocols/drawbacks. I know those can be difficult to create, but they can also be helpful to get people up to speed w/ what certain solutions do/don't do. I believe it would be fairly easy to do a comparison between IC API, CHAPI, and OID4VCI/VP. Just a thought, not asking anyone to do this right now.

dlongley:
Note that this thread of discussion (from which more text may be potentially added to this PR) was auto-resolved because of a suggestion at the end of it:

https://github.com/WICG/identity-credential/pull/37#discussion_r1375474010

custom-schemes.md
```diff
@@ -0,0 +1,41 @@
...
+
+1. Can wallets reliably determine their invoker?
+
+   - Android and iOS offer no facility to determine the web origin which triggered a navigation to a custom URL scheme. 
+   - This seems to increase the opportunity for attacks such as replay attacks, man-in-the-middle attacks and phishing. It also seems to limit the opportunity for abuse detection and prevention.
+   - A browser API can securely pass the requesting origin on to the wallet so that the wallet can know with confidence what origin the user was on when the request was initiated.
+
+2. Can wallets limit requests to secure contexts?
+
+   - If a user uses a website via http (not https) and it invokes a custom scheme, it's possible that an attacker with access to the network could eavesdrop on or even modify the request (such as by replacing the reader public key with their own).
```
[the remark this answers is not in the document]
RByers: I think it's actually the reader (i.e. verifier's) public key that matters here. I'm suggesting a MITM attack on the response encryption. Would "verifier's public key" be more clear than "reader public key" (which maybe an mdoc term)?

RByers:
> Note that this thread of discussion (from which more text may be potentially added to this PR) was auto-resolved because of a suggestion at the end of it:
> 
> [#37 (comment)](https://github.com/WICG/identity-credential/pull/37#discussion_r1375474010)

Ah thank you! Unresolved.

custom-schemes.md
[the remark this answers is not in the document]
RByers: Yep sounds good. Thanks!

custom-schemes.md
RByers: > A browser-based API can also potentially provide a low-friction integration between a desktop browser and a website that provides an in-browser wallet experience on the same device.

Added along with the existing bullet.

> Yes. We've gone to great lengths in CHAPI to provide wallet selection between both native and web-based wallets in the same UX. Excluding native wallets, or excluding web-based wallets always resulted in gnashing of teeth among the wallet providers that were left out, and at the end of the day, when control is handed over to the wallets, the UX looks/feels very much the same (get consent for the sharing of certain data).

Thanks, filed #38 to track.

> Also keep in mind that mobile devices aren't the only secure ways of signing information -- Web apps + external authenticators exist and should be taken into consideration. We perform all sorts of really high assurance use cases with that tooling (or less capable tooling) available to us today.

Yeah it's a good point. Since that would still need some wallet-defined UI I think that's probably just an extension of the web wallet case, right? A web wallet that happens to use some variant of WebAuthn (whether internal or external authenticator) to provide a strong device-bound assertion?

custom-schemes.md
```diff
@@ -0,0 +1,41 @@
...
+5. What is the user experience on a desktop operating system?
+
+   - Most wallet applications exist only as mobile applications, and users are generally reluctant if not unable (due to enterprise security settings, for instance) to install native applications on desktop operating systems. What fallback will flows relying on custom URL schemes provide for users of desktop operating systems?
+   - A browser-based API can potentially provide low-friction integration between a desktop browser and a wallet application running on a mobile phone.
+
+6. What are the privacy implications of a wallet accepting custom schemes?
+
+   - In general, Android works to keep the list of installed applications private from other applications. However, applications registering to handle custom URL schemes effectively make their package name visible to other applications. By registering to support custom URL schemes, a wallet application may be leaking PII (such as the user's likely citizenship status or state of residency) to any other application on the device.
+   - A purpose-built OS API for wallet invocation (on top of which a browser API would be built) can avoid any such information leakage.
```
dlongley, line 35: Converting one of my comments into a suggestion here:

```suggestion
   - A purpose-built OS API for wallet invocation (on top of which a browser API would be built) can avoid any such information leakage.
   - A browser API could potentially keep private (from the relying verifier or issuer) the specific wallet the user selects to complete a request. This may be impossible or challenging when using other mechanisms (e.g., a custom URL scheme that includes an invitation to use a protocol that requires a wallet to directly communicate with the relying verifier or issuer).   
```

custom-schemes.md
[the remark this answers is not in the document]
TallTed: Thanks, @RByers. Fixed!

custom-schemes.md
RByers: > I was going off of this statement:

Ah sorry, my bad.

>  If so, it would help to provide a concrete example (state-issued mDL apps), if not, you might want to reword the above as it makes it seem like it's possible to bypass that protection today.

I see. I don't actually know if any apps in production are registering for the mdoc: scheme today. We're monitoring mdoc: usage in Chrome and it's currently so small that I assume it's just a few developers doing testing. I've expanded on this and added a link for more details. It's a little complex, but my understanding is that yes registering to handle any intents (including custom URL schemes) makes your package name discoverable by other apps.

> A browser API could potentially keep private (from the relying verifier / issuer) the specific wallet the user selects to complete a request. This may be impossible or challenging when using other mechanisms (e.g., a custom URL scheme that includes an invitation to use a protocol that requires a wallet to directly communicate as a client with the relying verifier / issuer).

Yeah I was wondering about saying something like that, but I didn't see how it would be materially different. Both setups involve some arbitrary data transfer from the wallet back to the verifier. That protocol could be designed to either rely on wallet keys (or other form of wallet identification) or not, right?

README.md
RByers: Nice, makes perfect sense. Thanks Manu!

custom-schemes.md
RByers: Yeah I was wondering about saying something like that, but I didn't see how it would be materially different. Both setups involve some arbitrary data transfer protocol from the wallet back to the verifier. That protocol could be designed to either rely on wallet keys (or other form of wallet identification) or not, right? Direct communication isn't, I don't think, any more or less identifiable (just like web servers can't necessarily identify a browser unless the browser advertises itself somehow either directly or indirectly via it's precise behavior).

RByers:
Thanks for all the input folks! I'm going to merge this now in preparation for our call tomorrow, but am happy to continue to land improvements for any further comments here.

custom-schemes.md
msporny: > my understanding is that yes registering to handle any intents (including custom URL schemes) makes your package name discoverable by other apps.

Right, and if that's the case, let's give people a concrete example: State-provided mDL apps identify that the individual resides in a particular state w/o the individual giving consent to releasing that information. This is something that CHAPI does not expose to the Verifier until after consent is provided by the Holder to release the information requested by the Verifier.

> > A browser API could potentially keep private (from the relying verifier / issuer) the specific wallet the user selects to complete a request. This may be impossible or challenging when using other mechanisms (e.g., a custom URL scheme that includes an invitation to use a protocol that requires a wallet to directly communicate as a client with the relying verifier / issuer).
> 
> Yeah I was wondering about saying something like that, but I didn't see how it would be materially different. Both setups involve some arbitrary data transfer from the wallet back to the verifier. 

The point I was trying to make was that a browser API could keep which wallets are available/selected hidden from the Verifier (and that would probably be ideal, even after consent to share a digital credential from the Holder). 

> That protocol could be designed to either rely on wallet keys (or other form of wallet identification) or not, right?

Yes, and I'll suggest that /one/ of the protocols is to just deliver the Presentation to the Verifier through the browser API w/o exposing anything about the wallet that performed the task. This optimizes for privacy in the default case. That is not to say that the Verifier couldn't also ask: "I want an attestation on which wallet you're using", but that has a bad "market competition" smell. There are some entities that believe that wallets must always identify themselves, which some in the ecosystem believe is an anti-pattern. The alternative is to perform capability checking on the wallet. In an ideal world, the Verifier should be asking:

* Provide a digital credential to me that verifies that the cryptographic key you are using to digitally sign the presentation is housed in an HSM.
* Provide a liveness credential that verifies that your face matches the face on the identity document you are supplying.
* Provide a captcha credential that proves that you solved this captcha I've provided that is specific to this credential exchange.

and not: "Provide a credential for the BigBrand wallet you are using so I can determine if I want to interact with your wallet because I have vetted it for usage for my use case." (Doing "wallet sniffing" feels quite similar to "browser sniffing", which we've tried desperately to move away from over the past two decades on the Web Platform).

In short, wallet identification feels like an anti-pattern and we should instead follow the path that the browsers have taken wrt. user agent capability detection.

custom-schemes.md
dlongley: > Yeah I was wondering about saying something like that, but I didn't see how it would be materially different. Both setups involve some arbitrary data transfer protocol from the wallet back to the verifier. That protocol could be designed to either rely on wallet keys (or other form of wallet identification) or not, right? Direct communication isn't, I don't think, any more or less identifiable (just like web servers can't necessarily identify a browser unless the browser advertises itself somehow either directly or indirectly via it's precise behavior).

Yeah, I think that's right, but I was thinking about this less from the perspective of different protocol designers that equally consider this privacy characteristic -- and more from the perspective of wallet providers that are looking to implement existing protocols on offer.

I think there's an advantage to say that a browser API could provide this sort of privacy "by default", simply because of the architectural pattern that places the browser as a mediator between the RP and the wallet. Other protocols on offer that lack this architecture, especially those that do not consider this sort of privacy in their design, may incidentally have elements that make it an active challenge to try to provide it.

Some of this has come up before with protocols that require direct communication from Web wallet backend services because RPs are not CORS-enabled, leading, perhaps, to more easily identifying the wallet services chosen by the user -- or more challenges to wallet providers to help avoid this. These kinds of concerns and challenges can potentially be avoided entirely by wallet providers based on a browser-as-mediator architecture.

But I take the point that other protocols can attempt to build in this sort of privacy as well if they want to.

custom-schemes.md
```diff
@@ -0,0 +1,41 @@
+# Concerns with custom schemes for identity presentment
+
+Existing specifications like [OpenID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-wallet-invocation) and [ISO 18013-7](https://www.iso.org/standard/82772.html) currently define the use of custom URL schemes (eg. `openid4vp:` and `mdoc:`) to enable verifier websites to invoke holder (wallet) applications. This approach works in some scenarios, but has a number of problems we aim to solve by designing a browser API which can replace the use of custom URL schemes in these specifications. This DRAFT document aims to outline the questions and concerns with the use of custom URL schemes for identity presentment in order to provide more detail on potential benefits of a dedicated browser API.
+
+1. Can wallets reliably determine their invoker?
+
+   - Android and iOS offer no facility to determine the web origin which triggered a navigation to a custom URL scheme. 
+   - This seems to increase the opportunity for attacks such as replay attacks, man-in-the-middle attacks and phishing. It also seems to limit the opportunity for abuse detection and prevention.
```
Sakurann, line 8: android and iOS might not allow it, but that's why protocols like OID4VP/18013-7 talk about client_id_schemes and verifier authentication which allows the wallet to reliably determine the invoker.

Sakurann:
what is the merge policy? the PR got merged in 3 days - I had it on my todo list to review assuming I had more time than 3 days...

timcappalli:
> what is the merge policy? the PR got merged in 3 days - I had it on my todo list to review assuming I had more time than 3 days...

Explainer-type documents are are the discretion of the author. We are using GitHub instead of Google Docs to keep everything together.

RByers:
Hey @Sakurann, sorry I missed your comment after the PR landed (darn github notification overload). The group continued to iterate on this based on additional feedback after it landed (eg. see #41 and #45). I'm happy to either review a PR of additional suggestions or put one up myself based on your comment here?

PR #41, opened 2023-11-01
custom-schemes.md: additional concerns

hober:
1. Add two additional points.
2. Use headings instead of numbers so that nesting is more straightforward.
3. Separate the conclusion from the list of items.

### DIFF
--- a/custom-schemes.md
+++ b/custom-schemes.md
@@ -12,7 +12,7 @@ Existing specifications like [OpenID4VP](https://openid.net/specs/openid-4-verif
 
   - Does the page give the wallet a URL to navigate back to as part of the custom scheme URL it uses to launch the wallet?
 
-If so, that's a problem because it won't go back to the same tab. Apps don't have the ability to return information to the same page that invoked them after a custom scheme invocation; they can just open a new page in a new tab.
+  - If so, that's a problem because it won't go back to the same tab. Apps don't have the ability to return information to the same page that invoked them after a custom scheme invocation; they can just open a new page in a new tab.
 
 ## Can wallets limit requests to secure contexts?
 
--- a/custom-schemes.md
+++ b/custom-schemes.md
@@ -10,7 +10,7 @@ Existing specifications like [OpenID4VP](https://openid.net/specs/openid-4-verif
    
 ## How are wallet apps supposed to pass info back to the invoking page?
 
-Does the page give the wallet a URL to navigate back to as part of the custom scheme URL it uses to launch the wallet?
+  - Does the page give the wallet a URL to navigate back to as part of the custom scheme URL it uses to launch the wallet?
 
 If so, that's a problem because it won't go back to the same tab. Apps don't have the ability to return information to the same page that invoked them after a custom scheme invocation; they can just open a new page in a new tab.
 
--- a/custom-schemes.md
+++ b/custom-schemes.md
@@ -39,7 +39,7 @@ Existing specifications like [OpenID4VP](https://openid.net/specs/openid-4-verif
 
 ### Friction and app switching
 
-Opening a custom URL scheme will switch apps, which is a high friction user flow. A browser API could present relevant UI (even from third party wallet apps) in a sheet over the webpage, which is a less disruptive flow.
+  - Opening a custom URL scheme will switch apps, which is a high friction user flow. A browser API could present relevant UI (even from third party wallet apps) in a sheet over the webpage, which is a less disruptive flow.
 
 ## What are the privacy implications of a wallet accepting custom schemes?
 
--- a/custom-schemes.md
+++ b/custom-schemes.md
@@ -2,40 +2,54 @@
 
 Existing specifications like [OpenID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-wallet-invocation) and [ISO 18013-7](https://www.iso.org/standard/82772.html) currently define the use of custom URL schemes (eg. `openid4vp:` and `mdoc:`) to enable verifier websites to invoke holder (wallet) applications. This approach works in some scenarios, but has a number of problems we aim to solve by designing a browser API which can replace the use of custom URL schemes in these specifications. This DRAFT document aims to outline the questions and concerns with the use of custom URL schemes for identity presentment in order to provide more detail on potential benefits of a dedicated browser API.
 
-1. Can wallets reliably determine their invoker?
+## Can wallets reliably determine their invoker?
 
-   - Android and iOS offer no facility to determine the web origin which triggered a navigation to a custom URL scheme. 
+   - Android and iOS offer no facility to determine the web origin which triggered a navigation to a custom URL scheme.
    - This seems to increase the opportunity for attacks such as replay attacks, man-in-the-middle attacks and phishing. It also seems to limit the opportunity for abuse detection and prevention.
    - A browser API can securely pass the requesting origin on to the wallet so that the wallet can know with confidence what origin the user was on when the request was initiated.
+   
+## How are wallet apps supposed to pass info back to the invoking page?
 
-2. Can wallets limit requests to secure contexts?
+Does the page give the wallet a URL to navigate back to as part of the custom scheme URL it uses to launch the wallet?
 
-   - If a user uses a website via http (not https) and it invokes a custom scheme, it's possible that an attacker with access to the network could eavesdrop on or even modify the request (such as by replacing the verifier's public key with their own). 
-   - A browser API could take the position that it will function only within the context of a secure context (https).
+If so, that's a problem because it won't go back to the same tab. Apps don't have the ability to return information to the same page that invoked them after a custom scheme invocation; they can just open a new page in a new tab.
 
-3. What is the user experience when multiple wallet apps are installed?
+## Can wallets limit requests to secure contexts?
+
+   - If a user uses a website via http (not https) and it invokes a custom scheme, it's possible that an attacker with access to the network could eavesdrop on or even modify the request (such as by replacing the verifier's public key with their own).
+   - A browser API could (and almost certainly should) take the position that it will function only within the context of a secure context (https).
+
+## User experience concerns
+
+### What is the user experience when multiple wallet apps are installed?
 
    - When multiple Android applications register for the same custom URL scheme, the operating system presents the user with a picker every time that scheme is invoked. Even if the verifier provides the user with appropriate context, a user may have trouble remembering which application has which credentials and so be more likely to cancel or fail the request.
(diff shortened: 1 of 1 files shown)


custom-schemes.md
```diff
@@ -2,40 +2,54 @@
...
 
-   - Android and iOS offer no facility to determine the web origin which triggered a navigation to a custom URL scheme. 
+   - Android and iOS offer no facility to determine the web origin which triggered a navigation to a custom URL scheme.
    - This seems to increase the opportunity for attacks such as replay attacks, man-in-the-middle attacks and phishing. It also seems to limit the opportunity for abuse detection and prevention.
    - A browser API can securely pass the requesting origin on to the wallet so that the wallet can know with confidence what origin the user was on when the request was initiated.
+   
+## How are wallet apps supposed to pass info back to the invoking page?
 
-2. Can wallets limit requests to secure contexts?
+Does the page give the wallet a URL to navigate back to as part of the custom scheme URL it uses to launch the wallet?
```
RByers, line 13:
```suggestion
  - Does the page give the wallet a URL to navigate back to as part of the custom scheme URL it uses to launch the wallet?
```

custom-schemes.md
```diff
@@ -2,40 +2,54 @@
...
 
-   - If a user uses a website via http (not https) and it invokes a custom scheme, it's possible that an attacker with access to the network could eavesdrop on or even modify the request (such as by replacing the verifier's public key with their own). 
-   - A browser API could take the position that it will function only within the context of a secure context (https).
+If so, that's a problem because it won't go back to the same tab. Apps don't have the ability to return information to the same page that invoked them after a custom scheme invocation; they can just open a new page in a new tab.
```
RByers, line 15:
```suggestion
  - If so, that's a problem because it won't go back to the same tab. Apps don't have the ability to return information to the same page that invoked them after a custom scheme invocation; they can just open a new page in a new tab.
```

custom-schemes.md
```diff
@@ -2,40 +2,54 @@
...
-5. What is the user experience on a desktop operating system?
+### What is the user experience on a desktop operating system?
 
    - Most wallet applications exist only as mobile applications, and users are generally reluctant if not unable (due to enterprise security settings, for instance) to install native applications on desktop operating systems. What fallback will flows relying on custom URL schemes provide for users of desktop operating systems?
    - A browser-based API can potentially provide low-friction integration between a desktop browser and a wallet application running on a mobile phone or a website that provides an in-browser wallet experience on the same device.
 
-6. What are the privacy implications of a wallet accepting custom schemes?
+### Friction and app switching
+
+Opening a custom URL scheme will switch apps, which is a high friction user flow. A browser API could present relevant UI (even from third party wallet apps) in a sheet over the webpage, which is a less disruptive flow.
```
RByers, line 42:
```suggestion
  - Opening a custom URL scheme will switch apps, which is a high friction user flow. A browser API could present relevant UI (even from third party wallet apps) in a sheet over the webpage, which is a less disruptive flow.
```

RByers (changes requested):
Thanks! Reformatting looks good, just suggest using bullets consistently.

custom-schemes.md
```diff
@@ -2,40 +2,54 @@
...
 
-   - If a user uses a website via http (not https) and it invokes a custom scheme, it's possible that an attacker with access to the network could eavesdrop on or even modify the request (such as by replacing the verifier's public key with their own). 
-   - A browser API could take the position that it will function only within the context of a secure context (https).
+If so, that's a problem because it won't go back to the same tab. Apps don't have the ability to return information to the same page that invoked them after a custom scheme invocation; they can just open a new page in a new tab.
 
-3. What is the user experience when multiple wallet apps are installed?
+## Can wallets limit requests to secure contexts?
+
+   - If a user uses a website via http (not https) and it invokes a custom scheme, it's possible that an attacker with access to the network could eavesdrop on or even modify the request (such as by replacing the verifier's public key with their own).
+   - A browser API could (and almost certainly should) take the position that it will function only within the context of a secure context (https).
```
danielfett, line 20: That's more of a problem of the website and ecosystems can take precautions against this (rules, checks for HTTPS redirections, HSTS, etc.).

custom-schemes.md
```diff
@@ -2,40 +2,54 @@
...
    - A browser-based API can potentially provide low-friction integration between a desktop browser and a wallet application running on a mobile phone or a website that provides an in-browser wallet experience on the same device.
 
-6. What are the privacy implications of a wallet accepting custom schemes?
+### Friction and app switching
+
+Opening a custom URL scheme will switch apps, which is a high friction user flow. A browser API could present relevant UI (even from third party wallet apps) in a sheet over the webpage, which is a less disruptive flow.
+
+## What are the privacy implications of a wallet accepting custom schemes?
 
    - [Android aims](https://developer.android.com/training/package-visibility) to keep the list of installed applications private from other applications, but this capability is reduced for applications which handle arbitrary intents like custom URL schemes. By registering to support custom URL schemes, a wallet application may effectively be leaking PII (such as the user's likely citizenship status or state of residency) to other applications on the device.
```
danielfett, line 46: The leak happens only when the website learns whether the app was opened or not (or am I missing anything?). Therefore, having e.g., a user interaction before the wallet calls back to the website (via a backchannel) would help to reduce the impact of this privacy leak. (It cannot be exploited in large scale any longer.)

tantek (approved):
+1 on the additions and headings.

custom-schemes.md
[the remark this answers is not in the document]
msporny: Well, for some definition of "definitely work". :) It is possible to end in a "successful state", but that ignores the UX challenges of what happens when things go wrong. In the best case (on a same-device flow), presuming that the protocol has some sort of "navigate to this URL when you're done" feature, you're left with a "dead tab" in the browser -- something that the individual will never go back to and have to clean up later.

One of the design goals of [CHAPI](https://chapi.io/) was to not "break the flow" in the existing web page (and not to leave "dead tab"s around). This is also aligned w/ the design goals for Payment Request (and Payment Handler). That is, you want the browser API to always tell you when the interaction is done (success or failure) and return you to the exact place you were on the website that invoked the interaction in the first place. I expect FedCM has the same (ideal world) requirements.

There are at least three problems w/ invoking a native app or a web-app (using what's available today) in a separate window:

1. In a same-device flow, when the flow completes successfully, navigating back to the same tab/window that the individual was in (in the browser) is not possible. In the best case, you have to open a new tab/window, leaving "dead tab"s around. 
2. In a same device-flow, the UX is not ideal -- popping windows up (because you need to be in first party mode in many cases, and sometimes you need to log in -- which pops up more windows if you are using an external IdP) can cause the user to feel as if something bad is happening - "User: WTF is going on here? This doesn't feel right *closes window, breaking the interaction*").
3. In a multi-device or same-device flow, the user is never navigated back to the originating page when there is an error or the interaction is canceled. They're left in the "invoked app" and have to manually switch back to their browser session, which gets no indication that things went wrong and therefore "hangs" for an unacceptably long time -- 5 minutes, 10 minutes... how long should a site give an individual to complete an out-of-band flow?

It's not just a developer ergonomics issue, it's a UX issue, both for same-device and multi-device flows.

custom-schemes.md
msporny: Leaving security up to "ecosystems" tends to result in more security holes than we'd like to see. The mere fact that you can invoke an OID4VP flow via an HTTP website is a security issue, and no amount of lecturing web developers will prevent that from happening. A browser API would help guide developers towards only operating in HTTPS mode, universally, which is a good thing.

custom-schemes.md
msporny: @danielfett see this thread for background on this statement: https://github.com/WICG/identity-credential/pull/37#discussion_r1375474010

custom-schemes.md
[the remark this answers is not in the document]
RByers: > In a same-device flow, when the flow completes successfully, navigating back to the same tab/window that the individual was in (in the browser) is not possible. In the best case, you have to open a new tab/window, leaving "dead tab"s around.

Definitely agree not having a way to navigate back to the browser app is a problem that we should add to this list.

I had assumed the "best case" wasn't that the wallet would open a new browser window (which browser?), but that it would send a message to a server which the existing web page was waiting on and the user would manually switch back to their last application and see it complete. But there are problems with that as well (eg. under memory pressure the browser may have been killed).

The jarring UX of a full screen app switch is also a problem I can add to the list. I think I've already captured your "3".

custom-schemes.md
RByers: I'm OK with relying on the ecosystem but only if we're confident the incentives are aligned with the agency. I think it's issuers who have the most incentive for security and so they're going to want to certify wallets. Most wallets will probably want to ensure they're invoked securely (even if only to earn the trust of the credential issuers). But verifiers may not have as much incentive, they're more incentivized to make adoption as easy as possible. So if wallets can't impose security requirements on verifiers then I think we have a break in the accountability chain and should expect security for users to suffer. Having the platform in the middle seems to fix this because the platform is incentivized to earn the trust of wallets/issuers (otherwise they'll go around the platform API), and so will impose security constraints on verifiers which increase trust of the wallets and issuers.

custom-schemes.md
RByers: Yeah that's not the problem. It's that registering an intent filter on Android makes your Android package name discoverable by other Android apps (silently and at any time) which say they need to query the system for apps satisfying that intent filter.

custom-schemes.md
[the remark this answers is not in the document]
msporny: > the user would manually switch back to their last application and see it complete

This is a tall UX ask of the user. Some of the testing we performed showed some individuals not understanding what the next step in the process was... so, after sending the presentation, they just sat there looking at the wallet, not knowing what to do (because some of them thought they were still in the browser and others didn't know about the usage pattern where you complete something in an app and then switch back to a browser -- it's a rare usage pattern on mobile phones to bounce from browser->app->browser for the same logical flow).

Some were expecting wallet to tell them the next step, and in some cases, the wallet just said "transaction complete!" and then nothing. This could be mitigated by the wallet showing a UI telling them to switch back to their browser (but that feels like a UX failure). To get a consistent experience across all wallets for that UX flow, you'd have to get all wallets to implement that (hacky) UX hint. We've also found that in some cases, once the individual is in the wallet, the wallet software wants them to do other things and keep them there ("Get another coupon!" ... "Add a credit card!" ... "Wait, update your password!" ... "Turn on MFA!")

So yes:

1. Full-screen is jarring, a modal that allows the UX flow to be performed and then return the individual back to the page (with a concrete transaction result for the website to process) is better.
2. Giving the wallet some hint about where to navigate the user back to (if we don't do the modal thing above) might also be helpful.

We might consider something like the "peek" feature in Arc as the ideal UX pattern to follow here: 

https://www.reddit.com/r/ArcBrowser/comments/113ui8u/peek_by_arc_the_best_tab_never_made/

PR #45, opened 2023-11-03
Tweak custom-schemes

RByers:
Tweak and expand the issue of passing data back to the application based on [comments from @msporny](https://github.com/WICG/identity-credential/pull/41#discussion_r1380084601)

### DIFF
--- a/custom-schemes.md
+++ b/custom-schemes.md
@@ -15,11 +15,6 @@ Existing specifications like [OpenID4VP](https://openid.net/specs/openid-4-verif
 
 ## User experience concerns
 
-### What is the user experience of the wallet operation?
-
-   - Launching an entirely new application can be jarring for users, taking them out of the flow of the journey they were trying to complete on a website. This risks abandonment, and so is a disincentive for verifiers to invoke the flow.
-   - A browser API could integrate with the operating system to provide a less jarring wallet overlay on top of the web page, keeping the user feeling in flow with the site's experience.
-
 ### What is the user experience when the wallet has completed?
 
    - After launching an application via a custom URL scheme, mobile operating systems don't provide a mechanism for switching back to the invoking application / browser tab.
@@ -45,7 +40,8 @@ Existing specifications like [OpenID4VP](https://openid.net/specs/openid-4-verif
 
 ### Friction and app switching
 
-   - Opening a custom URL scheme will switch apps, which is a high friction user flow. A browser API could present relevant UI (even from third party wallet apps) in a sheet over the webpage, which is a less disruptive flow.
+   - Opening a custom URL scheme will launch an entirely new application, which can be jarring for users, taking them out of the flow of the journey they were trying to complete on a website. This risks abandonment, and so is a disincentive for verifiers to invoke the flow.
+   - A browser API could integrate with the operating system to provide a less jarring wallet overlay on top of the web page, keeping the user feeling in flow with the site's experience.
 
 ## What are the privacy implications of a wallet accepting custom schemes?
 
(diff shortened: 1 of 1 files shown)


RByers:
FYI @msporny

custom-schemes.md
```diff
@@ -8,19 +8,25 @@ Existing specifications like [OpenID4VP](https://openid.net/specs/openid-4-verif
...
 
    - If a user uses a website via http (not https) and it invokes a custom scheme, it's possible that an attacker with access to the network could eavesdrop on or even modify the request (such as by replacing the verifier's public key with their own).
    - A browser API could (and almost certainly should) take the position that it will function only within the context of a secure context (https).
 
 ## User experience concerns
 
+### What is the user experience of the wallet operation?
+
+   - Launching an entirely new application can be jarring for users, taking them out of the flow of the journey they were trying to complete on a website. This risks abandonment, and so is a disincentive for verifiers to invoke the flow.
+   - A browser API could integrate with the operating system to provide a less jarring wallet overlay on top of the web page, keeping the user feeling in flow with the site's experience.
```
msporny, line 21: Yes! Exactly.

What we might be looking for here is something like the "peek" feature in the Arc browser:

https://www.reddit.com/r/ArcBrowser/comments/113ui8u/peek_by_arc_the_best_tab_never_made/

In fact, that would allow a web-based wallet to pop up in first party mode /and/ a native app to pop up using the same UI (a good thing). The only thing missing would be the "cross-tab" communication between the initiating website and the initiated digital wallet.

If we're looking for the "minimal thing that could be done to enable the ecosystem", this might be it. To be clear, you would still do "wallet/credential selection" in this "peek overlay" (to hide the interaction from all wallets until the user makes a choice)... but then, you don't do a jarring full-screen switch -- you just show the wallet content in the "peek overlay" and once the interaction is done, you get rid of the overlay and give the website the result (resolve the Promise w/ the data from the wallet).

msporny (approved):
Agree that these are additional considerations for how a wallet is invoked and for how the wallet provides control back to the browser.

custom-schemes.md
```diff
@@ -8,19 +8,25 @@ Existing specifications like [OpenID4VP](https://openid.net/specs/openid-4-verif
...
-  - If so, that's a problem because it won't go back to the same tab. Apps don't have the ability to return information to the same page that invoked them after a custom scheme invocation; they can just open a new page in a new tab.
-
 ## Can wallets limit requests to secure contexts?
 
    - If a user uses a website via http (not https) and it invokes a custom scheme, it's possible that an attacker with access to the network could eavesdrop on or even modify the request (such as by replacing the verifier's public key with their own).
    - A browser API could (and almost certainly should) take the position that it will function only within the context of a secure context (https).
 
 ## User experience concerns
 
+### What is the user experience of the wallet operation?
```
hober, line 18: Is this redundant with the "Friction and app switching" bullet point?

hober (approved):
Looks good to me, though I suspect we could make things more succinct (see other comment).

custom-schemes.md
RByers: Yep. FWIW I've had a "peak overlay" UI experiment [on my back-burner](https://bugs.chromium.org/p/chromium/issues/detail?id=1364321) for awhile. It seemed to me like it might be possible to make "window.open(...,"popup") just use this by default on mobile. It turned out to be harder to implement than I thought (lots of web platform APIs aren't hooked up in this preview UI we already have in Chrome), but I think this is something Chrome could prioritize if there was sufficient demand.

custom-schemes.md
RByers: Oh whoops, good point. Thanks, I merged them.

PR #51, opened 2023-12-05
Add clarity for call schedule

timcappalli:
Explicitly list out dates for calls


timcappalli:
@RByers I lost access to direct committing in the past few weeks or so for some reason.

Issue #56, opened 2024-01-24
Using CryptoKey for public keys - what are the restrictions?


marcoscaceres:
Right now, we had proposed passing around DOM strings for public keys... However, `CryptoKey` serves as a nice primitive we already have in the platform from Web Crypto. 

I think we can reuse the `crypto.subtle.subtle.importKey()` method to get the `CryptoKey`, which we can then pass into `requestIdentity()` as part of the `IdentityRequestOptions`.

OR13:
Does this issue imply mandatory support for `jwk`, `pkcs8`, and `raw` ?

Does this forbid certificates?

marcoscaceres:
I guess it would be whatever Web Crypto mandates for formats, as it's the one providing the API to create `CryptoKey`. We can further restrict that in our spec. 

However, new format support would need to get added to Web Crypto so we could use them (if needed).

jogu:
Can someone confirm if using Cryptokey allows us to preserve potentially important information like the 'kid' from a JWK please? From a brief read of the API it looks like that information would be lost.

marcoscaceres:
Note that nothing need be "lost"... we can just amend Web Crypto if we need to and something is important to us. 

What uses `kid` and why does it matter? (I know very little about JWK, so all 👂)

jogu:
> Note that nothing need be "lost"... we can just amend Web Crypto if we need to and something is important to us.

Fair enough, but as per my comment on #49 I'm unclear why we even want to go in this direction.

> What uses kid and why does it matter? (I know very little about JWK, so all 👂)

It's a key identifier. The header of an encrypted JWT usually includes the keyid of the key that was used to perform the encryption, so that when the encrypted JWT arrives at the verifier's backend it knows which key to try decrypting it with. (There are other ways to achieve the same result of course, but this is an existing commonly used mechanism.)

Issue #55, opened 2024-01-23
Rename repo to "digital-credentials"


marcoscaceres:
Given that we found that "identity credentials" had been squatted by another spec (fedcm maybe?), can we rename this repo to "digital-credentials"? 

The proposal talks about ["Digital Credential"](https://github.com/WICG/identity-credential/blob/main/digital-credentials-2-proposal.md).

samuelgoto:
> "identity credentials" had been squatted by another spec (fedcm maybe?),

I wouldn't call it squatting, but yes, FedCM is built on top of the [`IdentityCredential`](https://fedidcg.github.io/FedCM/#browser-api-identity-credential-interface) type, and Chrome designed `IdentityCredential` to extend beyond federation ([this](https://github.com/fedidcg/FedCM/blob/main/meetings/2021/Web%20Identity%20API.pdf) should give you an idea of the use cases we are actively working on).

> an we rename this repo to "digital-credentials"?

I personally really like "Digital Credentials": it maps well to how the OpenID Foundation is organizing themselves to develop the OpenID4VP specs under the [Digital Credentials Working Group](https://openid.net/wg/digital-credentials-protocols/).

So, +1 from my side.

marcoscaceres:
@samuelgoto wrote:

>I wouldn't call it squatting,

Apologies, it was a poor word choice.  

@TallTed, as I see it, there are multiple motivating factors for the change:

 * the repository's name is for the community (primarily CG members, eventually some WG) to easily find the work. I personally keep getting confused trying to find the thing and I'm trying to stop typing "identity-credential".  
 
Secondly, yes, there is the larger question of scope of the work: 

 * requesting an digital identity credential. 
 * The resulting "digital credential" (`DigitalCredential`). 

On the second point, I'd like to keep the generic "Digital Credential" because the `DigitalCredential` interface should be a generic platform primitive that represents any "digital credential" (we will be building up to this over the next few months or year(s)). 

As for the W3C part, this is not yet a "W3C" thing... its a "WICG" thing, which is denoted by the "WICG/" org... eventually we will move this to "W3C/digital-credential".

marcoscaceres:
As agreed to on the call, renamed the repo to "digital-credentials". 

We can continue the more interesting questions around representing `DigitalCredentials` (as well as overlap with FedCM) as part of the spec development process and other issues.

Issue #49, opened 2023-12-04
Do we really need `publicKey` to be outside of the `request` property?


samuelgoto:
In [this](https://github.com/WICG/identity-credential/blob/main/digital-credentials-2-proposal.md), `publicKey` is part of the top-level parameter. 

```javascript
dictionary IdentityRequestProvider {
 required DOMString protocol; // See "Protocol Registry" below
 required DOMString request;
 required DOMString publicKey;
}
```

Should `publicKey` be (opaquely, optionally and per protocol) part of `request` instead of `IdentityRequestProvider`?

hlozi:
So, having the public key on the top level is part of making encryption work for all protocols. It's like having a common response encryption rule that everyone using the Web API follows. This has a lot of benefits for security. For one, it makes encryption more consistent and easier to implement, which means there's less chance of mistakes or weaknesses.
Response encryption provides security mitigations like phishing protection. This requires interaction between the browser and the wallet application. Standardizing this makes it easier for browsers and wallets to interact.
So, in a nutshell, using a standardized HPKE Encryption for response encryption when using the Web API, makes encryption work across different protocols while the wallet still does the encryption by itself of the response E2E to the RP.

OR13:
`DOMString publicKey` .. I'd prefer a more precise type for the public key, recommend jwk mandatory to support, pem optional, etc... don't like the idea of arbitrary strings.

hlozi:
It could also be a certificate. But agreed that the field "publicKey " should be well-defined.

timcappalli:
2023-12-28 call:

Consensus on the following:

- the public key property should be required and should also live at the top level (independent of protocol)
- the value should be more structured than just a DOMString
- may be redundant with a key acquired via an out of band method, but cannot be contradictory
- need to further refine HPKE, including crypto agility

marcoscaceres:
Could publicKey be a [`CryptoKey`](https://www.w3.org/TR/WebCryptoAPI/#dfn-CryptoKey)?

```idl
interface CryptoKey {
  readonly attribute KeyType type;
  readonly attribute boolean extractable;
  readonly attribute object algorithm;
  readonly attribute object usages;
};
```

marcoscaceres:
Moved the discussion around the shape of the public key here: https://github.com/WICG/identity-credential/issues/56

[...]

PR #57, opened 2024-01-24
Add initial spec

### LINKED ISSUE #61: Is "requestIdentity()" the best method name and do we need it?
[As part of PR 57](https://github.com/WICG/digital-credentials/pull/57#discussion_r1465445128), @samuelgoto wrote:

> We don't have to do this all at once, but I'm wondering if something like a requestDigitalCredential may be a better (more specific ) method name in comparison to requestidentity ... again, because this is going to collide with Chrome's implementation of the IdentityCredential which FedCM uses.

We should consider some alternatives or decide if we are ok with `requestIdentity()`.

marcoscaceres:
Adds the initial spec based on [latest proposal](https://github.com/WICG/identity-credential/blob/main/digital-credentials-2-proposal.md).

closes #61

***

### :boom: Error: 502 Bad Gateway :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 3, 2024, 12:42 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2FWICG%2Fdigital-credentials%2F06d9a61d1e82545e5fcdf2d9e27daaddd080a78c%2Findex.html%3Fforce%3D1%26isPreview%3Dtrue)

```
error code: 502
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/digital-credentials%2357.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,308 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>
+      Digital Credentials API
+    </title>
+    <script src="https://www.w3.org/Tools/respec/respec-w3c" class=
+    "remove"></script>
+    <script class="remove">
+    "use strict";
+    // See https://github.com/w3c/respec/wiki/ for how to configure ReSpec
+    var respecConfig = {
+      github: "WICG/digital-credentials",
+      editors: [
+        {
+          name: "Marcos Caceres",
+          email: "marcosc@apple.com",
+          company: "Apple Inc.",
+          companyURL: "https://apple.com",
+        },
+        // Add additional editors here.
+        // https://github.com/w3c/respec/wiki/editors
+      ],
+      shortName: "digital-credentials",
+      specStatus: "CG-DRAFT",
+      group: "wicg",
+      localBiblio: {
+        "ISO-18013-5": {
+          title: "ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving licence — Part 5: Mobile driving licence (mDL) application",
+          href: "https://www.iso.org/standard/69084.html",
+          publisher: "ISO",
+          date: "2021-09",
+          status: "Published",
+          seriesInfo: "ISO/IEC 18013-5:2021",
+        },
+      },
+      xref: {
+        profile: "web-platform",
+      },
+    };
+    </script>
+  </head>
+  <body data-cite="credential-management-1 WebCryptoAPI">
+    <section id="abstract">
+      <p>
+        This document standardizes an API solution for requesting digital
+        identity credentials (e.g., a drivers license) or verifying some aspect
+        of a digital identity credential (e.g., proof of age). The API builds
+        on the [[[credential-management-1]]] as a means to request a digital
+        credential from the user agent or underlying platform.
+      </p>
+    </section>
+    <section id="sotd">
+      <p>
+        This is an unofficial proposal.
+      </p>
+    </section>
+    <h2>
+      Introduction
+    </h2>
+    <p>
+      TBW
+    </p>
+    <h2>
+      Model
+    </h2>
+    <dl class="definitions">
+      <dt>
+        <dfn>digital identity credential</dfn>
+      </dt>
+      <dd>
+        A digital identity credential is a digital representation of a physical
+        identity credential (e.g., a drivers license, passport, etc). A digital
+        identity credential represents a set of claims about the subject of the
+        credential (e.g., the holder of the credential). A digital identity
+        credential is issued by an issuer (e.g., a government agency).
+      </dd>
+      <dt>
+        <dfn>Identity credential provider</dfn>
+      </dt>
+      <dd>
+        <p>
+          A credential provider is a an application or OS-level service that
+          presents UI to the user to request a digital identity credentia For
+          example,s a credential provider could be a mobile wallet application
+          that requests a mobile drivers license from the user.
+        </p>
+        <p>
+          A credential provider is responsible for presenting the user with a
+          list of [=digital credential/request protocol=] specific providers
+          (e.g., a list of mobile drivers license or applications that hold
+          driver's licenses).
+        </p>
+      </dd>
+      <dt>
+        <dfn for="digital identity">request protocol</dfn>
+      </dt>
+      <dd>
+        The request protocol is a standardized format/structure to [=request
+        digital identity credentials=]. For example, a request protocol could
+        be a <a>"mdoc"</a> for mobile drivers licenses as per [[?ISO-18013-5]].
+      </dd>
+    </dl>
+    <h2>
+      Scope
+    </h2>
+    <p>
+      The following is in is scope for the this specification:
+    </p>
+    <ul>
+      <li>Requesting a [=digital identity credentials=].
+      </li>
+    </ul>
+    <p>
+      The following is out of scope
+    </p>
+    <ul>
+      <li>Creating, updating, deleting, or otherwise managing [=digital
+      identity credentials=].
+      </li>
+    </ul>
+    <h2>
+      Extensions to the `CredentialsContainer` interface
+    </h2>
+    <pre class="idl">
+    partial interface CredentialsContainer {
+      Promise&lt;DigitalCredential&gt; requestIdentity(IdentityRequestOptions options);
+    };
+    </pre>
+    <h3>
+      The `requestIdentity()` method
+    </h3>
+    <p data-dfn-for="CredentialsContainer">
+      When the <dfn>requestIdentity()</dfn> method is called, the user agent
+      MUST run the [=request digital identity credentials=] steps.
+    </p>
+    <h2>
+      The `IdentityRequestOptions` dictionary
+    </h2>
+    <p>
+      The {{IdentityRequestOptions}} dictionary is used to specify options when
+      requesting a digital identity credential.
+    </p>
+    <pre class="idl">
+    dictionary IdentityRequestOptions {
+      AbortSignal signal;
+      required sequence&lt;IdentityRequestProvider&gt; providers;
+    };
+    </pre>
+    <p>
+      The <dfn data-dfn-for="IdentityRequestOptions">providers</dfn> member is
+      a sequence of [=digital credential/request protocol=] specific providers.
+    </p>
+    <h3>
+      The `signal` member
+    </h3>
+    <p>
+      The <dfn data-dfn-for="IdentityRequestOptions">signal</dfn> member is an
+      {{AbortSignal}} that allows the caller to abort an ongoing [=request
+      digital identity credentials=].
+    </p>
+    <h2>
+      The `IdentityRequestProvider` dictionary
+    </h2>
+    <p>
+      The {{IdentityRequestProvider}} dictionary is used to specify a [=digital
+      credential/request protocol=] specific provider, while also handling
+      cryptographic requirements.
+    </p>
+    <pre class="idl">
+    dictionary IdentityRequestProvider {
+      required DOMString protocol;
+      required ArrayBuffer request;
+      required CryptoKey publicKey;
+    };
+    </pre>
+    <h3>
+      The `protocol` member
+    </h3>
+    <p>
+      The <dfn data-dfn-for="IdentityRequestProvider">protocol</dfn> member
+      denotes the [=digital credential/request protocol=] when requesting an
+      identify credential via {{CredentialsContainer/requestIdentity()}}.
+    </p>
+    <p>
+      The {{IdentityRequestProvider/protocol}} member's value is be one of the
+      keys defined in [[[#protocol-registry]]].
+    </p>
+    <h3>
+      The `request` member
+    </h3>
+    <p>
+      The <dfn data-dfn-for="IdentityRequestProvider">request</dfn> member is
+      the request to be sent to the [=identity credential provider=].
+    </p>
+    <h3>
+      The `publicKey` member
+    </h3>
+    <p>
+      The <dfn data-dfn-for="IdentityRequestProvider">publicKey</dfn> member is
+      the public key of the [=identity credential provider=].
+    </p>
+    <h2>
+      Requesting a digital identity credential
+    </h2>
+    <p>
+      The following steps are taken to <dfn>request digital identity
+      credentials</dfn>, with {{IdentityRequestOptions}} |options|:
+    </p>
+    <ol>
+      <li>Let |promise| be [=a new promise=].
+      </li>
+      <li>Return |promise| and continue running the following steps [=in
+      parallel=].
+      </li>
+      <li>For each |provider:IdentityRequestProvider| of |options|'s
+      {{IdentityRequestOptions/providers}}.
+        <ol>
+          <li>[=verify the public key=] of |provider|.
+          </li>
+        </ol>
+      </li>
+      <li>...present [=identity credential provider=] to user.
+      </li>
+    </ol>
+    <h2>
+      The `DigitalCredential` interface
+    </h2>
+    <p>
+      The {{DigitalCredential}} interface represents a [=digital identity
+      credential=].
+    </p>
+    <pre class="idl">
+    [Exposed=Window, SecureContext]
+    interface DigitalCredential : Credential {
+      readonly attribute DOMString protocol;
+      readonly attribute ArrayBuffer data;
+    };
+    </pre>
+    <h3>
+      The `protocol` member
+    </h3>
+    <p>
+      The <dfn data-dfn-for="DigitalCredential">protocol</dfn> member is the
+      [=digital credential/request protocol=] that was used to request the
+      credential.
+    </p>
+    <h3>
+      The `data` member
+    </h3>
+    <p>
+      The <dfn data-dfn-for="DigitalCredential">data</dfn> member is the
+      credential's encrypted data.
+    </p>
+    <h2>
+      Public key verification
+    </h2>
+    <p>
+      The following steps are taken to <dfn>verify the public key</dfn> of a
+      [=identity credential provider=], with {{CryptoKey}} |key|:
+    </p>
+    <ol>
+      <li>...restrictions on type, format, and usage of |key|...
+      </li>
+    </ol>
+    <h2 id="protocol-registry">
+      Identity Request Protocols Registry
+    </h2>
+    <p>
+      The following is the initial registry of [=digital credential/request
+      protocols=] that are supported by this specification.
+    </p>
+    <p>
+      User agents MUST support the following [=digital credential/request
+      protocols=]:
+    </p>
+    <table class="data">
+      <thead>
+        <tr>
+          <th>
+            Protocol identifier
+          </th>
+          <th>
+            Description
+          </th>
+          <th>
+            Specification
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td>
+            <code><dfn>"mdoc"</dfn></code>
+          </td>
+          <td>
+            A mobile driving license (mDL) request.
+          </td>
+          <td>
+            [[ISO-18013-5]]
+          </td>
+        </tr>
+      </tbody>
+    </table>
+    <section id="conformance"></section>
+  </body>
+</html>
--- a/index.html
+++ b/index.html
@@ -43,9 +43,14 @@
     <section id="abstract">
       <p>
         This document specifies an API to allow user agents to mediate access
-        to, and representing, a verifiably-issued digital identity (e.g. a government issued digital driver's license, a passport issued by a country, a student card or diploma issued by a university, an employee card issued by a company, a passenger's boarding pass issued by an airline, a profile issued by an online social network, membership cards, vaccination records, etc)  
-        license). The API builds on [[[credential-management-1]]] as a means to
-        request a digital identity from the user agent or underlying platform.
+        to, and representing, a verifiably-issued digital identity (e.g. a
+        government issued digital driver's license, a passport issued by a
+        country, a student card or diploma issued by a university, an employee
+        card issued by a company, a passenger's boarding pass issued by an
+        airline, a profile issued by an online social network, membership
+        cards, vaccination records, etc) license). The API builds on
+        [[[credential-management-1]]] as a means to request a digital identity
+        from the user agent or underlying platform.
       </p>
     </section>
     <section id="sotd">
@@ -69,8 +74,10 @@
       <dd>
         <p>
           A specialized type of [=credential=] corresponding to the real-world
-          identity of a person enabling a [=verifier=] to make authentication
-          decisions based on identity statements verifiably made by an [=issuer=].
+          identity of a person enabling a <a data-cite=
+          "vc-data-model-2.0#dfn-verifier">verifier</a> to make authentication
+          decisions based on identity statements verifiably made by an
+          [=issuer=].
         </p>
         <aside class="Note">
           [=Credentials=] that are classified as [=identity credentials=]
@@ -198,7 +205,8 @@
     </p>
     <p>
       The {{IdentityRequestProvider/protocol}} member's value is be one of the
-      well-defined keys defined in [[[#protocol-registry]]] or any other custom one.
+      well-defined keys defined in [[[#protocol-registry]]] or any other custom
+      one.
     </p>
     <h3>
       The `request` member
--- a/index.html
+++ b/index.html
@@ -44,10 +44,10 @@
   <body data-cite="credential-management-1 WebCryptoAPI">
     <section id="abstract">
       <p>
-        This document specifies an API solution for requesting digital
-        identity credentials (e.g. a drivers license) or verifying some aspect
-        of a digital identity credential (e.g., proof of age). The API builds
-        on the [[[credential-management-1]]] as a means to request a digital
+        This document specifies an API solution for requesting digital identity
+        credentials (e.g. a driver's license) or verifying some aspect of a
+        digital identity credential (e.g., proof of age). The API builds on the
+        [[[credential-management-1]]] as a means to request a digital
         credential from the user agent or underlying platform.
       </p>
     </section>
@@ -71,10 +71,11 @@
       </dt>
       <dd>
         A digital identity credential is a digital representation of a physical
-        identity credential (e.g., a drivers license, passport, etc). A digital
-        identity credential represents a set of claims about the subject of the
-        credential (e.g., the holder of the credential). A digital identity
-        credential is issued by an issuer (e.g., a government agency).
+        identity credential (e.g., a driver's license, passport, etc). A
+        digital identity credential represents a set of claims about the
+        subject of the credential (e.g., the holder of the credential). A
+        digital identity credential is issued by an issuer (e.g., a government
+        agency).
       </dd>
       <dt>
         <dfn>Identity credential provider</dfn>
@@ -84,12 +85,12 @@
           A credential provider is a an application or OS-level service that
           presents UI to the user to request a digital identity credentia For
           example,s a credential provider could be a mobile wallet application
-          that requests a mobile drivers license from the user.
+          that requests a mobile driver's license from the user.
         </p>
         <p>
           A credential provider is responsible for presenting the user with a
           list of [=digital credential/request protocol=] specific providers
-          (e.g., a list of mobile drivers license or applications that hold
+          (e.g., a list of mobile driver's license or applications that hold
           driver's licenses).
         </p>
       </dd>
@@ -99,7 +100,8 @@
       <dd>
         The request protocol is a standardized format/structure to [=request
         digital identity credentials=]. For example, a request protocol could
-        be a <a>"mdoc"</a> for mobile drivers licenses as per [[?ISO-18013-5]].
+        be a <a>"mdoc"</a> for mobile driver's licenses as per
+        [[?ISO-18013-5]].
       </dd>
     </dl>
     <h2>
--- a/index.html
+++ b/index.html
@@ -83,14 +83,14 @@
         <p>
           A credential provider is a an application or OS-level service that
           presents UI to the user to request a [=digital credential=]. For
-          example, a credential provider could be a wallet application
-          that requests a driver's license from the user.
+          example, a credential provider could be a wallet application that
+          requests a driver's license from the user.
         </p>
         <p>
           A credential provider is responsible for presenting the user with a
           list of [=digital credential/request protocol=] specific providers
-          (e.g., a list of driver's license or applications that hold
-          driver's licenses).
+          (e.g., a list of driver's license or applications that hold driver's
+          licenses).
         </p>
       </dd>
       <dt>
--- a/index.html
+++ b/index.html
@@ -240,7 +240,7 @@
     [Exposed=Window, SecureContext]
     interface DigitalIdentity : Identity {
       readonly attribute DOMString protocol;
-      readonly attribute ArrayBuffer data;
+      readonly attribute DOMString data;
     };
     </pre>
     <h3>
--- a/index.html
+++ b/index.html
@@ -44,7 +44,7 @@
   <body data-cite="credential-management-1 WebCryptoAPI">
     <section id="abstract">
       <p>
-        This document standardizes an API solution for requesting digital
+        This document specifies an API solution for requesting digital
         identity credentials (e.g., a drivers license) or verifying some aspect
         of a digital identity credential (e.g., proof of age). The API builds
         on the [[[credential-management-1]]] as a means to request a digital
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title>
-      Digital Credentials API
+      Digital Credentials
     </title>
     <script src="https://www.w3.org/Tools/respec/respec-w3c" class=
     "remove"></script>
--- a/index.html
+++ b/index.html
@@ -83,13 +83,13 @@
         <p>
           A credential provider is a an application or OS-level service that
           presents UI to the user to request a [=digital credential=]. For
-          example, a credential provider could be a mobile wallet application
-          that requests a mobile driver's license from the user.
+          example, a credential provider could be a wallet application
+          that requests a driver's license from the user.
         </p>
         <p>
           A credential provider is responsible for presenting the user with a
           list of [=digital credential/request protocol=] specific providers
-          (e.g., a list of mobile driver's license or applications that hold
+          (e.g., a list of driver's license or applications that hold
           driver's licenses).
         </p>
       </dd>
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title>
-      Digital Credentials
+      Digital Identities
     </title>
     <script src="https://www.w3.org/Tools/respec/respec-w3c" class=
     "remove"></script>
@@ -19,10 +19,16 @@
           company: "Apple Inc.",
           companyURL: "https://apple.com",
         },
+        {
+          name: "Sam Goto",
+          email: "goto@google.com",
+          company: "Google Inc.",
+          companyURL: "https://google.com",
+        },
         // Add additional editors here.
         // https://github.com/w3c/respec/wiki/editors
       ],
-      shortName: "digital-credentials",
+      shortName: "digital-identity",
       specStatus: "CG-DRAFT",
       group: "wicg",
       localBiblio: {
@@ -44,11 +50,10 @@
   <body data-cite="credential-management-1 WebCryptoAPI">
     <section id="abstract">
       <p>
-        This document specifies an API solution for requesting digital
-        credentials (e.g. a driver's license) or verifying some aspect of a
-        digital credential (e.g., proof of age). The API builds on the
-        [[[credential-management-1]]] as a means to request a digital
-        credential from the user agent or underlying platform.
+        This document specifies an API to allow user agents to mediate access
+        to, and representing, a digital identity (e.g. a digital driver's
+        license). The API builds on [[[credential-management-1]]] as a means to
+        request a digital identity from the user agent or underlying platform.
       </p>
     </section>
     <section id="sotd">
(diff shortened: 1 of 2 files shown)


hlozi (approved):

index.html
```diff
@@ -0,0 +1,293 @@
...
+          A credential provider is a an application or OS-level service that
+          presents UI to the user to request a digital identity credentia For
+          example,s a credential provider could be a mobile wallet application
+          that requests a mobile drivers license from the user.
+        </p>
+        <p>
+          A credential provider is responsible for presenting the user with a
+          list of [=digital credential/request protocol=] specific providers
+          (e.g., a list of mobile drivers license or applications that hold
+          driver's licenses).
```
TallTed, line 93: Is it `drivers license` or `driver's license`? 

In either case, these should be consistent throughout, which they currently aren't.

TallTed (changes requested):

index.html
```diff
@@ -0,0 +1,293 @@
...
+    var respecConfig = {
+      github: "WICG/digital-credentials",
+      editors: [
+        {
+          name: "Marcos Caceres",
+          email: "marcosc@apple.com",
+          company: "Apple Inc.",
+          companyURL: "https://apple.com",
+        },
+        // Add additional editors here.
```
samuelgoto, line 22: Can we have someone from Chrome here (you can use me as a placeholder, we'll find someone to edit this more long term) so that we can make sure that what's in the spec also represents what we'd implement in Chrome.

Also, procedurally, can we agree on blocking normative PR merges with an LGTM from the two implementions?

It would be good if we could also get Ben from Mozilla here to help edit, in case Firefox has an immediate desire to implement this too.

index.html
```diff
@@ -0,0 +1,293 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>
+      Digital Credentials API
```
samuelgoto, line 6: I personally think that the term Digital Credentials is the best that we could use, but I hear that this isn't something particularly shared by everyone. Maybe we could put up for discussion in a issue and rename it as a separate step?

index.html
```diff
@@ -0,0 +1,293 @@
...
+    </title>
+    <script src="https://www.w3.org/Tools/respec/respec-w3c" class=
+    "remove"></script>
+    <script class="remove">
+    "use strict";
+    // See https://github.com/w3c/respec/wiki/ for how to configure ReSpec
+    var respecConfig = {
+      github: "WICG/digital-credentials",
```
samuelgoto, line 14: digital-credentails doesn't exist yet, does it?

index.html
```diff
@@ -0,0 +1,293 @@
...
+      xref: {
+        profile: "web-platform",
+      },
+    };
+    </script>
+  </head>
+  <body data-cite="credential-management-1 WebCryptoAPI">
+    <section id="abstract">
+      <p>
+        This document standardizes an API solution for requesting digital
```
samuelgoto, line 47:
```suggestion
        This document specifies an API solution for requesting digital
```

CGs can't create standards, right? We'd standardize this much later in the process when we move this spec into a WG?

index.html
```diff
@@ -0,0 +1,293 @@
...
+        identity credentials (e.g., a drivers license) or verifying some aspect
```
samuelgoto, line 48:
```suggestion
        identity credentials (e.g. a drivers license) or verifying some aspect
```

index.html
```diff
@@ -0,0 +1,293 @@
...
+        of a digital identity credential (e.g., proof of age). The API builds
```
samuelgoto, line 49:
```suggestion
        of a digital identity credential (e.g. proof of age). The API builds
```

index.html
```diff
@@ -0,0 +1,293 @@
...
+      <li>Creating, updating, deleting, or otherwise managing [=digital
+      identity credentials=].
+      </li>
+    </ul>
+    <h2>
+      Extensions to the `CredentialsContainer` interface
+    </h2>
+    <pre class="idl">
+    partial interface CredentialsContainer {
+      Promise&lt;DigitalCredential&gt; requestIdentity(IdentityRequestOptions options);
```
samuelgoto, line 128:
```suggestion
      Promise&lt;DigitalCredential&gt; requestIdentity(DigitalCredentialRequestOptions options);
```

Maybe we should use the DigitalCredential prefix for these mas, because otherwise we are going to run very quickly into a collision with the Identity Credential maps defined [here](https://fedidcg.github.io/FedCM/). For example, the [IdentityCredentialRequestOptions](https://fedidcg.github.io/FedCM/#enumdef-identitycredentialrequestoptionscontext).

samuelgoto, line 128: We don't have to do this all at once, but I'm wondering if something like a `requestDigitalCredential` may be a better (more specific ) method name in comparison to `requestidentity` ... again, because this is going to collide with Chrome's implementation of the `IdentityCredential` which FedCM uses.

samuelgoto: Point being: we don't have to rename this now in this PR, but wanted to give the heads up that Chrome ships IdentityCredential and we'd like to avoid the conflicts here.

index.html
```diff
@@ -0,0 +1,293 @@
...
+    </p>
+    <h2>
+      The `IdentityRequestOptions` dictionary
+    </h2>
+    <p>
+      The {{IdentityRequestOptions}} dictionary is used to specify options when
+      requesting a digital identity credential.
+    </p>
+    <pre class="idl">
+    dictionary IdentityRequestOptions {
```
samuelgoto, line 146:
```suggestion
    dictionary DigitalCredentialRequestOptions {
```

As I said earlier, I'd use a `DigitalCredential` prefix as opposed to `Identity` prefix, because otherwise Chrome is going to quickly run into a conflict with the interfaces defined by the `IdentityCredential` maps, for example [this](https://fedidcg.github.io/FedCM/#dictdef-identitycredentialrequestoptions) one.

index.html
```diff
@@ -0,0 +1,293 @@
...
+    <h2>
+      The `DigitalCredential` interface
+    </h2>
+    <p>
+      The {{DigitalCredential}} interface represents a [=digital identity
+      credential=].
+    </p>
+    <pre class="idl">
+    [Exposed=Window, SecureContext]
+    interface DigitalCredential : Credential {
```
samuelgoto, line 236: I think inheriting from `Credential` would create a problem for implementations, because, according to the Credential Manager spec, implementations MUST loop through the subclasses and look the internal methods.

https://www.w3.org/TR/credential-management-1/#credential-internal-methods

```
Unless otherwise specified, each [interface object](https://heycam.github.io/webidl/#dfn-interface-object) created for interfaces which [inherit](https://heycam.github.io/webidl/#dfn-inherit) from [Credential](https://www.w3.org/TR/credential-management-1/#credential) MUST provide implementations for at least one of these internal methods, overriding [Credential](https://www.w3.org/TR/credential-management-1/#credential)'s default implementations, as appropriate for the [credential](https://www.w3.org/TR/credential-management-1/#concept-credential) type. E.g., [§3.2 The PasswordCredential Interface](https://www.w3.org/TR/credential-management-1/#passwordcredential-interface), [§4.1 The FederatedCredential Interface](https://www.w3.org/TR/credential-management-1/#federatedcredential-interface), and [[WEBAUTHN]](https://www.w3.org/TR/credential-management-1/#biblio-webauthn).
```

We could emphasize on the "Unless otherwise specified" part, but the challenge here is that this `DigitalCredential` here really isn't a Credential Manager `Credential` in the sense that all of the other ones are (PublicKeyCredential, PasswordCredential, FederatedCredential and IdentityCredential).

So, I think we need to make a decision as to whether we want to integrate properly with the Credential Management API (e.g. implement the other methods, like `[[CollectFromCredentialStore]]`) or not (e.g. not subclass `Credential` here).

WDYT?

index.html
```diff
@@ -0,0 +1,293 @@
...
+          </th>
+          <th>
+            Description
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td>
+            `"mdoc"`
```
samuelgoto, line 280: Can we add this section on a separate PR? Or, otherwise, add VC to this list? And if Safari is not interested in supporting VCs, than remove mDocs so that we can reserve the right to not support that either?

samuelgoto (commented):
Thanks for putting this together!! It is a great starting point and is really going to help us find the convergence that we need.

Comments inline!

index.html
TallTed: [`e.g.,` and `i.e.,` should always be followed by (and usually preceded by) a comma](https://www.dailywritingtips.com/comma-after-i-e-and-e-g/).

If you don't like those commas for whatever reason, the abbreviations should be replaced by text like `for example` or `that is to say`, which may or may not also call for following and/or preceding commas.

index.html
```diff
@@ -0,0 +1,293 @@
...
+    <p>
+      The {{IdentityRequestProvider}} dictionary is used to specify a
+      [=digital credential/request protocol=] specific provider, while also handling
+      cryptographic requirements.
+    </p>
+    <pre class="idl">
+    dictionary IdentityRequestProvider {
+      required DOMString protocol;
+      required ArrayBuffer request;
+      required CryptoKey publicKey;
```
samuelgoto, line 175: Can we remove this until we get some convergence on this:

https://github.com/WICG/identity-credential/issues/56

I, for one, am not convinced that we need to have `publicKey` here at all.

index.html
samuelgoto: Yeah, OWNERs could help here too.

index.html
OR13: I agree with this comment, I would also suggest adding a citation to at least one digital credential format, that can be obtained at no cost, per: https://www.w3.org/Guide/process/tilt/normative-references.html

index.html
OR13: I think it can be valuable to have the public key here, but in the interest of landing the initial spec in github, I am supportive of seeing it added in a separate PR after this one.

index.html
OR13: If these APIs end up very different, and the only difference in their names is the word `Digital` I think that will lead to a lot of bug reports and developer confusion.

index.html
```diff
@@ -0,0 +1,293 @@
...
+      group: "wicg",
+      localBiblio: {
+        "ISO-18013-5": {
+          title: "ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving licence — Part 5: Mobile driving licence (mDL) application",
+          href: "https://www.iso.org/standard/69084.html",
+          publisher: "ISO",
+          date: "2021-09",
+          status: "Published",
+          seriesInfo: "ISO/IEC 18013-5:2021",
+        },
```
OR13, line 36:
```suggestion
      localBiblio: {
        "ISO-18013-5": {
          title: "ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving licence — Part 5: Mobile driving licence (mDL) application",
          href: "https://www.iso.org/standard/69084.html",
          publisher: "ISO",
          date: "2021-09",
          status: "Published",
          seriesInfo: "ISO/IEC 18013-5:2021",
        },
        "vc-data-model-2.0": {
          title: "The Verifiable Credentials Data Model v2.0",
          href: "https://www.w3.org/TR/vc-data-model-2.0/",
          publisher: "W3C Verifiable Credentials Working Group",
          status: "WD"
        }
```

index.html
```diff
@@ -0,0 +1,293 @@
...
+          <td>
+            `"mdoc"`
+          </td>
+          <td>
+            [[[ISO-18013-5]]]
+          </td>
+          <td>
+            A mobile driving license (mDL) request.
+          </td>
+        </tr>
```
OR13, line 288:
```suggestion
        <tr>
          <td>
            `"mdoc"`
          </td>
          <td>
            [[[ISO-18013-5]]]
          </td>
          <td>
            A mobile document (mDL) request.
          </td>
        </tr>
        <tr>
          <td>
            `"vc"`
          </td>
          <td>
            [[[vc-data-model-2.0]]]
          </td>
          <td>
            A verifiable credential (vc) request.
          </td>
        </tr>
```

index.html
```diff
@@ -0,0 +1,293 @@
...
+      <tbody>
+        <tr>
+          <td>
+            `"mdoc"`
+          </td>
+          <td>
+            [[[ISO-18013-5]]]
+          </td>
+          <td>
+            A mobile driving license (mDL) request.
```
OR13, line 286: does this work for other document types, or just mobile drivers licenses?

index.html
OR13: actually based on https://github.com/WICG/identity-credential/blob/main/digital-credentials-2-proposal.md#the-identityrequestoptions-dictionary I think perhaps this should be `oid4vp` ?

awoie (changes requested):
It would be good to have IETF SD-JWT VC as well.

awoie (changes requested):
VCDM 2.0 is not a protocol. It technically is agnostic to securing mechanisms. IMO, this should be more specific.

Sakurann (changes requested):
Should also include IETF SD-JWT VC and not just ISO mdoc and W3C VCDM

index.html
hlozi: > Can we remove this until we get some convergence on this:
> 
> #56
> 
> I, for one, am not convinced that we need to have `publicKey` here at all.

We discussed this in the  2023-12-18 call:

meeting notes: https://github.com/WICG/identity-credential/wiki/2023-12-18-Meeting-Notes#do-we-really-need-publickey-to-be-outside-of-the-request-property

Consensus on the following: ( copied from the issue) 

the public key property should be required and should also live at the top level (independent of protocol)
the value should be more structured than just a DOMString
maybe redundant with a key acquired via an out-of-band method, but cannot be contradictory
need to further refine HPKE, including crypto agility

Happy to re-discuss it if needed.

index.html
Sakurann: 18013-5 says `driving licence` (british english) :)

index.html
Sakurann: not sure how to parse `can we agree on blocking normative PR merges with an LGTM from the two implementions`. Do you intend to merge a PR if it has approvals from at least 2 implementors regardless of the number of people requesting for changes?

index.html
Sakurann: How much of the other methods of CredMan API do we think digital credentials API will reuse?
It has been pointed out number of times that "credential" in a conventional/CredMan refers to the things like password, wubauthn keypair, which is different from how "verifiable/digital credentials" refer to collection of attributes of the user attested by the issuer. It might be that despite all this, a lot of the same methods can be reused, I don't know, but that sounds like something that could influence the decision here.

index.html
tplooker: Was there a reason to not use the "Identity Credential" terminology here?

index.html
```diff
@@ -0,0 +1,293 @@
...
+      </dt>
+      <dd>
+        A digital identity credential is a digital representation of a physical
+        identity credential (e.g., a drivers license, passport, etc). A digital
+        identity credential represents a set of claims about the subject of the
+        credential (e.g., the holder of the credential). A digital identity
+        credential is issued by an issuer (e.g., a government agency).
+      </dd>
+      <dt>
+        <dfn>Identity credential provider</dfn>
```
tplooker, line 80: Just noting here the use of "identity credential provider" when the top level term is "digital credential", I think these terms should be consistent.

index.html
```diff
@@ -0,0 +1,293 @@
...
+    <ol>
+      <li>Let |promise| be [=a new promise=].
+      </li>
+      <li>Return |promise| and continue running the following steps [=in
+      parallel=].
+      </li>
+      <li>For each |provider:IdentityRequestProvider| of |options|'s
+      {{IdentityRequestOptions/providers}}.
+        <ol>
+          <li>[=verify the public key=] of |provider|.
```
tplooker, line 220: Do we mean verify here in the sense of a cryptographic operation, or do we mean validate? e.g, make sure it is well formed and the type is expected/supported?

index.html
```diff
@@ -0,0 +1,293 @@
...
+          driver's licenses).
+        </p>
+      </dd>
+      <dt>
+        <dfn for="digital identity">request protocol</dfn>
+      </dt>
+      <dd>
+        The request protocol is a standardized format/structure to [=request
+        digital identity credentials=]. For example, a request protocol could
+        be a "mDoc" for mobile drivers licenses as per [[?ISO-18013-5]].
```
Sakurann, line 102:
```suggestion
        be a "mdoc" for mobile drivers licenses as per [[?ISO-18013-5]].
```

18013-5 terminology is `mdoc` all lowercase i think

index.html
tplooker: Apologies I didn't see #55 had been raised.

index.html
tplooker: If we mean verify can we clarify what is being verified?

OR13: I assumed something like cert check here, I imagined something like x5c in jwk.

index.html
OR13: thats how I read it, you would just set implementation reps as code owners, and require approval from them.

index.html
```diff
@@ -0,0 +1,308 @@
...
+      <li>Creating, updating, deleting, or otherwise managing [=digital
+      identity credentials=].
+      </li>
+    </ul>
+    <h2>
+      Extensions to the `CredentialsContainer` interface
+    </h2>
+    <pre class="idl">
+    partial interface CredentialsContainer {
+      Promise&lt;DigitalCredential&gt; requestIdentity(IdentityRequestOptions options);
```
bvandersloot-mozilla, line 128: Reiterating from the call: It may be useful to abstract the DigitalCredential return type here and design an algorithm that both FedCM and DigitalCredential can hook into to provide identities. That way FedCM can intermingle in UI with 3-party ids and also with WebAuthN, but mDLs and WebAuthN can't be requested interchangably, which was the motivating concern for this API shape IIRC.

index.html
marcoscaceres: It does now :)

index.html
marcoscaceres: Yeah, basically. That would mirror (somewhat) the [WHATWG working mode](https://whatwg.org/working-mode) and well as the [WebAppsWG working mode](https://www.w3.org/2022/04/webapps-wg-charter.html#working-mode), both of which work really well. 

We can add OWNERS as that automatically flags folks for review, or we can easily set that up via the GitHub settings (that someone with write access needs to also approve it, and we basically just agree on who needs to give a ✅.... it's a lightweight process).

index.html
marcoscaceres: @OR13, vc-data-model-2.0 is already in specref: https://www.specref.org/?q=vc-data-model-2.0

The reason I had to add "ISO-18013-5" to the local biblio is that it's not coming up in specref. I filed a bug to find out why: 
https://github.com/tobie/specref/issues/776

Would you mind dropping this suggestion.

index.html
marcoscaceres: Yeah, it should be "driver's license"... will fix throughout. 

Noting also that W3C specs are conical in  US-EN, but can be translated to other language variants (should we want to do a British English version).

index.html
marcoscaceres: Yeah, good idea.

index.html
marcoscaceres: I haven't read the FedCM spec yet... but yeah, so long as this can be a generic thing without a normative requirement on FedCM, then that works. 

I'm imagining `DigitalCredential` (or whatever we end up calling it) will end up in the Credential Management spec eventually.

index.html
marcoscaceres: We would be verifying that a `CryptoKey` is of the right type. 

A crypto key has the following properties: 

```WebIDL
enum KeyType { "public", "private", "secret" };

enum KeyUsage { "encrypt", "decrypt", "sign", "verify", "deriveKey", "deriveBits", "wrapKey", "unwrapKey" };

[SecureContext,Exposed=(Window,Worker)]
interface CryptoKey {
  readonly attribute KeyType type;
  readonly attribute boolean extractable;
  readonly attribute object algorithm;
  readonly attribute object usages;
};
```

So we would be **verifying** that the `type` is  "public", and reject "private" and "secret" keys... and then continue verifying whatever else we can from the above things.

marcoscaceres: Let's discuss here https://github.com/WICG/digital-credentials/issues/56

index.html
marcoscaceres: @samuelgoto, I agree that we should be building on the Credential Manager spec. The goal is be to eventually copy/paste this over to that spec. And yes, I agree that integration needs to be coherent and we are not doing a separate thing. 

At the same time, there was good reason for not using `.get()`: we can more strictly restrict the call site to just `DigitalCredentialRequestOptions` without allowing the mixing of credential types in the request (yes, we could also enforce that in `.get()`, but it's a bit more messy because it can't be enforced with WebIDL.... nevertheless, whatever we do spec should be `.get()` compatible). Also, calling `.requestIdentity()` was supposed to make it clear to developers which sheet/UI is going to be show: the one to pick a driver's license or similar credential, not for passwords or whatever other things are possible with straight up `.get()`. 

About who we should talk to, I think we should just talk to Credential Manager API folks... speaking as a ex-TAG member, I don't think the TAG will have much to offer here TBH (this is outside their purview, and can only provide very surface guidance... and will probably just tell us to go to talk to WebAppSec 🫠).

index.html
marcoscaceres: Filed https://github.com/WICG/digital-credentials/issues/60 for OWNERS.

index.html
marcoscaceres: Yeah, I kinda also like the generic "credential provider"... let's start with that. 

Eventually we should have a little diagram showing the model.

index.html
marcoscaceres: @samuelgoto,  filed https://github.com/WICG/digital-credentials/issues/61 so we don't lose it. Let's pick this up there.

index.html
marcoscaceres: I also agree. I'll take it out.

marcoscaceres: Filed https://github.com/WICG/digital-credentials/issues/58 for us to figure the registry stuff out.

index.html
marcoscaceres: Taking all this out... let's move to https://github.com/WICG/digital-credentials/issues/58

index.html
marcoscaceres: I was just starting with this because it will be a miracle if we even get driver licenses.

marcoscaceres: Note: removing this and moving discussion to https://github.com/WICG/digital-credentials/issues/58

index.html
marcoscaceres: Just noting that I'm removing this... was just showing what it could look like. 

@Sakurann, looking forward to discussing in https://github.com/WICG/digital-credentials/issues/58

marcoscaceres:
Thanks again everyone for the amazing and super timely feedback 🙏 

Please check that I have addressed your feedback and please give an explicit ✅ if you think this is enough for us to start with. 

Remember, nothing is set in stone here... this is just an unofficial strawperson and, to paraphrase what @samuelgoto said during the call: we assume the whole thing is wrong and likely will need a few rewrites till we get something stable. 

Please file issues and we can go from there!

index.html
marcoscaceres: @samuelgoto, can you live with this here for now? (we need to put the publicKey somewhere, so this seems like an ok start?)

I added a link to #49 below so we can continue to discuss.

marcoscaceres:
I've dropped publicKey for now until we sort out when/how it will be used... maybe we don't need it? cc @samuelgoto

index.html
marcoscaceres: Resolving as not needed.

index.html
bvandersloot-mozilla: That sounds perfect. That just means FedCM would subclass DigitalCredential, in the same way it currently subclasses Credential and would put itself in the registry.

I agree that the generic `DigitalCredential` (or whatever) would live well in the Credential Management spec.

index.html
bvandersloot-mozilla: Chiming in here that making this extend `Credential` would be something I believe we explicitly looked to avoid with the decision to use `.requestIdentity()`. If we subclass, that makes the more "login" centric `.get()` able to request an identity.

I think that the options we have are (1) making a parallel construction for driver's-license like things where we reuse algorithms as helpers or (2) making the `Credential` inherit a more generic object that we lift the shared bits into.

index.html
samuelgoto: Yeah, agreed that we don't want to block this initial PR on this. Let me resolve this to unblock it here, and we can follow up on the issue you create. Thanks!

index.html
samuelgoto: >  I think we should just talk to Credential Manager API folks

Yeah, I think that's the right next step.

> that makes the more "login" centric .get()

I think there are a couple of things that I'm struggling with this proposition:

1) First, I don't share the interpretation that the Credential Manager API is for "login".
2) Second, it is clear to me that the Digital Credentials here are going to be used for "login": it even shows up in regulation.

I think some of my confusion here is that i think maybe we don't quite understand what we mean by "login" (e.g. i think there is a big distinction between sign-up - i.e. account creation and recovery -, and sign-in).

Chrome expects FedCM to be used for a lot of [use cases](https://github.com/fedidcg/FedCM/blob/main/meetings/2021/Web%20Identity%20API.pdf) that I don't think I would call "login", so that's what I'm struggling to reconcile.

index.html
samuelgoto: Thank you, resolving this here and following up on that issue. Thanks for kicking it off.

index.html
samuelgoto: It is not clear to me what the browser would (and should) do with the `publicKey`. Chrome implemented a prototype of this API deliberately delegating the validation of signatures to verifiers and holders, and it is not clear to me what benefit we get by pulling this responsibility to the browser. 

I can be convinced that interoperability/fragmentation is something that I think the browser can help, but I think it would be premature to over-generalize this aspect at the moment, since we don't even know what "protocols" are, let alone force all of them to comply to a specific way to check for digital signatures.

index.html
samuelgoto: Can we try to agree with the goals here first? 

> both FedCM and DigitalCredential can hook into to provide identities

This matches much of my intuition, and it is what Chrome would like to see this design converge to. 

@marcoscaceres, does this match your intuition?

index.html
samuelgoto: Resolving this, since you removed it in the last commit:

https://github.com/WICG/digital-credentials/pull/57#issuecomment-1911289914

index.html
samuelgoto: Re-opening this: didn't we agree to have co-editors (in addition to OWNERs) from other implementations on the last call?

index.html
bvandersloot-mozilla: By login I mean sign-in.  

> First, I don't share the interpretation that the Credential Manager API is for "login".

Sign-in is mentioned in the first sentence of the use cases section of that spec and the opening sentence of that spec is "Signing into websites is more difficult than it should be." 

> Second, it is clear to me that the Digital Credentials here are going to be used for "login": it even shows up in regulation.

I'm not a lawyer so I can't speak to what the EU is mandating, but is it something that we should encourage globally? I say not.

> Chrome expects FedCM to be used for a lot of [use cases](https://github.com/fedidcg/FedCM/blob/main/meetings/2021/Web%20Identity%20API.pdf) that I don't think I would call "login", so that's what I'm struggling to reconcile.

IMO the non-sign-in use cases covered by FedCM can be explained as the consequences of trying to sign in while simultaneously dealing with an out-of-the-page IDP in the best way possible.

index.html
marcoscaceres: https://github.com/WICG/digital-credentials/issues/66

index.html
marcoscaceres: I don't have a strong opinion (because I'm still not very familiar with FedCM). If both specs can use the same thing, then great. However, the problem remains about `.get()` conflating requesting multiple credential types, and it's impossible for an implementation to distinguish (because they are all part of the same `CredentialCreationOptions` dictionary).

In the [Extensions Points](https://www.w3.org/TR/credential-management-1/#implementation-extension) section, the CM spec encourages spec authors to define "a reasonable mechanism for dealing with requesting **credentials of distinct types**", which is what `. requestIdentity()` is trying to do, and then to pitch that to the Web App Sec WG. 

I think we need to document our options here and present our options to the Web App Sec WG to help us decide what the right course of action is here. 

I think we have general agreement that this poses issues:

```JS
const opts = {  publicKey , identity,  providers };
// who wins? what happens? What UI gets shown? 
navigator.credentials.get(opts);
``` 

Which, again, is the core of the problem. 

The `. requestIdentity()` disambiguates the problem by providing a unique method signature, only allowing this new `DigitalCredentialRequestOptions` dictionary, which deliberately does not inherit from `CredentialRequestOptions` to avoid its limitations, while providing the same functionality.

marcoscaceres: Just as tl;dr: let's move this discussion to https://github.com/WICG/digital-credentials/issues/61

marcoscaceres:
This PR is currently blocked on resolving #61. 

Next steps are to discuss the solutions in #61 and set up a call with the editors of the Credential Management Spec. If folks have other opinions or solutions you'd like to discuss, please raise them there.

index.html
```diff
@@ -0,0 +1,310 @@
...
+      group: "wicg",
+      localBiblio: {
+        "ISO-18013-5": {
+          title: "ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving licence — Part 5: Mobile driving licence (mDL) application",
+          href: "https://www.iso.org/standard/69084.html",
+          publisher: "ISO",
+          date: "2021-09",
+          status: "Published",
+          seriesInfo: "ISO/IEC 18013-5:2021",
+        },
```
OR13, line 36:
```suggestion
```

OR13: Currently unused, so not harm in removing... if the goal is to single intended supported digital credential formats, I suggest starting with a full set or starting with an empty set.

samuelgoto: +1 for the "full set or empty set" framing.

index.html
```diff
@@ -0,0 +1,299 @@
...
+    <p data-dfn-for="Navigator">
+      The <dfn>identity</dfn> attribute provides access to the the underlying
+      {{CredentialsContainer}} for managing [=identity credentials=].
+    </p>
+    <h3>
+      Extensions to Credential Management API
+    </h3>
+    <aside class="issue" data-number="65"></aside>
+    <h2>
+      The `DigitalIdentityRequestOptions` dictionary
```
marcoscaceres, line 161:
```suggestion
       Extensions to `CredentialRequestOptions` dictionary
```

index.html
```diff
@@ -0,0 +1,299 @@
...
+    </h2>
+    <pre class="idl">
+    dictionary DigitalIdentityRequestOptions : CredentialRequestOptions {
```
marcoscaceres, line 164:
```suggestion
    partial dictionary CredentialRequestOptions {
```

index.html
```diff
@@ -0,0 +1,299 @@
...
+      required sequence&lt;IdentityRequestProvider&gt; providers;
```
marcoscaceres, line 165:
```suggestion
     sequence&lt;IdentityRequestProvider&gt; providers;
```

index.html
```diff
@@ -0,0 +1,299 @@
...
+    <pre class="idl">
+    partial dictionary CredentialRequestOptions {
+     sequence&lt;IdentityRequestProvider&gt; providers;
+    };
+    </pre>
+    <h3>
+      The `providers` member
+    </h3>
+    <p>
+      The <dfn data-dfn-for="DigitalIdentityRequestOptions">providers</dfn>
```
marcoscaceres, line 172:
```suggestion
      The <dfn data-dfn-for="CredentialRequestOptions">providers</dfn>
```

index.html
```diff
@@ -0,0 +1,302 @@
...
+        profile: "web-platform",
+      },
+    };
+    </script>
+  </head>
+  <body data-cite="credential-management-1 WebCryptoAPI">
+    <section id="abstract">
+      <p>
+        This document specifies an API to allow user agents to mediate access
+        to, and representing, a digital identity (e.g. a digital driver's
```
samuelgoto, line 46:
```suggestion
        to, and representing, a verifiable issued digital identity (e.g. a government issued digital driver's license, a passport issued by a country, a student card or diploma issued by a university, an employee card issued by a company, a passenger's boarding pass issued by an airline, a profile issued by an online social network, membership cards, vaccination records, etc)  
```

index.html
```diff
@@ -0,0 +1,302 @@
...
+      group: "wicg",
+      localBiblio: {
+      },
+      xref: {
+        profile: "web-platform",
+      },
+    };
+    </script>
+  </head>
+  <body data-cite="credential-management-1 WebCryptoAPI">
```
samuelgoto, line 42:
```suggestion
  <body>
```

Bad copy/pasting?

index.html
```diff
@@ -0,0 +1,302 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>
+      Digital Identities
```
samuelgoto, line 6:
```suggestion
      Digital Identity API
```

Maybe more consistent with other spec names? That is, it is an API specifically, rather than an overview of digital identities broadly speaking?

index.html
samuelgoto: I think it would be useful to make sure we are all on the same page with regards to how we are expecting this API to be used, so I added a few more concrete examples that I hear.

Does this match your intuition?

index.html
```diff
@@ -0,0 +1,302 @@
...
+    </h2>
+    <p>
+      TBW
+    </p>
+    <h2>
+      Model
+    </h2>
+    <dl class="definitions" data-sort="">
+      <dt>
+        <dfn>Identity credential</dfn>
```
samuelgoto, line 67:
```suggestion
        <dfn>Identity Credential</dfn>
```

index.html
```diff
@@ -0,0 +1,302 @@
...
+      </dt>
+      <dd>
+        <p>
+          A specialized type of [=credential=] corresponding to the real-world
+          identity of a person enabling developers to make authentication
```
samuelgoto, line 72:
```suggestion
          identity of a person enabling a [=Verifier=] to make authentication
```

index.html
```diff
@@ -0,0 +1,302 @@
...
+          decisions based on verified identity information.
```
samuelgoto, line 73:
```suggestion
          decisions based on verified identity statements made by an [=Issuer=].
```

index.html
```diff
@@ -0,0 +1,302 @@
...
+      </p>
+    </section>
+    <h2>
+      Introduction
+    </h2>
+    <p>
+      TBW
+    </p>
+    <h2>
+      Model
```
samuelgoto, line 63: We are operating at the browser level, so our definition of "Credential" (in the Credential Manager sense) is different than other group's definition of "Credential", but there are many other concepts that I think are well articulated by the ISO and Verifiable Credentials community.

[Here](https://www.w3.org/TR/vc-data-model/) is one example of a data model that I think has a lot of good definitions.

I'm wondering if we are going to get held back agreeing on these definitions, and if we are better off just having the WebIDLs in this initial PR to unblock implementation ... WDYT?

Happy to go either way, but my intuition is that the WebIDL without the definitions more minimally viable.

index.html
```diff
@@ -0,0 +1,302 @@
...
+    </h2>
+    <p>
+      The {{IdentityRequestProvider}} dictionary is used to specify a [=digital
+      identity/request protocol=] and structured request, which the user agent
+      MAY match against a [=identity credential provider=].
+    </p>
+    <pre class="idl">
+    dictionary IdentityRequestProvider {
+      required DOMString protocol;
+      required ArrayBuffer request;
```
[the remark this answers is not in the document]
samuelgoto: I'm also wondering if `request` should be an `object` too, but I think we can leave that for a subsequent PR.

samuelgoto (commented):
This PR really captured what I think is a design that will reconcile many different aspects of the space: reconciling the two party model with the three party model, reconciling the distinction between identity verification and sign-in, reusing the Credential Management spec and algorithms, etc.

I made some comments to try to clarify what I think is already shared understanding. 

The backbone and the structure of this PR is solid. I'm playing with the code to see what would be involved in implementing, but my intuition is that this would set us in the right long term direction.

@bvandersloot-mozilla WDYT?

index.html
```diff
@@ -0,0 +1,302 @@
...
+    <h3>
+      Extensions to Credential Management API
+    </h3>
+    <aside class="issue" data-number="65"></aside>
+    <h2>
+      Extensions to `CredentialRequestOptions` dictionary
+    </h2>
+    <pre class="idl">
+    partial dictionary CredentialRequestOptions {
+      sequence&lt;IdentityRequestProvider&gt; providers;
```
samuelgoto, line 165: So, something that i think is a bit inconsistent with the CredentailsContainer but I think is actually good, is that the CredMan spec goes over in the [Credential Registry](https://w3c.github.io/webappsec-credential-management/#sctn-cred-type-registry) the [options member identifier](https://w3c.github.io/webappsec-credential-management/#credential-type-registry-options-member-identifier), which is an WebIDL attribute that tells you which credential type to use.

```js
await navigator.credentials.get({
  publicKey: /** give me a passkey */,
  otp: /** give me an SMS OTP */,
  identity: /** give me a federated assertion */,
});
```

Whereas `providers` here, would be for `DigitalIdentity`, but that seems like a general term that could apply to other `Identity` subclasses.

```js
await navigator.identity.get({
  providers: [] // what Identity interface do we expect to provide?
});
```

For example, if we introduced a `FederatedIdentity` in the future, then we wouldn't know whether "digital" identities or "federated" identities are being requested.

So, this is a bit different from the `navigator.credentials.get()` API, but I'm not sure if it is necessarily bad.

It is hard to entirely predict the future, but I think the biggest differentiation in the subclasses of `Identity` is actually going to come as a result of introducing new `protocols`. That is, I think that the only distinction between a `decentralized` identity and a `federated` identity is the mechanism that defines how to request them.

That is, if/when we'd like to expose FedCM into this API, we could introduce a new `protocol`, add it to the registry in this spec here, and continuing returning a `DigitalIdentity` or, if necessary, create a subclass of `Identity` that is a `FederatedIdentity`, and the way we would figure out which one to construct is based on `protocol`, NOT the [options member identifier](https://w3c.github.io/webappsec-credential-management/#credential-type-registry-options-member-identifier)  as used in `navigator.credentials.get()`.

I think this could work, but since it deviates from `navigator.credentials.get()` it seemed worth noting.

@bvandersloot-mozilla @marcoscaceres WDYT?

index.html
```diff
@@ -0,0 +1,302 @@
...
+  </head>
+  <body data-cite="credential-management-1 WebCryptoAPI">
+    <section id="abstract">
+      <p>
+        This document specifies an API to allow user agents to mediate access
+        to, and representing, a digital identity (e.g. a digital driver's
+        license). The API builds on [[[credential-management-1]]] as a means to
+        request a digital identity from the user agent or underlying platform.
+      </p>
+    </section>
```
samuelgoto, line 50: Maybe a big example up front would help explain?

```suggestion
<div class=example>
Example showing how a website can request a digital identity.

```javascript
let {protocol, data} = await navigator.identity.get({
  signal: controller,
  mediation: "optional", 
  providers: [{
    protocol: "fake-protocol",
    request: "{type: 'drivers-license'}",
    nonce: "0xCAFE",
  }]
});
```
</div>

    </section>
```

File: index.html (hunk 25 above)
Comment (TallTed) [reply index.html:46]: Whether or not the suggestion above is adopted, it is unclear what `and representing` is connected to.
* Does it mean `to allow user agents to represent a digital identity`?
* Does it mean `an API representing a digital identity`?
* Does it mean something else?

Once this is answered, I will draft a rephrasing ... unless such a rephrasing is how the answer is provided.

----

Separately, I suggest changing `verifiable issued digital identity` to `verifiably-issued digital identity` or `verifiable digital identity`, unless it is really meant to say that an _issued digital identity_ is _verifiable_, which makes me want a clear definition of how an _issued digital identity_ differs from an _**un**issued digital identity_

File: index.html (hunk 30 above)
Comment (TallTed) [reply index.html:73]: This distinction is important. Verifiable Credentials do not provide any indication of the veracity of the statements they contain; they only offer verification that the Issuer made those statements. 
```suggestion
          decisions based on identity statements verifiably made by an [=Issuer=].
```

index.html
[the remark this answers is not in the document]
bc-pi: I suspect most of the folks working on "credentials" from IETF/OIDF communities are envisioning the protocol-specific request content as a string. OpenID4VP is all string based, for example, so whatever incarnation of that in conjunction with this API eventually emerges would be most naturally expressed as a string. I think I share @samuelgoto's view here.

index.html
marcoscaceres: We've generally moved away from having "API" in the names of specs (e.g., "Credential Management", "Permissions", "Fetch", etc.) to cover both the generalized concepts and the API.

index.html
marcoscaceres: This one is ok (except we don't need Web Crypto). We cite "credential-management-1" in various places, so this lets Respec know the context.

index.html
marcoscaceres: It would be good to include these here now, but eventually move the expanded examples into the yet to be written introduction. 

Also agree with @TallTed, should be "verifiably-issued digital identity". I'll change your suggestion to that. 

@TallTed, can you file a bug instead for us to work on the abstract, intron, and scope?

index.html
marcoscaceres: Definitely want to add examples, but I think we should hold till we have the first item in the registry. Then we can add real functional examples.

index.html
marcoscaceres: @TallTed filed https://github.com/WICG/digital-credentials/issues/68

index.html
marcoscaceres: Filed https://github.com/WICG/digital-credentials/issues/69

index.html
samuelgoto: Ha, didn't realize that was the case. I'll update a few more of my specs that I maintain to follow that then.

Resolving.

samuelgoto, line 6: I personally feel that "digital identities" are inclusive of "federated identities", but just wanted to make sure that that's the case, because, if not, then, maybe we'd need to use a more general term (i.e. remove the "digital" prefix)?

My intuition is that "digital identities" is inclusive and a good starting point, but just wanted to be transparent of how I'd want to see this spec evolve over time.

index.html
samuelgoto: Yep, my suggestion was just to remove the "WebCryptoAPI" here.

index.html
marcoscaceres: I've prefer we keep "Credential" lower-case, just for consistency.

index.html
samuelgoto: > Also agree with @TallTed, should be "verifiably-issued digital identity". I'll change your suggestion to that.
"Verifiably-issued digital identity" works for me for now.

index.html
marcoscaceres: we probably don't need to upper case concepts, as they are linked (and we can alternatively do that with CSS if we want to make them more salient).

index.html
samuelgoto: Sure, that works too.

index.html
marcoscaceres: Good point, @TallTed.

index.html
samuelgoto: non-camel-case sgtm

index.html
```diff
@@ -0,0 +1,302 @@
...
+    <ul>
+      <li>Requesting a [=digital identity=], including mechanisms for secure
+      presentation.
+      </li>
+    </ul>
+    <p>
+      The following items are out of scope:
+    </p>
+    <ul>
+      <li>Enrollment process for establishing a [=digital identity=].
```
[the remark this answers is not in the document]
marcoscaceres: basically, what I mean is: "I go to the DMV (or equivalent in whatever country) and I get issued a physical and digital driver's license... somehow, the driver's license ends up in my digital wallet". 

I agree that that will eventually be in scope. Just trying to keep the problem space very constrained for now.

index.html
marcoscaceres, line 188:
```suggestion
      required DOMString request;
```

index.html
```diff
@@ -0,0 +1,302 @@
...
+      The `protocol` member
+    </h3>
+    <p>
+      The <dfn data-dfn-for="IdentityRequestProvider">protocol</dfn> member
+      denotes the [=digital identity/request protocol=] when requesting an
+      identify credential.
+    </p>
+    <p>
+      The {{IdentityRequestProvider/protocol}} member's value is be one of the
+      keys defined in [[[#protocol-registry]]].
```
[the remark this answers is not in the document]
marcoscaceres: Let's keep protocol as a DOMString because we are going to move the Registry to another document (which will make updating the enum painful once this is on the W3C Rec Track). We can add a note that it should be implemented as an enum, and mimic enum behavior via prose. We do the same thing in the Permission spec.

index.html
marcoscaceres: Tbh, I think we might go through a few more renames over the next few weeks.

index.html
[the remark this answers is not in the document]
samuelgoto: SGTM

index.html
samuelgoto: Agreed. I think Digital Identities is a great starting point. Resolving.

index.html
marcoscaceres: Let's run with this for now but let's definitely start taking definitions/concepts from other specs if we can. However, we should control whatever we feel strongly fall squarely in the purview of our spec.

index.html
marcoscaceres: We should maybe just note this issue and say that `providers` may change based on what you describe above.

index.html
```diff
@@ -0,0 +1,302 @@
...
+    <ul>
+      <li>Requesting a [=digital identity=], including mechanisms for secure
+      presentation.
+      </li>
+    </ul>
+    <p>
+      The following items are out of scope:
+    </p>
+    <ul>
+      <li>Enrollment process for establishing a [=digital identity=].
```
marcoscaceres, line 134:
```suggestion
      <li> Issuance process for establishing a [=digital identity=].
```

index.html
```diff
@@ -0,0 +1,302 @@
...
+    <ul>
+      <li>Requesting a [=digital identity=], including mechanisms for secure
+      presentation.
+      </li>
+    </ul>
+    <p>
+      The following items are out of scope:
+    </p>
+    <ul>
+      <li> Issuance process for establishing a [=digital identity=].
```
marcoscaceres, line 134:
```suggestion
      <li>Issuance process for establishing a [=digital identity=].
```

index.html
samuelgoto: Ok, I'm going to resolve this, since @bvandersloot-mozilla 's [alternative design](https://github.com/WICG/digital-credentials/issues/61#issuecomment-1921637318) really captures the essence of what I was struggling with, which is now reflected in this PR.

index.html
samuelgoto: Ok, I think we arrived at a good conclusion on https://github.com/WICG/digital-credentials/issues/61. This issue here is now resolved from where I stand, and since we are following @bvandersloot-mozilla 's suggestion, I think it would be consistent with his thoughts. @marcoscaceres is there anything else that you think is actionable here?

index.html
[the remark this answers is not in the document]
samuelgoto: Resolving this, since we can have this discussion elsewhere in an issue.

index.html
samuelgoto: I'm going to resolve this here to unblock this draft, so that we can resolve asynchronously here in this issue:

https://github.com/WICG/digital-credentials/issues/65#issuecomment-1924941840

index.html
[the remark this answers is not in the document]
samuelgoto: @marcoscaceres resolved this and changed to DOMString in this version of the spec. Resolving this.

index.html
```diff
@@ -0,0 +1,310 @@
...
+    </h2>
+    <p>
+      The <dfn>DigitalIdentity</dfn> interface represents a [=digital
+      identity=].
+    </p>
+    <pre class="idl">
+    [Exposed=Window, SecureContext]
+    interface DigitalIdentity : Identity {
+      readonly attribute DOMString protocol;
+      readonly attribute ArrayBuffer data;
```
samuelgoto, line 243:
```suggestion
      readonly attribute DOMString data;
```

samuelgoto (approved):
LGTM

There is a lot missing in this spec right now, but it contains a great baseline that we can work from!

Thanks for putting this together!

index.html
marcoscaceres: I'm happy to proceed. Let's do additional discussions in a follow up.

RByers:
retroactive LGTM as well, I like this direction. Thanks Marcos and Sam!

bc-pi:
Is the "rendered" version of this available and if so where? (sorry if that's an ignorant question - I am largely unfamiliar with w3c tooling and stuff)

samuelgoto:
https://wicg.github.io/digital-credentials/

bc-pi:
Thanks @samuelgoto !

marcoscaceres:
Renamed it to https://wicg.github.io/digital-identities/ ... to match the new spec title.

Issue #58, opened 2024-01-25
Registry inclusion criteria


marcoscaceres:
We need to come up with a registry governance and inclusion criteria. 

For inclusion, at a minimum, there should be implementation support, and we talked about having some privacy checks too.

OR13:
For some inspiration: https://datatracker.ietf.org/doc/html/rfc8126#section-4.6

marcoscaceres:
Thanks @OR13! Adding for reference: 
https://www.w3.org/2021/Process-20211102/#registries

Which defines the W3C requirements for a Registry. 

We can copy/:pasta: and adapt the text from the [Permissions Registry](https://w3c.github.io/permissions-registry/#change-process), which should be similar enough.

marcoscaceres:
Just from call on Feb 8:

 * selective disclosure 
 * encrypted response
 * Structure that is "web IDL" compatible.

npdoty:
Protocol needs to provide detail in the query such that the browser can understand what is a compatible credential and what are the specific data elements being requested, such that the UA can help the user to understand the request prior to selecting a wallet/credential.

npdoty:
Protocol needs to enable passing an indication to the user to explain the context of the request (who is requesting it, why, with what privacy protections, etc.), and/or the elements in the webpage that provided that information.

tplooker:
To expand on the last point from @marcoscaceres here which I believe was related to a point I raised. It would be a failure if say the response from the `.get()` API for a specific protocol expected additional out of band steps for the relying party to actually get the response, such as following a redirect or calling an API etc.

OR13:
Thinking more about this, I am not sure it makes sense to create a registry at all.

Instead it might be better to internalize a fixed set of supported protocols, and require a full document update in order to add further support.

If there is more than 1 protocol listed, there MUST be a single mandatory to implement.

samuelgoto:
> Thinking more about this, I am not sure it makes sense to create a registry at all.

The other thing that I think is worth considering is that the more protocols we have, the fewer interoperable implementations we'll have, and fragment the ecosystem. So, more here isn't necessarily better.

timcappalli:
From #43:
> support for selective disclosure could be mandated to be included in the registry for "protocol"

Sakurann:
> Thinking more about this, I am not sure it makes sense to create a registry at all.

[...]

PR #83, opened 2024-02-17
Align Model terminology with multiple digital credential ecosystems.

msporny:
This PR attempts to align the terminology used in the Data Model section with language that is compatible with multiple digital credential ecosystems.

***
<a href="https://pr-preview.s3.amazonaws.com/msporny/digital-identities/pull/83.html" title="Last updated on Mar 6, 2024, 11:33 PM UTC (6859ed9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/digital-identities/83/292a71a...msporny:6859ed9.html" title="Last updated on Mar 6, 2024, 11:33 PM UTC (6859ed9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -76,10 +76,11 @@
       Model
     </h2>
 
-    <p class="issue" title="Definitions under discussion">
-The goal of the definitions in this section is to re-use or establish
+    <p class="note" title="Definitions under discussion">
+The goal of the definitions in this section is to reuse or establish
 terminology that is common across a variety of digital credential formats and
-protocols.
+protocols. Discussions surrounding these definitions are active and
+the definitions are likely to change over the next several months.
     </p>
 
     <dl class="definitions" data-sort="" data-cite="vc-data-model-2.0">
--- a/index.html
+++ b/index.html
@@ -107,8 +107,7 @@ protocols.
         Software that provides a user interface for querying, selecting, and
         establishing consent for the [=presentation=] of a [=digital
         credential=]. Examples include a digital wallet, or a login manager,
-        that manages various
-        <a href="#credential-type-examples">credential types</a>.
+        that manages various [=example credential types|credential types=].
       </dd>
       <dt>
         <dfn>Presentation request</dfn>
--- a/index.html
+++ b/index.html
@@ -114,7 +114,7 @@ protocols.
         <dfn>Presentation request</dfn>
       </dt>
       <dd>
-        A format that a [=verifier=] uses, via an [=exchange protocol=] to
+        A format that a [=verifier=] uses, via an [=exchange protocol=], to
         request a [=digital credential=] from a [=credential provider=].
       </dd>
       <dt>
@@ -122,7 +122,7 @@ protocols.
       </dt>
       <dd>
         A format that a [=credential repository=] uses, via an
-        [=exchange protocol=] to respond to a [=presentation request=] by
+        [=exchange protocol=], to respond to a [=presentation request=] by
         a [=verifier=].
       </dd>
       <dt>
--- a/index.html
+++ b/index.html
@@ -125,7 +125,7 @@ protocols.
         a [=verifier=].
       </dd>
       <dt>
-        <dfn>Exchange protocol</dfn>
+        <dfn data-dfn-for="digital credential">Exchange protocol</dfn>
       </dt>
       <dd>
         A protocol used for exchanging a [=digital credential=] between a
--- a/index.html
+++ b/index.html
@@ -104,7 +104,7 @@ the definitions are likely to change over the next several months.
         protocol=], to request a [=digital credential=] from a [=holder=].
       </dd>
       <dt>
-        <dfn>Presentation</dfn>
+        <dfn>Presentation response</dfn>
       </dt>
       <dd>
         A format that a [=holder=] uses, via an [=digital
--- a/index.html
+++ b/index.html
@@ -31,15 +31,7 @@
       shortName: "digital-identity",
       specStatus: "CG-DRAFT",
       group: "wicg",
-      localBiblio: {
-        "ISO18013-5": {
-          title: "ISO/IEC 18013-5:2021 ISO-compliant driving licence, Part 5: Mobile driving licence (mDL) application",
-          authors: ["ISO/IEC JTC 1/SC 17"],
-          date: "September 2021",
-          publisher: "International Organization for Standardization",
-          href: "https://www.iso.org/standard/69084.html"
-        }
-      },
+      localBiblio: {},
       xref: {
         profile: "web-platform",
       },
--- a/index.html
+++ b/index.html
@@ -101,19 +101,20 @@ the definitions are likely to change over the next several months.
         that manages various [=example credential types|credential types=].
       </dd>
       <dt>
-        <dfn>Presentation request</dfn>
+        <dfn data-dfn-for="digital credential">query</dfn>
       </dt>
       <dd>
-        A format that a [=verifier=] uses, via an [=exchange protocol=], to
-        request a [=digital credential=] from a [=credential provider=].
+        A format that a [=verifier=] uses, via an [=digital credential/exchange
+        protocol=], to request a [=digital credential=] from a [=credential
+        provider=].
       </dd>
       <dt>
         <dfn>Presentation</dfn>
       </dt>
       <dd>
-        A format that a [=credential repository=] uses, via an
-        [=exchange protocol=], to respond to a [=presentation request=] by
-        a [=verifier=].
+        A format that a [=credential repository=] uses, via an [=digital
+        credential/exchange protocol=], to respond to a [=digital
+        credential/query=] by a [=verifier=].
       </dd>
       <dt>
         <dfn data-dfn-for="digital credential">Exchange protocol</dfn>
@@ -193,16 +194,17 @@ the definitions are likely to change over the next several months.
     </h3>
     <p>
       The <dfn data-dfn-for="DigitalCredentialRequestOptions">providers</dfn>
-      specify an [=exchange protocol=] and [=presentation request=],
-      which the user agent MAY match against a [=credential provider=].
+      specify an [=digital credential/exchange protocol=] and [=digital
+      credential/query=], which the user agent MAY match against a [=credential
+      provider=].
     </p>
     <h2>
       The `IdentityRequestProvider` dictionary
     </h2>
     <p>
-      The {{IdentityRequestProvider}} dictionary is used to specify an
-      [=exchange protocol=] and a [=presentation request=], which the user agent
-      MAY match against a [=credential provider=].
+      The {{IdentityRequestProvider}} dictionary is used to specify an [=digital
+      credential/exchange protocol=] and a [=digital credential/query=], which
+      the user agent MAY match against a [=credential provider=].
     </p>
     <pre class="idl">
     dictionary IdentityRequestProvider {
@@ -215,7 +217,7 @@ the definitions are likely to change over the next several months.
     </h3>
     <p>
       The <dfn data-dfn-for="IdentityRequestProvider">protocol</dfn> member
-      denotes the [=exchange protocol=] when requesting an
+      denotes the [=digital credential/exchange protocol=] when requesting an
       identify credential.
     </p>
     <p>
@@ -250,7 +252,7 @@ the definitions are likely to change over the next several months.
     </h3>
     <p>
       The <dfn data-dfn-for="DigitalCredential">protocol</dfn> member is the
-      [=exchange protocol=] that was used to request the
+      [=digital credential/exchange protocol=] that was used to request the
       [=digital credential=].
     </p>
     <h3>
@@ -264,7 +266,7 @@ the definitions are likely to change over the next several months.
       Registry of protocols for requesting digital credential
     </h2>
     <p>
-      The following is the registry of [=exchange protocols=]
+      The following is the registry of [=digital credential/exchange protocols=]
       that are supported by this specification.
     </p>
     <p class="note" title="Official Registry" data-cite="w3c-process">
@@ -279,11 +281,13 @@ the definitions are likely to change over the next several months.
     </p>
     <aside class="issue" data-number="58"></aside>
     <p>
-      [=User agents=] MUST support the following [=exchange protocols=]:
+      [=User agents=] MUST support the following [=digital credential/exchange
+      protocols=]:
     </p>
     <table class="data">
       <caption>
-        Table of officially registered [=exchange protocols=].
+        Table of officially registered [=digital credential/exchange
+        protocols=].
       </caption>
       <thead>
         <tr>
--- a/index.html
+++ b/index.html
@@ -92,7 +92,7 @@ the definitions are likely to change over the next several months.
         [=claims=] made by an [=issuer=] about one or more [=subjects=].
       </dd>
       <dt>
-        <dfn>Credential provider</dfn>
+        <dfn>Credential manager</dfn>
       </dt>
       <dd>
         Software that provides a user interface for querying, selecting, and
@@ -106,13 +106,13 @@ the definitions are likely to change over the next several months.
       <dd>
         A format that a [=verifier=] uses, via an [=digital credential/exchange
         protocol=], to request a [=digital credential=] from a [=credential
-        provider=].
+        manager=].
       </dd>
       <dt>
         <dfn>Presentation</dfn>
       </dt>
       <dd>
-        A format that a [=credential repository=] uses, via an [=digital
+        A format that a [=credential manager=] uses, via an [=digital
         credential/exchange protocol=], to respond to a [=digital
         credential/query=] by a [=verifier=].
       </dd>
@@ -121,7 +121,7 @@ the definitions are likely to change over the next several months.
       </dt>
       <dd>
         A protocol used for exchanging a [=digital credential=] between a
-        [=credential provider=] and a [=verifier=]. See section
+        [=credential manager=] and a [=verifier=]. See section
         [[[#protocol-registry]]].
       </dd>
     </dl>
@@ -186,17 +186,17 @@ the definitions are likely to change over the next several months.
     </h2>
     <pre class="idl">
     dictionary DigitalCredentialRequestOptions {
-      sequence&lt;IdentityRequestProvider&gt; providers;
+      sequence&lt;IdentityRequestProvider&gt; managers;
     };
     </pre>
     <h3>
-      The `providers` member
+      The `managers` member
     </h3>
     <p>
-      The <dfn data-dfn-for="DigitalCredentialRequestOptions">providers</dfn>
+      The <dfn data-dfn-for="DigitalCredentialRequestOptions">managers</dfn>
       specify an [=digital credential/exchange protocol=] and [=digital
       credential/query=], which the user agent MAY match against a [=credential
-      provider=].
+      manager=].
     </p>
     <h2>
       The `IdentityRequestProvider` dictionary
@@ -204,7 +204,7 @@ the definitions are likely to change over the next several months.
     <p>
       The {{IdentityRequestProvider}} dictionary is used to specify an [=digital
       credential/exchange protocol=] and a [=digital credential/query=], which
-      the user agent MAY match against a [=credential provider=].
+      the user agent MAY match against a [=credential manager=].
     </p>
     <pre class="idl">
     dictionary IdentityRequestProvider {
@@ -231,7 +231,7 @@ the definitions are likely to change over the next several months.
     <p>
       The <dfn data-dfn-for="IdentityRequestProvider">request</dfn> member is
       the request to be handled by the user's selected [=credential
-      provider=].
+      manager=].
     </p>
     <h2>
       The `DigitalCredential` interface
--- a/index.html
+++ b/index.html
@@ -107,9 +107,9 @@ the definitions are likely to change over the next several months.
         <dfn>Presentation response</dfn>
       </dt>
       <dd>
-        A format that a [=holder=] uses, via an [=digital
-        credential/exchange protocol=], to respond to a [=digital
-        credential/query=] by a [=verifier=].
+        A format that a [=holder|holder's=] software, such as a digital wallet,
+        uses, via an [=digital credential/exchange protocol=], to respond to a
+        [=digital credential/query=] by a [=verifier=].
       </dd>
       <dt>
         <dfn data-dfn-for="digital credential">Exchange protocol</dfn>
--- a/index.html
+++ b/index.html
@@ -90,6 +90,11 @@ the definitions are likely to change over the next several months.
       <dd>
         A cryptographically signed digital document containing one or more
         [=claims=] made by an [=issuer=] about one or more [=subjects=].
+
+        <p class="note" title="Focus on digital credentials about people">
+          This specification is focused on digital credentials pertaining to
+          people.
+        </p>
       </dd>
       <dt>
         <dfn>Credential manager</dfn>
--- a/index.html
+++ b/index.html
@@ -88,7 +88,7 @@ the definitions are likely to change over the next several months.
         <dfn>Digital credential</dfn>
       </dt>
       <dd>
-        A cryptographically secured digital document containing one or more
+        A cryptographically signed digital document containing one or more
         [=claims=] made by an [=issuer=] about one or more [=subjects=]. Examples include
         ISO mobile driver's licenses [[ISO18013-5]] and W3C [=verifiable
         credentials=] [[VC-DATA-MODEL-2.0]].
--- a/index.html
+++ b/index.html
@@ -32,6 +32,13 @@
       specStatus: "CG-DRAFT",
       group: "wicg",
       localBiblio: {
+        "ISO18013-5": {
+          title: "ISO/IEC 18013-5:2021 ISO-compliant driving licence, Part 5: Mobile driving licence (mDL) application",
+          authors: ["ISO/IEC JTC 1/SC 17"],
+          date: "September 2021",
+          publisher: "International Organization for Standardization",
+          href: "https://www.iso.org/standard/69084.html"
+        }
       },
       xref: {
         profile: "web-platform",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -67,30 +74,55 @@ <h2>
...
-        <p>
-          Is a [=verifiable credential=] about a person.
-        </p>
+        A cryptographically secured digital document containing one or more
+        [=claims=] made by an [=issuer=] about a [=subject=]. Examples include
+        ISO mobile driver's licenses [[ISO18013-5]] and W3C [=verifiable
+        credentials=] [[VC-DATA-MODEL-2.0]].
+      </dd>
+      <dt>
+        <dfn>Credential provider</dfn>
```
msporny, line 95: Alternates include:

* Credential store (which is what CREDMAN calls this)
* Credential repository (which is what VCDM calls this)

index.html
```diff
@@ -67,30 +74,55 @@ <h2>
...
+    <dl class="definitions" data-sort="" data-cite="vc-data-model-2.0">
       <dt>
         <dfn>Digital credential</dfn>
       </dt>
       <dd>
-        <p>
-          Is a [=verifiable credential=] about a person.
-        </p>
+        A cryptographically secured digital document containing one or more
+        [=claims=] made by an [=issuer=] about a [=subject=]. Examples include
```
OR13, line 90: VCDM says multiple subjects are allowed in a credential.

TallTed:
```suggestion
        [=claims=] made by an [=issuer=] about one or more [=subjects=]. Examples include
```

index.html
```diff
@@ -162,17 +194,16 @@ <h3>
...
     </p>
     <h2>
       The `IdentityRequestProvider` dictionary
     </h2>
     <p>
-      The {{IdentityRequestProvider}} dictionary is used to specify a [=digital
-      identity/request protocol=] and structured request, which the user agent
-      MAY match against a [=identity credential provider=].
+      The {{IdentityRequestProvider}} dictionary is used to specify an
+      [=exchange protocol=] and [=presentation request=], which the user agent
```
TallTed, line 205:
```suggestion
      [=exchange protocol=] and a [=presentation request=], which the user agent
```

TallTed (changes requested):

index.html
msporny: Fixed in 002d8ae2eb75b88d60b2e7d3b44ac5c05ddcf397

index.html
```diff
@@ -32,6 +32,13 @@
       specStatus: "CG-DRAFT",
       group: "wicg",
       localBiblio: {
+        "ISO18013-5": {
```
marcoscaceres, line 35: FWIW: https://github.com/tobie/specref/issues/776 ... we might want to just send this to specref, as I've not managed to get an answer as to why it's not showing up in SpecRef... 

Also, we purposely avoided mentioning mDoc until we get things into the registry 🙈

index.html
```diff
@@ -67,30 +74,55 @@ <h2>
     <h2>
       Model
     </h2>
-    <dl class="definitions" data-sort="" data-cite="vc-data-model">
+
+    <p class="issue" title="Definitions under discussion">
+The goal of the definitions in this section is to re-use or establish
+terminology that is common across a variety of digital credential formats and
+protocols.
+    </p>
```
marcoscaceres, line 82: Let's live dangerously... but seriously, we are ok to just define things within the W3C and our own ecosystem. 

```suggestion
```

index.html
```diff
@@ -67,30 +74,55 @@ <h2>
...
+
+    <dl class="definitions" data-sort="" data-cite="vc-data-model-2.0">
       <dt>
         <dfn>Digital credential</dfn>
       </dt>
       <dd>
-        <p>
-          Is a [=verifiable credential=] about a person.
-        </p>
+        A cryptographically secured digital document containing one or more
```
marcoscaceres, line 89: I'm uncomfortable with the with the word "secured" here... I'm not sure cryptographically necessarily means something is secure. 
```suggestion
        A cryptographically authenticated digital document containing one or more
```

index.html
marcoscaceres: Yeah, we wanted something that covers basically these things being "remote" too (i.e., you can pull them from the browser, or multiple wallet apps, etc.)... so it's kinda like a manager?

index.html
```diff
@@ -67,30 +74,55 @@ <h2>
...
+        [=claims=] made by an [=issuer=] about one or more [=subjects=]. Examples include
+        ISO mobile driver's licenses [[ISO18013-5]] and W3C [=verifiable
+        credentials=] [[VC-DATA-MODEL-2.0]].
```
marcoscaceres, line 92: I'd be hesitant to just say "[=verifiable credentials=]" as it's overly broad, and covers "all the things"... we are just dealing with digital documents that have attributes/claims relating to people (for now).

```suggestion
        [=claims=] made by an [=issuer=] about one or more [=subjects=]. Examples include
        ISO mobile driver's licenses such as those specified in [[?ISO18013-5]].
```

index.html
```diff
@@ -67,30 +74,55 @@ <h2>
...
+      </dd>
+      <dt>
+        <dfn>Credential provider</dfn>
+      </dt>
+      <dd>
+        Software that provides a user interface for querying, selecting, and
+        establishing consent for the [=presentation=] of a [=digital
+        credential=]. Examples include a digital wallet, or a password manager,
+        that manages various
+        <a href="#credential-type-examples">credential types</a>.
```
marcoscaceres, line 102: Nit: we might want to `<dfn>` the "example credential types" just to avoid `<a href="">` to them.

index.html
```diff
@@ -67,30 +74,55 @@ <h2>
...
       <dd>
-        An application or service that provides a user interface for selecting
-        and/or querying a [=digital credential=], such as a digital wallet that
-        manages various identity documents and credentials.
+        A format that a [=verifier=] uses, via an [=exchange protocol=], to
+        request a [=digital credential=] from a [=credential provider=].
       </dd>
       <dt>
-        <dfn data-for="digital credential">Request protocol</dfn>
+        <dfn>Presentation</dfn>
```
marcoscaceres, line 112: Maybe: 

```suggestion
        <dfn>Presentation response</dfn>
```

index.html
```diff
@@ -67,30 +74,55 @@ <h2>
...
+        [=exchange protocol=], to respond to a [=presentation request=] by
+        a [=verifier=].
+      </dd>
+      <dt>
+        <dfn>Exchange protocol</dfn>
+      </dt>
+      <dd>
+        A protocol used for exchanging a [=digital credential=] between a
+        [=credential provider=] and a [=verifier=]. See section
+        [[[#protocol-registry]]].
```
marcoscaceres, line 125: Maybe 
```suggestion
        Standardized data structures for [=presentation request|requesting=] and [=presentation response|receiving=] 
        a [=digital credential=] between a [=credential provider=] and a [=verifier=]. See section
        [[[#protocol-registry]]] for supported protocols.
```

index.html
```diff
@@ -67,30 +74,55 @@ <h2>
...
       <dd>
-        A standardized format for requesting a [=digital credential=], designed
-        to ensure the secure, private, and interoperable exchange of identity
-        information. See section [[[#protocol-registry]]].
+        A format that a [=credential repository=] uses, via an
+        [=exchange protocol=], to respond to a [=presentation request=] by
+        a [=verifier=].
+      </dd>
+      <dt>
+        <dfn>Exchange protocol</dfn>
```
marcoscaceres, line 120: We probably want to make these less generalized... so, probably: 

```suggestion
        <dfn data-dfn-for="digital credential">Exchange protocol</dfn>
```

index.html
```diff
@@ -67,30 +74,55 @@ <h2>
...
+      <dd>
+        Software that provides a user interface for querying, selecting, and
+        establishing consent for the [=presentation=] of a [=digital
+        credential=]. Examples include a digital wallet, or a password manager,
+        that manages various
+        <a href="#credential-type-examples">credential types</a>.
       </dd>
       <dt>
-        <dfn>Identity credential provider</dfn>
+        <dfn>Presentation request</dfn>
```
marcoscaceres, line 105: Maybe we don't need this... "exchange protocol" could cover both request and response. WDYT?

index.html
msporny: Agreed that it belongs in SpecRef, I searched for it there first and didn't find it, so added it manually. 

We can use the SpecRef reference when it shows up there.

index.html
msporny: @marcoscaceres, you rebel, you. :)

I was trying to convey to readers, as this document matures,  that the goal is to align w/ multiple ecosystems and not re-define yet another set of terminology that is inconsistent. I'm fine w/ removing this if others want it gone, but do think it has value to convey that we're still working to align terminology and will continue to strive towards that goal.

Looking for more opinions wrt. revising or removing this block.

index.html
OR13: https://datatracker.ietf.org/doc/html/rfc4949 has some terminology you might consider:

```
credential
      1. (I) /authentication/ "identifier credential": A data object
      that is a portable representation of the association between an
      identifier and a unit of authentication information, and that can
      be presented for use in verifying an identity claimed by an entity
      that attempts to access a system. Example: X.509 public-key
      certificate. (See: anonymous credential.)
```

index.html
msporny: Yes, agreed that "manager" is probably the better term. CREDMAN uses it in a slightly different way (AND doesn't define it)... 

https://www.w3.org/TR/credential-management-1/#introduction

"Credential manager" in the CREDMAN usage of the term ties the "credential manager" to the user agent, often referring to it as "the user agent's credential manager". For this API, the two aren't tightly coupled, where the "credential manager" (digital wallet) is typically a completely separate application from the "user agent" (web browser).

index.html
msporny: Fixed in: ab95b494064e4cb7df7e601f74ce5d4c39695989

index.html
msporny: The term doesn't get into how its represented, but media types are one way that we could keep the flexibility in the API to be able to add different presentation formats over time.

index.html
msporny: Hmm, there is more to a protocol than standardized data structures, no? What about this?

```suggestion
        A protocol used for [=presentation request|requesting=] and [=presentation response|receiving=] 
        a [=digital credential=] between a [=credential provider=] and a [=verifier=]. See section
        [[[#protocol-registry]]] for supported protocols.
```

index.html
Sakurann:
```suggestion
        A cryptographically signed digital document containing one or more
```

index.html
```diff
@@ -64,33 +71,60 @@ <h2>
...
       </dd>
       <dt>
-        <dfn>Identity credential provider</dfn>
+        <dfn>Presentation</dfn>
       </dt>
       <dd>
-        An application or service that provides a user interface for selecting
-        and/or querying a [=digital credential=], such as a digital wallet that
-        manages various identity documents and credentials.
+        A format that a [=credential repository=] uses, via an
```
Sakurann, line 117: presentation is not a format.

index.html
Sakurann: +1 to removing presentation request.

If the intention is to separate three things you list, manu, i think it is cleaner to use terms `query language`, `credential format`, `exchange protocol` respectively.

index.html
Sakurann: per comment above +1 to removing presentation,  and instead define terms `query language`, `credential format`, `exchange protocol`, which seems to be closer to the intent here.

Sakurann (changes requested):
made suggestions in the comments, but overall there is a room to improve the current set of terminology proposed in this PR to better reflects the most commonly used terms

index.html
marcoscaceres: Yeah, I'll send a PR if the above doesn't get fixed... actually, I'll just send it now...

marcoscaceres: https://github.com/tobie/specref/pull/779

index.html
```diff
@@ -32,6 +32,13 @@
       specStatus: "CG-DRAFT",
       group: "wicg",
       localBiblio: {
+        "ISO18013-5": {
+          title: "ISO/IEC 18013-5:2021 ISO-compliant driving licence, Part 5: Mobile driving licence (mDL) application",
+          authors: ["ISO/IEC JTC 1/SC 17"],
+          date: "September 2021",
+          publisher: "International Organization for Standardization",
+          href: "https://www.iso.org/standard/69084.html"
+        }
```
marcoscaceres, line 41:
```suggestion
```

index.html
marcoscaceres: Oh! "singed" is spot on, @Sakurann! Let's go with that.

index.html
marcoscaceres: > I am also hesitant regarding W3C [=verifiable credentials=]. apple-to-apple comparison here would be mdocs and w3c vcdm and ietf sd-jwt vc

@Sakurann agree. That's exactly how I was feeling about it. 

I think it might be better to drop the example specs until we have the Registry #85 #58... we can add them once we agree on the mandated formats. 

```suggestion
        [=claims=] made by an [=issuer=] about one or more [=subjects=].
```

index.html
marcoscaceres: "Wallet" feels very concrete and product...y..., but it's definitely the right concept. 

However, I'm hesitant to use "wallet" because it's a loaded from a UI perspective. 

I guess what we'd like to convey here is more around the "role" that this bit of middleware plays... like, it may be a piece of software that acts as a wallet, but also allows you to select other wallets... plus coordinates the flow of data between itself and other wallets (as well as presenting UI).

index.html
marcoscaceres: I'd be hesitant to say "query language", as it seems to imply something like SQL. We are just passing a structure around, so, maybe, "query structure", or, just "query" (remember that in context it will be `[=digital credential/query=]`).

index.html
marcoscaceres: Yeah, "credential format" feels right here. And the format could have a MIME type or whatever... we could definitely add that. Note that the request protocol is always part of the response in the API, so you kinda know what you are going to get... however, it might be nice to have the MIME type if a query might result in multiple credential formats.

index.html
marcoscaceres: > Hmm, there is more to a protocol than standardized data structures, no?

Yeah, you are definitely not wrong, but protocol is somewhat incomplete... the challenge with the formats is that we need to convert them to something that browsers can ingest, and it's unlikely those formats will define (JS compatible) error handling behavior, Web IDL conversions, etc. 

So, what I was trying to convey is that the request can only be a data structure, the browser will do a bunch of security checks and conversion on that data structure (strip out any garbage, reject anything that looks suspect, etc.) and only then will it pass that to the platform.

index.html
marcoscaceres: Ok, to keep it, but change it to a "note" instead of an issue.

index.html
msporny: Ok, I'll drop the local ref presuming tobie/specref#779 will go in soon.

index.html
msporny: Changed to NOTE in 6dd3f57d6893649d848fda10077e2b840d89dad8.

index.html
msporny: I've made an attempt at using `query` in cd3dc633f50cf30e27de9474e919ea354a822de1.

FWIW, "query" is what the Verifiable Presentation Request specification uses: https://w3c-ccg.github.io/vp-request-spec/#query-by-example

index.html
msporny: > I guess what we'd like to convey here is more around the "role" that this bit of middleware plays... like, it may be a piece of software that acts as a wallet, but also allows you to select other wallets... plus coordinates the flow of data between itself and other wallets (as well as presenting UI).

Sounds like a "mediator" to me, but "credential mediator" is the role the browser is playing, not the role the wallet is playing?

"Wallet" is not at the right conceptual level... a wallet is one form of this role, but it's not the only form.

I'll give "credential manager" a shot if we think that "credential provider" isn't going to work out.

Updated in 964c9c410943caf65e2d1e634b892d2d637c1afc.

index.html
msporny: Hrm, unfortunately, "credential format" typically only covers the format for the original credential... the original "credential format" can then be derived into another credential/presentation or wrapped (to counter-sign the original credential). "Credential format" doesn't cover the response use cases.

index.html
msporny: Right, but then what do we call OID4 or VC API or the protocols that define more than just "standardized data structures"?

index.html
```diff
@@ -64,33 +63,60 @@ <h2>
...
+        Software that provides a user interface for querying, selecting, and
+        establishing consent for the [=presentation=] of a [=digital
+        credential=]. Examples include a digital wallet, or a login manager,
+        that manages various [=example credential types|credential types=].
+      </dd>
+      <dt>
+        <dfn data-dfn-for="digital credential">query</dfn>
+      </dt>
+      <dd>
+        A format that a [=verifier=] uses, via an [=digital credential/exchange
```
samuelgoto, line 101:
```suggestion
        A format that a [=verifier=] or [=user agent=] uses, via an [=digital credential/exchange
```

So, one of the big architectural elements here is that, at least currently in Chrome/Android, the "user agent" (the browser or the operating system), does the credential selection before handing the request to the holder. So, the "query" is actually *also* used by the [=user agent=] to filter out which holders would be able to answer the query.

index.html
```diff
@@ -64,33 +63,60 @@ <h2>
...
       </dd>
       <dt>
-        <dfn>Identity credential provider</dfn>
+        <dfn>Presentation</dfn>
       </dt>
       <dd>
-        An application or service that provides a user interface for selecting
-        and/or querying a [=digital credential=], such as a digital wallet that
-        manages various identity documents and credentials.
+        A format that a [=credential manager=] uses, via an [=digital
```
samuelgoto, line 109:
```suggestion
        A format that a [=holder=] uses, via an [=digital
```

Isn't the response produced by the holder?

index.html
samuelgoto: > I'll give "credential manager" a shot if we think that "credential provider" isn't going to work out.

"manager" is highly overloaded in Web Platform, and "credential manager", specifically means a lot of other things. "credential provider" isn't great, but I think is less ambiguous than "credential manager".

But also, I think we are missing one big architectural element here (speaking only for the choices that Android has made so far), which is that the [=user agent=] has a "registry" of all of the credentials from [=holders=] which it uses to build a cross-wallet credential selector. That is, I think we need to find a way to introduce this level of aggregation across wallets, because that's what the JS API "talks to", rather than directly to the actual [=holders=] up front.

Am I making sense?

index.html
samuelgoto: "response" is also an "instance" (of specific bytes) rather than a "class" (of formats) right? that is, "response" isn't a "format", it is a "series of bytes encoded in a specific format", right?

index.html
marcoscaceres: Yeah, that makes sense @samuelgoto. I think you articulated it well.

index.html
marcoscaceres: Good point!

index.html
marcoscaceres: Correct.

index.html
marcoscaceres: I guess, yes... I was thinking the holder was a human... but, you are right that the holder is just the software.

index.html
msporny: >  "credential provider" isn't great, but I think is less ambiguous than "credential manager".

*lol* -- alright, back to "credential provider" it is! I'll make that change in an upcoming commit. Having a list of "managers" in the WebIDL felt off to me, but at least we tried it out and decided it's a direction we didn't want to go in.

> That is, I think we need to find a way to introduce this level of aggregation across wallets, because that's what the JS API "talks to", rather than directly to the actual [=holders=] up front. Am I making sense?

Yes, I think I get the general concern you're highlighting. Ignore the words I'm using for now, just trying to identify concepts... I believe you're identifying the "matching services" (the sandboxed WASM code), provided by each "credential provider" (digital wallet).

I was trying to avoid talking about those concepts in this PR (don't want the PR to add so much that it becomes a permathread and never gets merged because it's trying to do too much). Would it be okay if we introduced these concepts in a second PR? It seems orthogonal to the concept of a "credential provider"?

index.html
msporny: Yes, right, I'm going to merge your change suggestion in @samuelgoto, but wanted to make a distinction before doing so.

In the many years that led up to W3C Verifiable Credentials, we found that it was useful to separate the "roles" in the ecosystem (e.g. "user agent") from the software that performs actions on behalf of those roles (e.g., "user agent").

The roles that seem to be common to all these digital credential ecosystems are: issuer, holder, verifier.

Each of those roles uses software to achieve their goals in the ecosystem. Issuers (role) use issuance software (software agent) to issue verifiable credentials. Holders (role) use digital wallets (software agent) to manage verifiable credentials. Verifiers (role) use verification software (software agent) to verify verifiable credentials. A single entity can play many roles in a particular workflow -- for example, an organization might request prerequisite VCs as a verifier before it issues a subsequent VC that is dependent on the prerequisite VCs.

That approach has served us well over the past decade, and it's not a new concept, most other software ecosystems are explained in this manner.

All this to say, we shouldn't conflate the "verifier" (role) with the "user agent" (software agent). So, if we have agreement on the above, your change suggestion would switch to the following, @samuelgoto:

```suggestion
        A format that [=verifier=] software or a [=user agent=] uses, via an [=digital credential/exchange
```

I know the change feels pedantic, but we've always gotten ourselves into trouble in the other groups when we weren't this precise with our language.

Thoughts on the above before I commit the update?

index.html
msporny: Hrm, no, I'm with @marcoscaceres on this one. The "holder" is a role that a human can play. Note that it doesn't need to be a human, it can be any entity that has agency such as an organization or AI -- which we don't need to talk about in this spec 'cause it'll just confuse people.

That holder uses software to achieve their goals, where the software is a "credential manager/provider" AND a "user agent" (they use both via this API) to achieve a specific workflow.

IOW, the response is provided by software, so we need to name that software. I do think "credential manager/provider" is the right term here. Thoughts, @samuelgoto?

index.html
marcoscaceres: FWIW, @msporny's (...dare I say it...) "ontology" 🙈 of people's roles and software agents is also the one that I've been subscribing to.

msporny: >  "ontology" 🙈

booo! Throw that heretic into the streets! There's no room for that sort of language in the digital credentials spec! :stuck_out_tongue_winking_eye:

samuelgoto (commented):
Taking another pass and asking clarifying questions ... apologies if it is taking me a while to catch up to these definitions, please be patient while we all try to find common ground! :)

index.html
```diff
@@ -64,33 +63,60 @@ <h2>
...
+    <dl class="definitions" data-sort="" data-cite="vc-data-model-2.0">
       <dt>
         <dfn>Digital credential</dfn>
       </dt>
       <dd>
-        <p>
-          Is a [=verifiable credential=] about a person.
-        </p>
+        A cryptographically signed digital document containing one or more
+        [=claims=] made by an [=issuer=] about one or more [=subjects=].
```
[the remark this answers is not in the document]
TallTed: Could a digital credential not be a [=verifiable credential=] about...
* a company?
* a pedigreed pet? 
* a provenanced antique or artwork? 
* a vehicle or other titled or registered asset?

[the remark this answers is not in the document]
marcoscaceres: To @samuelgoto point, we were purposely trying to restrict this to things about people. It might be good to go back to that and then broaden it in the future as we gain more implementation experience.

index.html
```diff
@@ -64,33 +63,60 @@ <h2>
...
+      </dd>
+      <dt>
+        <dfn>Credential manager</dfn>
```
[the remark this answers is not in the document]
marcoscaceres: Yeah, that's a good point... maybe we don't need to define credential manager at all... or, we ask Cred Man to define it more formally and link to it.  As the credential manager itself deals so much with UI, I'd be ok to leave it undefined.

index.html
marcoscaceres: (see suggestion about dropping "credential manager" entirely... that way we just side step this whole discussion, same as Cred Man does)

index.html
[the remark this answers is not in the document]
msporny: > do you feel that the original definition ("a digital credential is a [=verifiable credential=] about a person") is incorrect? 

Yes, as @TallTed mentioned, it's incorrect because "digital credential" and "identity credential" are about more than just people. 

We could say "a **personal** credential is a [=verifiable credential=] about a person"... but then, like "identity credentials", if we start naming the WebIDL and parameters in the API around the concept of a "person" or "identity", we paint ourselves into a corner wrt. other digital credential types in the future (which we know there is interest in from the digital credentials ecosystem).

We could also say that "this specification is currently scoped to verifiable credentials related to people"?

I've made that change in 38139921ddc3b46b54051134b2674fbe4ab9a149. This allows us to focus on people in the spec, but allow for expansion into other things in the future. WDYT, @marcoscaceres, @samuelgoto, @TallTed ?

index.html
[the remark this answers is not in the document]
msporny: I've removed the definition of "Credential Manager" in 08bd487317c6986b52fbd3ce4137fc2e955c0130 to try out the idea. Initial thoughts:

* It is possible to express what's in the specification currently using the term "holder" and "holder's software, such as a digital wallet".
* The language is a bit awkward in places w/o a specific term to use when discussing the concept of a "credential manager".

I think we can go with this for now, but we probably do need to talk about the concept of the browser mediating and the app/website that the request is finally handed off to. AS @marcoscaceres mentioned, maybe we just ask CredMan to define it and refer to that?

index.html
msporny: Done in 9e8500ef18e3f42c349d7e0652f25d5fdb7d77b0.

index.html
msporny: Fixed in b080af4b378f32f2f9ae2add76698dc03652cab7 by separating the concept of a "holder" from "holder's software" and adding an example of "digital wallet" for "holder's software".

index.html
msporny: I'm sorry @samuelgoto, I wasn't trying to disparage your good name! I was only trying to disparage @marcoscaceres! :P

@samuelgoto's snippet applied in e162c9e2502266508891e99b58603c100980e4df.

msporny:
I've applied your suggestions @samuelgoto and @marcoscaceres. Please re-review.

There are two remaining questions, one with a committed suggestion and another with a change suggestion that needs your feedback, @marcoscaceres: 

* https://github.com/WICG/digital-identities/pull/83#discussion_r1501922315
* https://github.com/WICG/digital-identities/pull/83#discussion_r1510007100

index.html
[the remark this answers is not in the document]
TallTed: "[This specification is focused on digital credentials pertaining to people.](https://github.com/WICG/digital-identities/commit/38139921ddc3b46b54051134b2674fbe4ab9a149)" is not quite "this specification is _currently_ scoped to verifiable credentials related to people". "Currently" is key to this statement.

I would be mostly OK with "VCDMv2 is scoped to digital credentials related to people; future versions are expected to broaden their scope to include digital credentials related to other entity types."

That said, please note that the [Traceability Vocabulary](https://w3c-ccg.github.io/traceability-vocab/) and [Traceability Interop](https://github.com/w3c-ccg/traceability-interop) Work Items of the [Credentials Community Group](https://www.w3.org/groups/cg/credentials/) is focused entirely on VCs related to non-people. Changing the scope of the VCDMv2 as described above would appear to make the Traceability Task Force's output invalid until VCDMv3, or only valid with VCs based on VCDMv1. Both of these are troublesome, to my thinking.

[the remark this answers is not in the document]
TallTed: D'oh! Sorry. Too many interleaved specs on my workbench!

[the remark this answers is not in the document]
TallTed: (That said, "currently" is what was [described here](https://github.com/WICG/digital-identities/pull/83#discussion_r1510007100), but not what was implemented in https://github.com/WICG/digital-identities/commit/38139921ddc3b46b54051134b2674fbe4ab9a149. I think "currently" or similar phrasing should be applied.)

index.html
```diff
@@ -76,30 +75,49 @@ <h2>
...
         <dfn>Digital credential</dfn>
       </dt>
       <dd>
-        <p>
-          Is a [=verifiable credential=] about a person.
+        A cryptographically signed digital document containing one or more
+        [=claims=] made by an [=issuer=] about one or more [=subjects=].
+
+        <p class="note" title="Focus on digital credentials about people">
+          This specification is focused on digital credentials pertaining to
```
marcoscaceres, line 95:
```suggestion
          This specification is currently focused on digital credentials pertaining to
```

samuelgoto (approved):
This isn't quite perfect, but I think is a great starting point and baseline that we can work from! Thank you so much for putting this together @msporny and helping us find the words to express the main concepts!

index.html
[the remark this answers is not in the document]
tplooker: Also are we considering credentials that describe multiple subjects in the same credential here, as that appears what the definition implies? I know VC DM supports this, but it does bring about a whole different set of complexities and most other credential formats such as SD-JWT and mDocs have stayed away from this. I'd prefer we limit to one subject

```suggestion
        [=claims=] made by an [=issuer=] about one or more [=subject=].
```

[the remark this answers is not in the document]
msporny: A driving license is about multiple subjects: the issuer, the driver, and the vehicle classes the driver is allowed to operate.

msporny:
Applied your final "currently" change as well @marcoscaceres.

marcoscaceres (approved):

marcoscaceres:
Thanks again @msporny! You're a gem 💎!

PR #89, opened 2024-03-02
Add scoping language related to matching and consent.

### LINKED ISSUE #86: Nothing is learned without consent
Was asked on the call today to create an issue for folks to agree or disagree to this statement:

"When making the API call, the website should not learn anything about the user's credentials or wallets unless the user agent has gained explicit user consent. Likewise wallet applications will not learn anything about a given request unless the user explicitly consents to select a credential from that wallet"

 This principle may put certain constraints on things like silent enumerate APIs or feature detection APIs. It likely also places a constraint on what information is required to be passed in the initial request and how wallets perform matching.

msporny:
Closes #86 by asserting scoping around matching, consent, and what is learned by a website and 3rd party software.

***
<a href="https://pr-preview.s3.amazonaws.com/msporny/digital-identities/pull/89.html" title="Last updated on Aug 26, 2024, 10:40 PM UTC (de6afd4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/digital-credentials/89/292a71a...msporny:de6afd4.html" title="Last updated on Aug 26, 2024, 10:40 PM UTC (de6afd4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -116,7 +116,7 @@
       anything about the a holder's [=digital credentials=] or their software
       unless the [=user agent=] has previously received user consent.
       </li>
-      <li>Ensuring that any non-[=user agent=] software will not learn anything
+      <li>Ensuring that any installed application software will not learn anything
       about a given [=request=] unless the [=holder=] explicitly consents to use that
       software.
       </li>
--- a/index.html
+++ b/index.html
@@ -114,10 +114,10 @@
       </li>
       <li>Ensuring that when an API call is made, the website does not learn
       anything about the a holder's [=digital credentials=] or their software
-      unless the [=user agent=] has gained explicit user consent.
+      unless the [=user agent=] has express user consent.
       </li>
       <li>Ensuring that any non-[=user agent=] software will not learn anything
-      about a given request unless the holder explicitly consents to use the
+      about a given [=request=] unless the [=holder=] explicitly consents to use that
       software.
       </li>
     </ul>
--- a/index.html
+++ b/index.html
@@ -112,6 +112,14 @@
       <li>Requesting a [=digital credential=], including mechanisms for secure
       presentation.
       </li>
+      <li>Ensuring that when an API call is made, the website does not learn
+      anything about the a holder's [=digital credentials=] or their software
+      unless the [=user agent=] has gained explicit user consent.
+      </li>
+      <li>Ensuring that any non-[=user agent=] software will not learn anything
+      about a given request unless the holder explicitly consents to use the
+      software.
+      </li>
     </ul>
     <p>
       The following items are out of scope:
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -112,6 +112,14 @@ <h2>
       <li>Requesting a [=digital credential=], including mechanisms for secure
       presentation.
       </li>
+      <li>Ensuring that when an API call is made, the website does not learn
+      anything about the a holder's [=digital credentials=] or their software
+      unless the [=user agent=] has gained explicit user consent.
```
marcoscaceres, line 117: Maybe better:

```suggestion
      unless the [=user agent=] has express user consent.
```

aligns a bit more with permissions API's [express permission](https://www.w3.org/TR/permissions/#dfn-express-permission)

index.html
```diff
@@ -112,6 +112,14 @@ <h2>
...
+      </li>
+      <li>Ensuring that any non-[=user agent=] software will not learn anything
+      about a given request unless the holder explicitly consents to use the
```
marcoscaceres, line 120:
```suggestion
      about a given [=request=] unless the [=holder=] explicitly consents to use the
```

index.html
```diff
@@ -112,6 +112,14 @@ <h2>
       <li>Requesting a [=digital credential=], including mechanisms for secure
       presentation.
       </li>
+      <li>Ensuring that when an API call is made, the website does not learn
+      anything about the a holder's [=digital credentials=] or their software
+      unless the [=user agent=] has gained explicit user consent.
+      </li>
+      <li>Ensuring that any non-[=user agent=] software will not learn anything
```
marcoscaceres, line 119: This one is little ambiguous.

msporny: I was trying to generalize @leecam's notion of "wallet software". I'm wondering if we should call these things "native apps"? ... but then that leaves out "web apps", which could also hold/provide digital credentials. I think we do mean "any software that is not the user agent", but agree that the language is a bit unwieldy.

OR13: I would avoid the term native as a prefix, based on comments of I seen on inclusive language elsewhere.

Platform specific is probably a better term.

msporny: What about this?

```suggestion
      <li>Ensuring that any platform-specific software will not learn anything
```

OR

```suggestion
      <li>Ensuring that any installed application software will not learn anything
```

/cc @samuelgoto

msporny: Hmm, what about "software that is not involved in the consent-seeking process"? (though that's an unfortunate mouthful).

What we're trying to say here is that "The request is supposed to start off super secret and is only to be exposed to the holder via the software that they use to signal consent (the browser). After that point, the software that they chose to receive the request, and only that piece of software, is supposed to receive the request."

samuelgoto (approved):
LGTM++ with the diff

marcoscaceres (approved):
lgtm... sorry for the delay @msporny

PR #98, opened 2024-04-16
Add initial explainer

RByers:
A first-cut (draft) of an initial explainer trying to cover the key requirements and design elements. This is just a starting place for further discussion.


RByers:
Ignore, see #99 instead

PR #129, opened 2024-06-21
Introduce DiscoverFromExternalSource activation + visibility requirements

### LINKED ISSUE #91: Consume transient activation
We should probably require to [consume transient activation](https://html.spec.whatwg.org/multipage/interaction.html#consume-user-activation) to call identity.get(), as it shows UI. 

Plus it would be otherwise annoying if sites could just pop up this dialog.

marcoscaceres:
Closes #91 

Implementation commitment:

- [x] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=275782) - [patch](https://github.com/WebKit/WebKit/pull/30499/)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [x] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/digital-credentials/pull/129.html" title="Last updated on Jul 15, 2024, 1:35 AM UTC (df772ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/digital-credentials/129/d7af7fd...df772ef.html" title="Last updated on Jul 15, 2024, 1:35 AM UTC (df772ef)">Diff</a>

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -34,26 +34,35 @@ At its core, the API is designed for a website ("verifier") to [transparently](h
 
 Here is an example of how the  the API might be used in practice:
 
-```javascript
-    const digitalCredential = await navigator.identity.get({
-      digital: {
-        providers: [{
-          // Protocol extensibility:
-          protocol: "oid4vp",
-          // An example of an OpenID4VP request to wallets.
-          // Based on https://github.com/openid/OpenID4VP/issues/125
-          request: {
-            nonce: "n-0S6_WzA2Mj",
-            presentation_definition: {
-              // Presentation Exchange request, omitted for brevity
-            }
-          }
-        }],
-      },
-    });
-    // To be decrypted on the server...
-    const encryptedData = digitalCredential.data;
+The API needs to be initiated through a user gesture, such as a button click:
+
+```html
+<button onclick="requestLicense()">Request Driver's license<button>
 ```
+
+
+```javascript
+async function requestLicense() {
+  const oid4pv = {
+    // Protocol extensibility:
+    protocol: "oid4vp", // An example of an OpenID4VP request to wallets. // Based on https://github.com/openid/OpenID4VP/issues/125
+    request: {
+      nonce: "n-0S6_WzA2Mj",
+      presentation_definition: {
+        // Presentation Exchange request, omitted for brevity
+      },
+    },
+  };
+  const digitalCredential = await navigator.identity.get({
+    digital: {
+      providers: [oid4pv],
+    },
+  });
+  // To be decrypted on the server...
+  const encryptedData = digitalCredential.data;
+}
+```
+
 You can read a more detailed and technical description of the API in the [specification draft](https://wicg.github.io/digital-identities/).
 
 
--- a/index.html
+++ b/index.html
@@ -281,27 +281,29 @@
       "DigitalCredential">[[\DiscoverFromExternalSource]](origin, options,
       sameOriginWithAncestors)</dfn> internal method MUST:
     </p>
-    <li>Let |global| be [=this=]'s [=relevant global object=].
-    </li>
-    <li>Let |document| be |global|'s [=associated `Document`=].
-    </li>
-    <li>If |document| is not a [=Document/fully active descendant of a
-    top-level traversable with user attention=], [=exception/throw=]
-    {{"NotAllowedError"}} {{DOMException}}.
-    </li>
-    <li>If |window| does not have [=transient activation=], [=exception/throw=]
-    {{"NotAllowedError"}} {{DOMException}}.
-    </li>
-    <li>[=Consume user activation=] of |window|.
-    </li>
-    <li>
-      <aside class="issue">
-        Details of how to actually get the [=digital credential=] are forthcoming.
-      </aside> 
-    </li>
-    <li>
-       Return a {{DigitalCredential}}.
-    </li>
+    <ol class="algorithm">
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -272,6 +272,28 @@ <h3>
...
+    </p>
+    <li>Let |global| be [=this=]'s [=relevant global object=].
+    </li>
+    <li>Let |document| be |global|'s [=associated `Document`=].
+    </li>
+    <li>If |document| is not a [=Document/fully active descendant of a
+    top-level traversable with user attention=], [=exception/throw=]
+    {{"NotAllowedError"}} {{DOMException}}.
+    </li>
+    <li>If |window| does not have [transient activation], [=exception/throw=]
```
marcoscaceres, line 292:
```suggestion
    <li>If |window| does not have [=transient activation=], [=exception/throw=]
```

index.html
```diff
@@ -272,6 +272,28 @@ <h3>
...
+    {{"NotAllowedError"}} {{DOMException}}.
+    </li>
+    <li>[=Consume user activation=] of |window|.
+    </li>
```
samuelgoto, line 296: This algorithm makes general sense up until this point, but for it to be type-correct, it has to return a Credential, right, since [DiscoverFromExternalSource](https://www.w3.org/TR/credential-management-1/#algorithm-discover-creds) returns a `DigitalCredential`?

While we are still figuring out the details of the algorithm from this point, but we could leave for now a TODO here to complete this algorithm or something general like "return a DigitalCredential from a wallet"?

marcoscaceres: Yes... we are not quite there yet lol. I'll add a note because it's confusing right now.

index.html
```diff
@@ -272,6 +272,28 @@ <h3>
...
+    <li>If |document| is not a [=Document/fully active descendant of a
+    top-level traversable with user attention=], [=exception/throw=]
+    {{"NotAllowedError"}} {{DOMException}}.
+    </li>
+    <li>If |window| does not have [=transient activation=], [=exception/throw=]
+    {{"NotAllowedError"}} {{DOMException}}.
+    </li>
+    <li>[=Consume user activation=] of |window|.
+    </li>
```
marcoscaceres, line 296:
```suggestion
    </li>
    <il>
      <aside class="issue">
        Details of how to actually get the [=digital credential=] are forthcoming
      </aside> 
    </li>
    <li>
       Return a {{DigitalCredential}}.
    </li>
```

index.html
marcoscaceres: Also, will switch up the terminology to talk about a "wallet".

index.html
```diff
@@ -272,6 +272,36 @@ <h3>
...
+    <il>
```
marcoscaceres, line 297:
```suggestion
    <li>
```

marcoscaceres:
WebKit tracking bug https://bugs.webkit.org/show_bug.cgi?id=275782

explainer.md
```diff
@@ -34,26 +34,35 @@ At its core, the API is designed for a website ("verifier") to [transparently](h
...
-          }
-        }],
-      },
-    });
-    // To be decrypted on the server...
-    const encryptedData = digitalCredential.data;
+async function requestLicense() {
+  const oid4pv = {
+    // Protocol extensibility:
+    protocol: "oid4vp", // An example of an OpenID4VP request to wallets. // Based on https://github.com/openid/OpenID4VP/issues/125
```
samuelgoto, line 48:
```suggestion
    protocol: "urn:openid.net:oid4vp", // An example of an OpenID4VP request to wallets. // Based on https://github.com/openid/OpenID4VP/issues/125
```

That's the value they are using here: https://github.com/openid/OpenID4VP/pull/155/files

samuelgoto (approved):
LGTM

RByers (approved):

explainer.md
marcoscaceres: I'll see if I can get them to drop that.... using a URN doesn't seem appropriate as otherwise we should parse these as URLs.

samuelgoto: I thought that we were deliberate in making [`protocol`](https://wicg.github.io/digital-credentials/#the-protocol-member) a `DOMString`, so that we wouldn't have to form an opinion on what these were.

Should we care if this is a URN or not? Or are you suggesting that we reject any `DOMString` that isn't a URL?

TallTed:
I suggest correcting the typo in the title, from `visility` to `visibility`

explainer.md
marcoscaceres: def should care if it's an URN... because URNs are URLs, and hence it implies URL parsing. We just don't want to get into a world where people are expecting a URL where we are using a DOMString.

marcoscaceres:
Just sent a PR to WebKit - https://github.com/WebKit/WebKit/pull/30499/  

I might send separate tests to WPT for this... The focus and visibility requirements can be tricky to tests.

marcoscaceres:
Just noting that this is blocked on two things right now...

1. tests
2. The steps we are hooking into run in parallel, which makes what we have specified not quite correct (as written, this assumes we are running in the main thread). I've spun up an issue for the on Cred Man: https://github.com/w3c/webappsec-credential-management/issues/243

marcoscaceres:
We can monkey patch for now though.

bvandersloot-mozilla (approved):

Issue #137, opened 2024-06-26
Add placeholder security and privacy considerations section


RByers:
In the meeting today @npdoty mentioned that reasonable people would be surprised not to see a privacy considerations section in our spec and I agree! We discussed that we should add a placeholder section for security and privacy considerations which links to the places where these are being discussed and incubated. In particular:

- Our [Explainer](https://github.com/WICG/digital-credentials/blob/main/explainer.md) which has the design properties we've agreed on.
- Our [TAG Security and Privacy considerations questionnaire](https://github.com/WICG/digital-credentials/blob/main/horizontal-reviews/security-privacy.md) - still lots of unanswered questions here, but reflects our group's current thinking on many of the details.
- [PING Credential Considerations repo](https://github.com/w3cping/credential-considerations) - much of the discussion that's broader than just the narrow DC API scope.

PR #138, opened 2024-06-26
Tweak explainer design goals

### LINKED ISSUE #86: Nothing is learned without consent

RByers:
Closes #86 

Be more explicit in the goal of sites learning nothing without user permission. Combined with #137, this should be enough for now to call #86 fixed.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -28,7 +28,7 @@ To balance this tension we propose an API with the following key properties:
 - By separating the act of requesting from the specific protocol, we can enable flexibility and adaptability in both the protocol and credential formats. This way, the pace of changes in browsers won't hinder progress or block new developments.
 - Require request transparency, enabling user-agent inspection for risk analysis
 - Assume response opacity (encrypted responses), enabling verifiers and holders to control where potentially sensitive PII is exposed
-- Sites learn nothing without user permission. Prevent websites from silently querying for the availability of digital credentials and communicating with wallet providers prior to the user confirming each action.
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -28,7 +28,7 @@ To balance this tension we propose an API with the following key properties:
 - By separating the act of requesting from the specific protocol, we can enable flexibility and adaptability in both the protocol and credential formats. This way, the pace of changes in browsers won't hinder progress or block new developments.
 - Require request transparency, enabling user-agent inspection for risk analysis
 - Assume response opacity (encrypted responses), enabling verifiers and holders to control where potentially sensitive PII is exposed
-- Prevent website from silently querying for the availability of digital credentials and communicating with wallet providers without explicit user consent 
+- Sites learn nothing without user permission. Prevent websites from silently querying for the availability of digital credentials and communicating with wallet providers prior to the user confirming each action.
```
marcoscaceres, line 31: I think "permission" is just one aspect (but it might not be the right word because it's not really a permission prompt... not in the traditional sense, like with Geolocation)... 

How about: 
```suggestion
- Websites cannot access any information without the user's interaction. This API ensures that sites cannot silently query for digital credentials or communicate with wallet providers without the user's active participation and confirmation of each action.
```

PR #156, opened 2024-08-07
Add request validation

### LINKED ISSUE #130: Define error handling
From discussion in F2F meeting today, we have a few goals:
- No silent information disclosure (eg. can't reveal no matching credential as distinct from user cancelled)
- Developers have a path to diagnose deployment issues

Our conclusions were:
- Most errors will come from the higher level protocol. In that case the wallet should return a protocol-specific object describing the error and the API will technically be a success. It's up to the wallet to ensure whatever information returned fits with their privacy model based on what consent has been collected from the user.
- We will have a few errors at our API level. Once the user has given permission to share a credential, we believe it's OK for us to reveal to the verifier whether there was a matching credential or not (even though they may not have provided consent to the wallet to share any information in particular).
- We want to re-use existing `DOMException` error names and don't believe it should be necessary to define any new ones.
- Wallets will also have the ability to just return a "user cancelled" error which will just return to the credential selector screen.

So the work now is probably twofold:
1. Add spec text (or a note) that describes that API success doesn't necessarily. mean a credential is being returned - there may be a protocol error. 
2. Identify appropriate errors for the API to return.

### LINKED ISSUE #152: Presentation request format validation
When ingesting any presentation format, presumably there will be a set of requirements/constraint that need to be satisfied. For example, expected_origins in OpenID4VP:

> expected_origins: REQUIRED when signed requests defined in [Appendix A.3.2](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#signed_request) are used with the W3C Digital Credentials API [[w3c.digital_credentials_api](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#w3c.digital_credentials_api)]. An array of strings, each string representing an origin of the Verifier that is making the request. The Wallet can detect replay of the request from a malicious Verifier by comparing values in this parameter to the origin asserted by the Web Platform.

 Parsing those URLs may result in parsing error, which should reported a `TypeError`.

I think generally any validation error in the request format will result in a `TypeError`.

marcoscaceres:
Closes #152
Closes #130

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/156.html" title="Last updated on Aug 15, 2025, 6:22 AM UTC (acd1878)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/156/fb88564...acd1878.html" title="Last updated on Aug 15, 2025, 6:22 AM UTC (acd1878)">Diff</a>


index.html
```diff
@@ -378,8 +401,9 @@ <h2 id="protocol-registry">
       Registry of protocols for requesting digital credential
     </h2>
     <p>
-      The following is the registry of [=digital credential/exchange
-      protocols=] that are supported by this specification.
+      The following is the <dfn data-lt="digital credentials registry">digital
+      credentials registryls exchange protocols</dfn> that are supported by
```
OR13, line 405:
```suggestion
      credentials registry exchange protocols</dfn> that are supported by
```

index.html
```diff
@@ -378,8 +401,9 @@ <h2 id="protocol-registry">
       Registry of protocols for requesting digital credential
     </h2>
     <p>
-      The following is the registry of [=digital credential/exchange
-      protocols=] that are supported by this specification.
+      The following is the <dfn data-lt="digital credentials registry">digital
```
OR13, line 404:
```suggestion
      The following are <dfn data-lt="digital credentials registry">digital
```

OR13 (approved):

index.html
```diff
@@ -306,6 +306,29 @@ <h3>
       </li>
       <li>If |providers| is empty, [=exception/throw=] a {{TypeError}}.
       </li>
+      <li>[=List/For each=] |provider| in |providers|:
+        <ol>
+          <li>If |provider|'s {{DigitalCredentialsProvider/protocol}} is does
```
timcappalli, line 311: " ... ~is~ does not match ... "

TallTed:
```suggestion
          <li>If |provider|'s {{DigitalCredentialsProvider/protocol}} does
```

index.html
```diff
@@ -378,8 +401,9 @@ <h2 id="protocol-registry">
       Registry of protocols for requesting digital credential
     </h2>
     <p>
-      The following is the registry of [=digital credential/exchange
-      protocols=] that are supported by this specification.
+      The following are <dfn data-lt="digital credentials registry">digital
+      credentials registry exchange protocols</dfn> supported by
```
TallTed, line 405:
```suggestion
      The following are <dfn data-lt="digital credentials registry">exchange
      protocols from the digital credentials registry</dfn>, supported by
```

TallTed (changes requested):

marcoscaceres:
You're the best @TallTed! Thank you again for those eagle eyes. 🦅

marcoscaceres:
See https://github.com/openid/OpenID4VP/issues/224 for a concrete example of what I'm expecting here...

index.html
```diff
@@ -292,7 +292,38 @@ <h3>
       <li>Let |providers| be |options|'s {{CredentialRequestOptions/digital}}'s
       {{DigitalCredentialRequestOptions/providers}} member.
       </li>
-      <li>If |providers| is empty, [=exception/throw=] a {{TypeError}}.
+      <li>If |providers| [=list/is empty=], [=exception/throw=] a
+      {{TypeError}}.
+      </li>
+      <li>Let |range| be [=list/get the indices=] of |providers|.
```
TallTed, line 298:
```suggestion
      <li>Let |range| be [=list/the indices=] of |providers|.
```

index.html
```diff
@@ -292,7 +292,38 @@ <h3>
...
+      </li>
+      <li>[=Set/For each=] |n| of the |range|:
+        <ol>
+          <li>Let |provider| be [=list/item=] at |providers|'s |n| index.
```
TallTed, line 302:
```suggestion
          <li>Let |provider| be [=list/the item=] at |providers|'s |n| index.
```

index.html
```diff
@@ -366,8 +397,9 @@ <h2 id="protocol-registry">
       Registry of protocols for requesting digital credential
     </h2>
     <p>
-      The following is the registry of [=digital credential/exchange
-      protocols=] that are supported by this specification.
+      The following are <dfn data-lt="digital credentials registry">exchange
+      protocols from the digital credentials registry</dfn>, supported by this
+      specification.
```
TallTed, line 402:
```suggestion
      The following <dfn data-lt="digital credentials registry">exchange
      protocols from the digital credentials registry</dfn> are supported by
      this specification.
```

TallTed (changes requested):

index.html
marcoscaceres: "get the indices" is an algorithm... but I can alias it.

```suggestion
      <li>Let |range| be [=list/get the indices|the indices=] of |providers|.
```

TallTed (approved):

index.html
```diff
@@ -292,7 +292,38 @@ <h3>
...
+          {{DigitalCredentialsProvider/protocol}}.
+          </li>
+          <li>Let |request| be |provider|'s
+          {{DigitalCredentialsProvider/request}} [=converted to an idl
+          value|converted=] to |dictionary|. [=exception/throw|Rethrow=] any
+          exceptions.
+          </li>
+          <li>Validate |request| against any validation rules set forth in the
+          corresponding [=digital credentials registry/specification=]. If any
+          of |request|'s members are invalid, [=exception/throw=] a
```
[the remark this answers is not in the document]
marcoscaceres: Right, but we can validate a lot of things when the request is passed to the browser. I agree that some aspects cannot be validated, and that can be left up to the wallet upon decrypting the request (if the request, or parts of it, is encrypted in the first place). 

At the same time, we need the assurances of what to do when the user agents receive a request in a particular format.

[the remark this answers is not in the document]
samuelgoto: > At the same time, we need the assurances of what to do when the user agents receive a request in a particular format.

My intuition is that there is a subset of the request that the browser cares about and should validate which is the query language, specifically, the [`presentation_definition`](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-presentation_definition-par) parameter in OpenID4VP which takes shape as a [Presentation Exchange](https://identity.foundation/presentation-exchange/).

That seems like something that the browser uses to help the user select which wallet has the credential that's being requested, as well as the privacy and security consequences of sharing that credential.

Beyond that, I think it would be harmful (from a security standpoint) to validate (e.g. drop an unknown parameter) the rest of the parameters. 

Looking at the list of parameters in the [Authorization Request](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-authorization-request) we have the following, in addition to `presentation_definition`:

- `client_id_scheme`
- `client_metadata`
- `nonce`
- `scope`
- `response_mode`

None of these seem like things that the browser needs to form an opinion on, right? 

> if the request, or parts of it, is encrypted in the first place

As soon as we introduce any parameter that is a `DOMString` (for example, `nonce`), verifiers and wallets can communicate freely through that channel. So, there is a escape hatch for any validation that holds the protocol designers and deployment back.

From a security standpoint, I think that's a feature, not a bug: if OpenID4VP (or developers running servers) figure out a safer way to do things, they should be able to develop that without having to ask for browsers to catch up. For example, say there is a production fire happening somewhere, you we wouldn't our browser users to wait until all browsers are patched and redeployed to fix the issue, right?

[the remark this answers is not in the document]
bc-pi: I share much of @samuelgoto's intuition FWIW. But also just want to note that there's work underway in OpenID4VP towards developing a new query language alternative to PE https://github.com/openid/OpenID4VP/pull/220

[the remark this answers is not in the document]
samuelgoto: I wrote down a proposal here: https://github.com/WICG/digital-credentials/issues/161#issue-2481939110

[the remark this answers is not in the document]
marcoscaceres: > As soon as we introduce any parameter that is a DOMString (for example, nonce), verifiers and wallets can communicate freely through that channel.

I don’t think that’s true. We could certainly enforce restrictions or other checks if we see ab/misuse. 

Playing whack-a-mole is not fun, but it’s absolutely something we would do and have done in the past. 

> Beyond that, I think it would be harmful (from a security standpoint) to validate (e.g. drop an unknown parameter) the rest of the parameters.

Potentially, yes. If the request was signed along with any non-standard parameters (which would get dropped by the browser before being passed to the wallet). At the same time, non-standard parameters shouldn’t be part of the request. An RP will likely need some means to check if the browser supports the request structure. 

My intuition, and because of our security architecture, WebKit will drop anything non-standard.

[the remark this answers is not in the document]
marcoscaceres: Spoke @samuelgoto about how this would be implemented in practice, and agreeing with @RByers that this should happen in the wallet or as it's being passed to the wallet, and not for all the things that are passed in.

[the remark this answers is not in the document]
marcoscaceres: I'll try to update this PR soon!

mohamedamir:
@marcoscaceres did you intentionally request review?
Not only this branch has merge conflicts, but it also uses the old "provider" terminology.

marcoscaceres:
Fixed up terminology... good catch. Hopefully better now.

marcoscaceres:
met with @mohamedamir, discussed changing this... 

@marcoscaceres to add a validatedRequest list, gather the validated results, and then continue using the validated requests.

index.html
```diff
@@ -688,20 +688,56 @@ <h3>
...
+              {{DigitalCredentialGetRequest/data}} to that dictionary.
+              </li>
+            </ol>
+          </li>
+          <li>Otherwise, set |processedData| to |request|'s
+          {{DigitalCredentialGetRequest/data}}.
+          </li>
+          <li>Validate |processedData| against any validation rules in the
+          corresponding [=registry/link|specification=]. If validation fails,
+          abort these substeps.
```
bc-pi: My intuition is that this shouldn't be mandated for the browser.

timcappalli: Yes agreed. There was previous consensus on this. Individual user agents may decide to do additional validation checks, but it should not be codified into the spec.

There is also discussion around removing the registry from the spec (#288).

c2bo: I agree and was under the impression that this was agreed upon previously?

mohamedamir: Thank you all for confirming!
It's mostly a left-over from the first version of the PR then.
I will update the PR according and wait for @marcoscaceres to be back from vacation to take a look!

index.html
```diff
@@ -688,19 +688,26 @@ <h3>
...
-          [=exception=].
+          <li><p>Run the following substeps for |request|. If any substep fails (e.g., an [=exception=] is thrown), abort these substeps and continue to the next |request|.</p>
+          </li>
+          <li>[=serialize a JavaScript value to a JSON string|Serialize=] |request|. If serialization fails, abort these substeps.
+          </li>
+          <li>Let |protocol| be |requests|'s {{DigitalCredentialGetRequest/protocol}}.
+          </li>
+          <li>If |protocol| is not allowed by the user agent, abort these substeps.
+          </li>
+          <li>If all previous substeps completed successfully, [=list/append=] |request| to |validRequests|.
```
```suggestion
          <li>[=serialize a JavaScript value to a JSON string|Serialize=] |request|. [=exception/throw|Rethrow=] any exception.</li>.
          </li>
          <li>Let |protocol| be |requests|'s {{DigitalCredentialGetRequest/protocol}}.
          </li>
          <li>If |protocol| is not allowed by the user agent, [=exception/throw=] a {{NotSupportedError}} {{DOMException}}.
          </li>
          <li>[=list/Append=] |request| to |validRequests|.
```

c2bo:
> So, I still don't agree with the assessment here that it's not needed. All the formats require some validation to happen at some point, and that needs to be codified somewhere because it results in some DOM Exception or TypeError.
> 
> As per collecting the valid request, I don't think we actually need to collect them because the result of failing validation is always a some error, which halts all processing:
> 
> 1. the protocol is not supported - error.
> 2. the protocol is known, but not allowed - error.
> 3. the request's payload is invalid (per the validation rules set forth in whatever spec) - error.

Regarding the protocol validation: I am not convinced there needs to be a full protocol-level validation in the Browser. There should probably be security related checks regarding size, valid json etc., but protocol-level validation only needs to happen on the smartphone/wallet side imho. Enforcing such checks in the browser will make it even more difficult to evolve the ecosystem (e.g., by introducing a v1.1 of the interaction protocol) for imho very little practical benefit.

c2bo:
> That's not how it works at least on Apple's platforms. The initial validation is done by OS level framework. Eventually, once deemed safe, the data is passed to the wallet, which may perform its own checks. However, no arbitrary data is passed to the wallet to minimize the risk of the wallet also becoming an attack vector and/or target.

In today's world without DC API, Wallets already parse potentially unsafe QRCodes/Requests and deal with the security concerns. I understand the wish to add additional security checks/layers, especially from the perspective of a user agent or OS implementor where the main focus is to provide a smooth and secure user experience. Part of the question is which layers should be active, what happens in those layers and at what cost and for what benefit.

> "User needs come before the needs of web page authors, which come before the needs of user agent implementors, which come before the needs of specification writers, which come before theoretical purity."

I absolutely agree, but part of the "user need" is also the freedom of choice which wallet or in that case underlying technology to use. If a country/region chooses other protocols or especially other credential/proof formats than the user agent implementors, then the user should still be able to benefit from the security & UX advantages of DC API. The security benefits/guarantees might be slightly worse (more dependant on wallet implementations) than with additional checks along the chain, but it would still be a big improvement, especially for cross device flows.

> The way to evolve the ecosystem is to go through the standardization process, reach consensus, have things implemented, etc... basically what we are doing here.

I fully agree, but who decides which standards should be used/supported? With your proposal, there is no consensus on support. The Browser/OS would become the gatekeeper to applications (wallets) that already fully implement and understand the protocols and formats in that space and said applications would become fully dependant on platform support for any changes (like for example the introduction of zero-knowledge proofs for stronger privacy guarantees).

For the upcoming discussion: Could you describe where exactly what checks are happening or planned to happen especially for a cross device flow?

c2bo:
To clarify since that came up in the discussion last Thursday and I was asked to add it as a comment here in the discussion:

One of the biggest fears or concerns I've heard people bring up in discussions regarding DC API is that its usage might/will be dependant on platform (OS & Browser) providers enabling support for specific protocols / credential formats / document types. Especially in the context of state issued wallets and credentials, that becomes a topic of governmental sovereignty. We should try to make sure we address these concerns and provide some level of guarantee in regards to the openness/neutrality of this interface in the spec. That is imho the biggest risk for the adoption of this API.

marcoscaceres:
@c2bo, let's approach this differently. Let's say I have the following openid4vp request:

```JS
{
  response_type: "this is wrong",
  // other stuff
}
```

When would be the appropriate time to throw an error because `response_type` is wrong?

marcoscaceres:
Closing... maybe no interop issues will emerge with processing formats 🤷

PR #157, opened 2024-08-07
Define registry inclusion rules

### LINKED ISSUE #58: Registry inclusion criteria

### LINKED ISSUE #124: [spec] add statement about responses with PII MUST be encrypted
from #109, specifically https://github.com/WICG/digital-credentials/issues/109#issuecomment-2145712700

Add something along the lines of:

> implementations which pass PII in the response MUST encrypt that information to the verifier in some fashion"

### LINKED ISSUE #191: What are the protocol identifier naming rules?
As we've learned with OID4VP, there are variations of request types (signed vs unsigned) [#185, openid/OpenID4VP/issues/326], and there will be future versions of the protocol. 

As the protocol identifier is essentially "typing" the `data` payload for downstream parties, we need to support a richer protocol identifier beyond a string.

This proposal is to make protocol identifier a URN. For example:

- Signed Requests in OpenID for Verifiable Presentations v1.0: `urn:openid:spec:openid4vp:v1.0:signed`
- Unsigned Requests in OpenID for Verifiable Presentations v1.0: `urn:openid:spec:openid4vp:v1.0:unsigned`
- OpenID for Verifiable Credential Issuance v1.0: `urn:openid:spec:openid4vci:v1.0:<blah>`

(for VCI, should the type be the two different flows (authz code and preauthz code)?)

These identifiers would still be defined in their respective specifications, but high level requirements would be defined here, including requiring this type of identifier for registry inclusion.

marcoscaceres:
Closes #58
Closes #191
Closes #124

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/157.html" title="Last updated on Apr 23, 2025, 10:29 PM UTC (d332dce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/157/5592be0...d332dce.html" title="Last updated on Apr 23, 2025, 10:29 PM UTC (d332dce)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -457,7 +457,7 @@
       <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
       [=digital credential/exchange protocol=] response structure (i.e., the
       [=dictionary=] to which the {{DigitalCredential}}'s
-      {{DigitalCredential/data}} is [=converted to idl values|=] to before it
+      {{DigitalCredential/data}} is [=converted to idl values|converted=] before it
       is made available to the relying party).
       </li>
       <li>MUST define validation rules for members of the request and response
--- a/index.html
+++ b/index.html
@@ -513,14 +513,13 @@
       <dd>
         The protocol identifier MUST be a unique string that is not already in
         use in the registry. Use only lowercase ASCII letters, digits, and
-        hyphens (e.g., "protocol", "the-protocol"). Avoid using version numbers
-        in the protocol identifier. The protocol identifier MUST uniquely
-        define the set of required parameters and/or behavior that a digital
-        credential provider implementation needs to support to successfully
-        handle the request. If the set of required parameters or behaviors is
-        updated in a way which would require a digital credential provider to
-        also require an update to remain functional, a new protocol identifier
-        MUST be assigned and be added to the registry.
+        hyphens (e.g., "protocol", "the-protocol"). The protocol identifier 
+        MUST uniquely define the set of required parameters and/or behavior
+        that a digital credential provider implementation needs to support 
+        to successfully handle the request. If the set of required parameters 
+        or behaviors is updated in a way which would require a digital credential
+        provider to also require an update to remain functional, a new protocol 
+        identifier MUST be assigned and be added to the registry.
       </dd>
       <dt>
         Specify a protocol type
--- a/index.html
+++ b/index.html
@@ -515,6 +515,14 @@
         in the protocol identifier.
       </dd>
       <dt>
+        Specify a protocol type
+      </dt>
+      <dd>
+        The protocol type is either "Presentation" for presentation protocols used with
+        `navigator.credentials.get` or "Issuance" for issuance protocols used
+        with `navigator.credentials.create`.
+      </dd>
+      <dt>
         Link to a Web IDL request dictionary
       </dt>
       <dd>
@@ -562,6 +570,9 @@
           <th>
             <dfn data-dfn-for="digital credentials registry">Protocol
             identifier</dfn>
+          </th>
+          <th>
+            <dfn data-dfn-for="digital credentials registry">Type</dfn>
           </th>
           <th>
             <dfn data-dfn-for="digital credentials registry">Web IDL request
--- a/index.html
+++ b/index.html
@@ -445,22 +445,24 @@
       protocol=]:
     </p>
     <ol>
-      <li>MUST be standardized at a <a href="https://www.w3.org/liaisons/">consortium the W3C liaises with</a>
-      <li>
-        MUST have a stable URL that points to a publicly available specification.
+      <li>MUST be standardized at a <a href=
+      "https://www.w3.org/liaisons/">consortium the W3C liaises with</a>
       </li>
+      <li>MUST have a stable URL that points to a publicly available
+      specification.
       </li>
-      <li>MUST define a representation, as either a [[WebIDL]] [=dictionary=] or a JSON object, of the
-      [=digital credential/exchange protocol=] request structure (i.e., the
-      [=dictionary=] to which the {{DigitalCredentialsProvider}}'s
-      {{DigitalCredentialsProvider/request}} is [=converted to idl
-      values|converted=] to before it is passed onto underlying platform).
+      <li>MUST define a representation, as either a [[WebIDL]] [=dictionary=]
+      or a JSON object, of the [=digital credential/exchange protocol=] request
+      structure (i.e., the [=dictionary=] to which the
+      {{DigitalCredentialsProvider}}'s {{DigitalCredentialsProvider/request}}
+      is [=converted to idl values|converted=] to before it is passed onto
+      underlying platform).
       </li>
-      <li>MUST define a representation, as either a [[WebIDL]] [=dictionary=] or a JSON object, of the
-      [=digital credential/exchange protocol=] response structure (i.e., the
-      [=dictionary=] to which the {{DigitalCredential}}'s
-      {{DigitalCredential/data}} is [=converted to idl values|converted=] before it
-      is made available to the relying party).
+      <li>MUST define a representation, as either a [[WebIDL]] [=dictionary=]
+      or a JSON object, of the [=digital credential/exchange protocol=]
+      response structure (i.e., the [=dictionary=] to which the
+      {{DigitalCredential}}'s {{DigitalCredential/data}} is [=converted to idl
+      values|converted=] before it is made available to the relying party).
       </li>
       <li>MUST define validation rules for members of the request and response
       structures.
@@ -475,8 +477,8 @@
       independent browser engines, to meet the W3C's adequate implementation
       experience requirements.
       </li>
-      <li>MUST be defined in a specification which is available publicly
-      at the URL listed in the registry.
+      <li>MUST be defined in a specification which is available publicly at the
+      URL listed in the registry.
       </li>
       <li>MUST have formally recorded consensus by the Working Group to be
       included in the registry.
@@ -491,12 +493,10 @@
       protocol=]:
     </p>
     <ol>
-      <li>
-        MUST support response encryption.
+      <li>MUST support response encryption.
       </li>
-      <li>
-        MUST encrypt any response containing personally identifiable 
-        information (PII).
+      <li>MUST encrypt any response containing personally identifiable
+      information (PII).
       </li>
     </ol>
     <h3>
@@ -514,21 +514,21 @@
         The protocol identifier MUST be a unique string that is not already in
         use in the registry. Use only lowercase ASCII letters, digits, and
         hyphens (e.g., "protocol", "the-protocol"). Avoid using version numbers
-        in the protocol identifier. The protocol identifier MUST uniquely define 
-        the set of required parameters and/or behavior that a digital credential
-        provider implementation needs to support to successfully handle the request.
-        If the set of required parameters or behaviors is updated in a way which 
-        would require a digital credential provider to also require an update to
-        remain functional, a new protocol identifier MUST be assigned and be added
-        to the registry.
+        in the protocol identifier. The protocol identifier MUST uniquely
+        define the set of required parameters and/or behavior that a digital
+        credential provider implementation needs to support to successfully
+        handle the request. If the set of required parameters or behaviors is
+        updated in a way which would require a digital credential provider to
+        also require an update to remain functional, a new protocol identifier
+        MUST be assigned and be added to the registry.
       </dd>
       <dt>
         Specify a protocol type
       </dt>
       <dd>
-        The protocol type is either "Presentation" for presentation protocols used with
-        `navigator.credentials.get` or "Issuance" for issuance protocols used
-        with `navigator.credentials.create`.
+        The protocol type is either "Presentation" for presentation protocols
+        used with `navigator.credentials.get` or "Issuance" for issuance
+        protocols used with `navigator.credentials.create`.
       </dd>
       <dt>
         Link to a Web IDL request dictionary
--- a/index.html
+++ b/index.html
@@ -469,8 +469,7 @@
       <li>MUST have undergone privacy review by the W3C's <a href="https://www.w3.org/Privacy/IG/">Privacy Interest
       Group</a> and <a href="https://www.w3.org/groups/wg/fedid/">Federated Identity Working Group</a>.
       </li>
-      <li>MUST have undergone security review by the Federated Identity Working
-      Group.
+      <li>MUST have undergone security review by the  <a href="https://www.w3.org/groups/wg/fedid/">Federated Identity Working Group</a>.
       </li>
       <li>MUST have implementation commitment from at least one browser engine,
       one credential provider/wallet, and one issuer or verifier (depending on
--- a/index.html
+++ b/index.html
@@ -476,7 +476,7 @@
       one credential provider/wallet, and one issuer or verifier (depending on
       the protocol type). Each component MUST be from independent organizations.
       </li>
-      <li>MUST have formally recorded consensus by the Working Group to be
+      <li>MUST have formally recorded consensus by the Federated Identity Working Group to be
       included in the registry.
       </li>
     </ol>
--- a/index.html
+++ b/index.html
@@ -488,7 +488,7 @@
       Presentation-specific inclusion criteria
     </h4>
     <p>
-      To be included in the presentation protocol registry (used with
+      To be included as a presentation protocol in the registry (used with
       `navigator.credentials.get`), the [=digital credential/exchange
       protocol=]:
     </p>
--- a/index.html
+++ b/index.html
@@ -426,6 +426,9 @@
       Registry of protocols
     </h2>
     <p>
+      Initiating the registration a protocol is done by <a href="https://github.com/w3c-fedid/digital-credentials/issues">filing an issue</a> in our GitHub repository.
+     </p>
+    <p>
       The following is the registry of [=digital credential/exchange
       protocols=] that are supported by this specification.
     </p>
--- a/index.html
+++ b/index.html
@@ -436,9 +436,101 @@
     <h3>
       Inclusion criteria
     </h3>
+    <aside class="note">
+      The below criteria are a work in progress and are likely to change as
+      this document evolves.
+    </aside>
     <p>
-      To be included in the registry...
+      To be included in the registry, the [=digital credential/exchange
+      protocol=]:
     </p>
+    <ol>
+      <li>MUST be standardized at a recognized standards organization and have
+      a stable URL.
+      </li>
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
+      [=digital credential/exchange protocol=] request structure (i.e., the
+      [=dictionary=] to which the {{DigitalCredentialsProvider}}'s
+      {{DigitalCredentialsProvider/request}} is [=converted to idl
+      values|converted=] to before it is passed onto underlying platform).
+      </li>
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
+      [=digital credential/exchange protocol=] response structure (i.e., the
+      [=dictionary=] to which the {{DigitalCredential}}'s
+      {{DigitalCredential/data}} is [=converted to idl values|=] to before it
+      is made available to the relying party).
+      </li>
+      <li>MUST define validation rules for members of the request and response
+      structures.
+      </li>
+      <li>MUST have undergone privacy review by the W3C's Privacy Interest
+      Group and Federated Identity Working Group.
+      </li>
+      <li>MUST have undergone security review by the Federated Identity Working
+      Group.
+      </li>
+      <li>MUST have implementation commitment from at least two implementers in
+      independent browser engines, to meet the W3C's adequate implementation
+      experience requirements.
+      </li>
+      <li>MUST have formally recorded consensus by the Working Group to be
+      included in the registry.
+      </li>
+    </ol>
+    <p>
+      Once the above criteria are met, the protocol will be included in the
+      registry.
+    </p>
+    <h3>
+      Change process
+    </h3>
+    <p>
+      To add a new [=digital credential/exchange protocol=]to the registry, or
+      to update an existing one:
+    </p>
+    <dl>
+      <dt>
+        Define a protocol identifier
+      </dt>
+      <dd>
+        The protocol identifier MUST be a unique string that is not already in
+        use in the registry. Use only lowercase ASCII letters, digits, and
+        hyphens (e.g., "protocol", "the-protocol"). Avoid using version numbers
+        in the protocol identifier.
+      </dd>
+      <dt>
+        Link to a Web IDL request dictionary
+      </dt>
+      <dd>
+        The <dfn>Web IDL request dictionary</dfn> MUST be a [=dictionary=] that
+        defines the structure of the request that is passed, via
+        {{DigitalCredentialsProvider}}'s
+        {{DigitalCredentialsProvider/request}}, to the holder's a digital
+        wallet.
+      </dd>
+      <dt>
+        Link to a Web IDL response dictionary
+      </dt>
+      <dd>
+        The <dfn>Web IDL response dictionary</dfn> MUST be a [=dictionary=]
+        that defines the structure of {{DigitalCredential}}'s
+        {{DigitalCredential/data}}.
+      </dd>
+      <dt>
+        Describe the protocol
+      </dt>
+      <dd>
+        The description MUST be a brief summary of the protocol's purpose and
+        use case.
+      </dd>
+      <dt>
+        Provide a link to the specification
+      </dt>
+      <dd>
+        The specification MUST be a stable URL that points to the authoritative
+        source for the protocol, including validation rules.
+      </dd>
+    </dl>
     <aside class="issue" data-number="58"></aside>
     <p>
       [=User agents=] MUST support the following [=digital credential/exchange
@@ -452,13 +544,20 @@
       <thead>
         <tr>
           <th>
-            Protocol identifier
+            <dfn data-dfn-for="digital credentials registry">Protocol
+            identifier</dfn>
           </th>
           <th>
-            Description
+            <dfn data-dfn-for="digital credentials registry">Web IDL request
+            dictionary</dfn>
           </th>
           <th>
-            Specification
+            <dfn data-dfn-for="digital credentials registry">Web IDL response
+            dictionary</dfn>
+          </th>
+          <th>
+            <dfn data-dfn-for=
+            "digital credentials registry">Specification</dfn>
           </th>
         </tr>
       </thead>
--- a/index.html
+++ b/index.html
@@ -423,7 +423,7 @@
       </p>
     </section>
     <h2 id="protocol-registry">
-      Registry of protocols for requesting digital credential
+      Registry of protocols
     </h2>
     <p>
       The following is the registry of [=digital credential/exchange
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -388,9 +388,101 @@ <h2 id="protocol-registry">
...
+    </aside>
+    <p>
+      To be included in the registry, the [=digital credential/exchange
+      protocol=]:
+    </p>
+    <ol>
+      <li>MUST be standardized at a recognized standards organization and have
+      a stable URL.
+      </li>
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
```
samuelgoto, line 403: This seems like a good goal to me, but seems hard to accomplish logistically. Where does the WebIDL live? In this spec? In the protocol spec? What happens when they get out of sync?

marcoscaceres: I was imagining the protocol spec. In our spec, we would only link to the dictionary (i.e., or spec would only have `[FooBar](https:/link-to-whatever.com/spec/#FooBar)`).

bvandersloot-mozilla: It seems like it would be easier to include the definition in this spec as part of the inclusion process.

index.html
```diff
@@ -388,9 +388,101 @@ <h2 id="protocol-registry">
...
+      </li>
+      <li>MUST define validation rules for members of the request and response
+      structures.
+      </li>
+      <li>MUST have undergone privacy review by the W3C's Privacy Interest
+      Group and Federated Identity Working Group.
+      </li>
+      <li>MUST have undergone security review by the Federated Identity Working
+      Group.
+      </li>
```
bvandersloot-mozilla, line 423: Do we want to define the criteria by which this group would review those?

index.html
marcoscaceres: It could certainly be done that way. I do share @samuelgoto’s concerns about this spec falling out of sync though. 

I was thinking that having just the name of dictionary in this spec minimizes the chance of things getting out of sync. Listing just the name is also what the Cred Man spec does in its registry:

https://www.w3.org/TR/credential-management-1/#credential-type-registry-appropriate-interface-object

index.html
marcoscaceres: Yes. I think so. I was hoping to let those folks chime in about how we might do it, as some are already members of this group.

index.html
```diff
@@ -388,9 +388,101 @@ <h2 id="protocol-registry">
...
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
+      [=digital credential/exchange protocol=] request structure (i.e., the
+      [=dictionary=] to which the {{DigitalCredentialsProvider}}'s
+      {{DigitalCredentialsProvider/request}} is [=converted to idl
+      values|converted=] to before it is passed onto underlying platform).
+      </li>
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
+      [=digital credential/exchange protocol=] response structure (i.e., the
+      [=dictionary=] to which the {{DigitalCredential}}'s
+      {{DigitalCredential/data}} is [=converted to idl values|=] to before it
```
TallTed, line 412:
```suggestion
      {{DigitalCredential/data}} is [=converted to idl values|converted=] before it
```

TallTed (changes requested):

index.html
```diff
@@ -388,9 +388,101 @@ <h2 id="protocol-registry">
...
+    <p>
+      Once the above criteria are met, the protocol will be included in the
+      registry.
+    </p>
+    <h3>
+      Change process
+    </h3>
     <p>
-      To be included in the registry...
+      To add a new [=digital credential/exchange protocol=]to the registry, or
```
timcappalli, line 440:
```suggestion
      To add a new [=digital credential/exchange protocol=] to the registry, or
```

index.html
```diff
@@ -388,9 +388,101 @@ <h2 id="protocol-registry">
...
+    <aside class="note">
+      The below criteria are a work in progress and are likely to change as
+      this document evolves.
+    </aside>
+    <p>
+      To be included in the registry, the [=digital credential/exchange
+      protocol=]:
+    </p>
+    <ol>
+      <li>MUST be standardized at a recognized standards organization and have
```
timcappalli, line 400:
```suggestion
      <li>MUST be standardized at a recognized standards development organization (SDO) and have
```

timcappalli: Also, does W3C maintain a list that we can reference?

index.html
timcappalli, line 403: Is it worth including a non-normative example?

index.html
marcoscaceres: Possibly https://www.w3.org/liaisons/

marcoscaceres: Actually, yes... that's the right one.

marcoscaceres: Updated the suggestion... dropped the abbreviation, as we don't repeat it.

index.html
marcoscaceres: Certainly could, but if we can get one into OpenID, then that can serve as a good basis. Alternatively, we are going to specify that fake test protocol... so we could point to that too as an example.

OR13 (approved):
This looks good.

If it's possible to use openid as the first example I think a working example will do a better job of explaining than a fake example.

The requirements seem good.

The privacy and security evaluation criteria need to be more explicit though, and there should be a documented appeal process, for rejections.

index.html
bvandersloot-mozilla: Either would work, but we shouldn't require other specs that may already be completed to reopen themselves if they could just drop some webidl that describes themselves here. Particularly where they don't already define their structures with webidl.

marcoscaceres: Yes, that's a good point. Someone can correct me, but I think the only candidates right now are OpendID4VP (and the mDoc part), and maybe W3C VC? I think all of those are in active development with the aim of integrating with this.

> but we shouldn't require other specs that may already be completed to reopen themselves if they could just drop some webidl that describes themselves here. 

Generally speaking, what I'm finding is that you can't take existing specs and expect them to work out of the box with the DC API. They really need to be developed in tandem because the Web and JS have very specific requirements. The IDL requirement, along with clearly specified validation criteria, is actually a forcing function for those specs to integrate properly with this spec.

My feeling right now is that they might need to reopen themselves if they want to integrate, because they would't of being designed to work with a JS API in the first place (and if any such spec exists).

marcoscaceres:
Agree with @OR13. I'm still hopeful that privacy folks will jump into this and help us define the criteria. I've pinged in #58 and let's see about getting this on the agenda for the next call.

npdoty:
This is a novel use of privacy and security reviews for Registry updates, so I don't know that we have settled guidance yet. Please bear with us / let's do our best to figure out the best way to do this.

We did discuss briefly in the issue giving more specifics on privacy-related requirements for the protocol level, but I think we're unlikely to include all of those details within this registry section. Would it still be useful to summarize the high-level needs for the protocol, even if the reviews will inevitably be more detailed?

One consideration is that we expect not only to have reviewed the protocol specs for privacy, _but also_ to have addressed any issues raised in those reviews, if the implications are significant for our privacy expectations for the system as a whole. So I suggest we:
* change the language to require a privacy review and issues addressed, 
* add a citation to the user considerations/threat model document in whatever state we can (so that we can give some set of expectations to those trying to determine whether a protocol is well-suited), and
* indicate that the group should evaluate the results of those reviews when making a consensus call.

(I'm recovering from COVID; please forgive belated replies or awkward wording.)

index.html
samuelgoto: I expanded on my concerns about this specific requirement here on this other PR, but I remain fairly concerned (security-wise) about requiring all browsers to redeploy before protocols can be extended:

https://github.com/WICG/digital-credentials/pull/156#discussion_r1727745625

index.html
```diff
@@ -388,9 +388,101 @@ <h2 id="protocol-registry">
...
+      </dt>
+      <dd>
+        The description MUST be a brief summary of the protocol's purpose and
+        use case.
+      </dd>
+      <dt>
+        Provide a link to the specification
+      </dt>
+      <dd>
+        The specification MUST be a stable URL that points to the authoritative
```
bc-pi, line 482: I'd like to also include some criteria about the authoritative source at that URL being open and freely available.

There's been a lot of talk about developer ergonomics in the context of these protocol identifiers. One of the least ergonomic things a developer can run into is hitting a pay-walled document when just trying to do their job.

timcappalli: @bc-pi, how does this sound? https://github.com/WICG/digital-credentials/pull/157/commits/296438428e10288d029115be9c921e36d4086537

index.html
```diff
@@ -386,11 +386,137 @@ <h2 id="protocol-registry">
...
+    <aside class="note">
+      The below criteria are a work in progress and are likely to change as
+      this document evolves.
+    </aside>
+    <p>
+      To be included in the registry, the [=digital credential/exchange
+      protocol=]:
+    </p>
+    <ol>
+      <li>MUST be standardized at a <a href="https://www.w3.org/liaisons/">recognized standards development organization</a> and have
```
marcoscaceres, line 400:
```suggestion
      <li>MUST be standardized at a <a href="https://www.w3.org/liaisons/">a consortium the W3C liaises with</a> and have
```

index.html
```diff
@@ -386,11 +386,137 @@ <h2 id="protocol-registry">
...
+      a stable URL.
```
marcoscaceres, line 401:
```suggestion
      <li>
        MUST have a stable URL that points to a  freely and publicly available specification.
      </li>
```

index.html
```diff
@@ -386,11 +386,137 @@ <h2 id="protocol-registry">
...
+      </li>
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
```
marcoscaceres, line 403:
```suggestion
      <li>MUST define a [[WebIDL]] [=dictionary=] representation, or similar JSON representation, of the
```

index.html
```diff
@@ -386,11 +386,137 @@ <h2 id="protocol-registry">
...
+      [=digital credential/exchange protocol=] request structure (i.e., the
+      [=dictionary=] to which the {{DigitalCredentialsProvider}}'s
+      {{DigitalCredentialsProvider/request}} is [=converted to idl
+      values|converted=] to before it is passed onto underlying platform).
+      </li>
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
```
marcoscaceres, line 409:
```suggestion
      <li>MUST define a [[WebIDL]] [=dictionary=] representation, or JSON object representation, of the
```

hlozi (changes requested):
Please remove the word “freely” and retain “publicly available”. It’s important for specifications to be publicly available, yet we know that some key formats and protocols for digital credentials are not always freely available. Discussing that in the context of the registry inclusion criteria seems neither suitable nor an effective use of our time.

index.html
marcoscaceres: We can't force "freely" available, unfortunately. ISO probably won't do it. At least it is publicly available.

index.html
```diff
@@ -397,16 +397,18 @@ <h3>
...
+      <li>
+        MUST have a stable URL that points to a publicly available specification.
+      </li>
       </li>
-      <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation, or JSON object representation, of the
       [=digital credential/exchange protocol=] request structure (i.e., the
       [=dictionary=] to which the {{DigitalCredentialsProvider}}'s
       {{DigitalCredentialsProvider/request}} is [=converted to idl
       values|converted=] to before it is passed onto underlying platform).
```
TallTed, line 409:
```suggestion
      <li>
        MUST define a representation, as either a [[WebIDL]] [=dictionary=] or a JSON object, of the
        [=digital credential/exchange protocol=] request structure (i.e., the
        [=dictionary=] to which the {{DigitalCredentialsProvider}}'s
        {{DigitalCredentialsProvider/request}} is [=converted to idl
        values|converted=] before it is passed on to underlying platform).
```

index.html
```diff
@@ -397,16 +397,18 @@ <h3>
...
       </li>
-      <li>MUST define a [[WebIDL]] [=dictionary=] representation of the
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation, or JSON object representation, of the
       [=digital credential/exchange protocol=] response structure (i.e., the
       [=dictionary=] to which the {{DigitalCredential}}'s
       {{DigitalCredential/data}} is [=converted to idl values|converted=] before it
```
TallTed, line 414:
```suggestion
      <li>
        MUST define a representation, as either a [[WebIDL]] [=dictionary=] or a JSON object, of the
        [=digital credential/exchange protocol=] response structure (i.e., the
        [=dictionary=] to which the {{DigitalCredential}}'s
        {{DigitalCredential/data}} is [=converted to idl values|converted=] before it
```

TallTed (changes requested):

index.html
bc-pi: > We can't force "freely" available, unfortunately.

Why not?

> ISO probably won't do it. 

ISO or any other SDO could absolutely chose to make a specification available freely and publicly at a URL.

bc-pi: > @bc-pi, how does this sound? [2964384](https://github.com/WICG/digital-credentials/commit/296438428e10288d029115be9c921e36d4086537)

Apologies for the slow response, I was on vacation with my family. But thank you, that text sounds good.

bc-pi:
> Please remove the word “freely” and retain “publicly available”. It’s important for specifications to be publicly available, yet we know that some key formats and protocols for digital credentials are not always freely available. Discussing that in the context of the registry inclusion criteria seems neither suitable nor an effective use of our time.


Discussing criteria for inclusion in the registry, as part of a PR literally titled "Define registry inclusion rules", seems not only suitable but essential for this group, which is tasked with defining both the API and its associated registry.

To my admittedly naive eye, pushback against requiring specifications to be both freely and publicly available feels antithetical to the W3C’s ethos of openness and accessibility. The W3C's core principles emphasize that Web standards should be available to everyone, to support interoperability, global participation, and the continued growth of the Web.

What would be an ineffective use of our time, in my view, is a prolonged debate over whether free and open access to specifications should be a baseline expectation.

index.html
jogu: > We can't force "freely" available, unfortunately. ISO probably won't do it. At least it is publicly available.

Isn't this a chicken and egg problem - if the requirement for inclusion in the registry is freely available, then ISO probably would do it?

jogu:
> To my admittedly naive eye, pushback against requiring specifications to be both freely and publicly available feels antithetical to the W3C’s ethos of openness and accessibility. The W3C's core principles emphasize that Web standards should be available to everyone, to support interoperability, global participation, and the continued growth of the Web.

I think @bc-pi does raise an interesting point here. Is there precedent for W3C working groups effectively endorsing non-freely available specifications?

Sakurann:
+1 to @bc-pi 's points, which are, in my view, in line with [Value of creating standards at W3C](https://www.w3.org/standards/#value).

Sakurann (approved):

index.html
```diff
@@ -386,11 +386,139 @@ <h2 id="protocol-registry">
...
+      </li>
+      <li>MUST have undergone privacy review by the W3C's Privacy Interest
+      Group and Federated Identity Working Group.
+      </li>
+      <li>MUST have undergone security review by the Federated Identity Working
+      Group.
+      </li>
+      <li>MUST have implementation commitment from at least two implementers in
+      independent browser engines, to meet the W3C's adequate implementation
+      experience requirements.
```
RByers, line 428: I'm not sure this makes sense. What examples of other W3C registries do we have and how they translate W3C interoperability requirements down to the things they point at?

index.html
```diff
@@ -386,11 +386,139 @@ <h2 id="protocol-registry">
...
+      </li>
+      <li>MUST be defined in a specification which is available freely and publicly
```
marcoscaceres, line 430:
```suggestion
      <li>MUST be defined in a specification which is available publicly
```

index.html
```diff
@@ -386,11 +386,139 @@ <h2 id="protocol-registry">
...
+      [=dictionary=] to which the {{DigitalCredentialsProvider}}'s
+      {{DigitalCredentialsProvider/request}} is [=converted to idl
+      values|converted=] to before it is passed onto underlying platform).
+      </li>
+      <li>MUST define a [[WebIDL]] [=dictionary=] representation, or JSON object representation, of the
+      [=digital credential/exchange protocol=] response structure (i.e., the
+      [=dictionary=] to which the {{DigitalCredential}}'s
+      {{DigitalCredential/data}} is [=converted to idl values|converted=] before it
+      is made available to the relying party).
+      </li>
```
marcoscaceres, line 416:
```suggestion
      <li>MUST define a JSON object representation, or [[WebIDL]] [=dictionary=] representation, of the
      [=digital credential/exchange protocol=] response structure (i.e., the
      [=dictionary=] to which the {{DigitalCredential}}'s
      {{DigitalCredential/data}} is [=converted to idl values|converted=] before it
      is made available to the relying party).
      </li>
```

index.html
RByers: In particular, the primary implementation of these protocols is not browsers, it's wallets and verifiers. Therefore according to [W3C process](https://www.w3.org/policies/process/#implementation-experience) I believe it's sufficient to just have multiple wallets supporting a protocol.

index.html
```diff
@@ -427,7 +427,7 @@ <h3>
       independent browser engines, to meet the W3C's adequate implementation
       experience requirements.
       </li>
-      <li>MUST be defined in a specification which is available freely and publicly
+      <li>MUST be defined in a specification which is available publicly
```
timcappalli, line 430: There was not consensus on the call about this.

bc-pi: I think it's fair to say there's not full consensus either way. But there has been support expressed by a number of folks in various places in this PR for inclusion of "freely".

index.html
TomCJones: what's the stable URL of the mDoc?

index.html
TomCJones: if w3c privacy review privacy why doesn't w3c security review security?

index.html
```diff
@@ -386,11 +386,135 @@ <h2 id="protocol-registry">
...
+        MUST have a stable URL that points to a publicly available specification.
+      </li>
+      </li>
+      <li>MUST define a JSON object representation, or [[WebIDL]] [=dictionary=] representation, of the
+      [=digital credential/exchange protocol=] request structure (i.e., the
+      [=dictionary=] to which the {{DigitalCredentialsProvider}}'s
+      {{DigitalCredentialsProvider/request}} is [=converted to idl
+      values|converted=] to before it is passed onto underlying platform).
+      </li>
+      <li>MUST define a JSON object representation, or [[WebIDL]] [=dictionary=] representation, of the
```
TallTed, line 411:
```suggestion
      <li>MUST define a representation, as either a [[WebIDL]] [=dictionary=] or a JSON object, of the
      [=digital credential/exchange protocol=] request structure (i.e., the
      [=dictionary=] to which the {{DigitalCredentialsProvider}}'s
      {{DigitalCredentialsProvider/request}} is [=converted to idl
      values|converted=] to before it is passed onto underlying platform).
      </li>
      <li>MUST define a representation, as either a [[WebIDL]] [=dictionary=] or a JSON object, of the
```

MasterKale:
I also think we should keep the requirement that an entry in the registry be freely available.

Making DC API a "pay to play" API if a more popular protocol in its registry requires implementers to pay for access to its spec is not considerate to the wider internet community. If a non-free protocol becomes popular enough then sure, people will buy access to it anyway. But as the product of W3C processes then the DC API's registry should reflect the values of the W3C and seek to make its capabilities available to as many people on the web as are interested in using it. I think this necessarily means we MUST (_normative_) require free access as part of making it into the registry.

index.html
```diff
@@ -434,11 +434,117 @@ <h2 id="protocol-registry">
...
     <p>
-      To be included in the registry...
+      To be included in the registry, the [=digital credential/exchange
+      protocol=]:
     </p>
+    <ol>
+      <li>MUST be standardized at a <a href=
+      "https://www.w3.org/liaisons/">consortium the W3C liaises with</a>
+      </li>
+      <li>MUST have a stable URL that points to a publicly available
```
bc-pi, line 451:
```suggestion
      <li>MUST have a stable URL that points to a freely and publicly available
```

index.html
bc-pi: By my count, there have been numerous individuals from a diverse set of organizations expressing support for the requirement that an entry in the registry be freely available and two individuals from a single organization saying that listing pay-walled specifications is okay.

Despite that count, the e11c2f70fb39f07092146b2554883dda79397df6 change removes the word "freely".

index.html
bc-pi: related to https://github.com/WICG/digital-credentials/pull/157#discussion_r2036167357

bc-pi (changes requested):
https://github.com/WICG/digital-credentials/pull/157#discussion_r2036167357

index.html
```diff
@@ -434,11 +434,117 @@ <h2 id="protocol-registry">
...
+    </p>
+    <dl>
+      <dt>
+        Define a protocol identifier
+      </dt>
+      <dd>
+        The protocol identifier MUST be a unique string that is not already in
+        use in the registry. Use only lowercase ASCII letters, digits, and
+        hyphens (e.g., "protocol", "the-protocol"). Avoid using version numbers
+        in the protocol identifier. The protocol identifier MUST uniquely
```
timcappalli, line 517: 2025-04-11 WG call: remove ` Avoid using version number in the protocol identifier.`

index.html
marcoscaceres: Depends how much they want to be part of the new hotness that is the DC API.

index.html
TallTed: Note that this avoidance results in a requirement that the protocol include version negotiation, even if that is as simple as the two sides of a protocol exchange announcing their version, each to the other, with the potential that either or both sides will respond, `I cannot interact with that version of this protocol. <Something> must be updated.`

index.html
timcappalli: Issue to track here: https://github.com/WICG/digital-credentials/issues/215

index.html
timcappalli: Discussed during the hybrid meeting on 2025-04-11 and there was loose consensus to remove this and re-evaluate in the future if needed.

wseltzer:
Discussed at 21 April meeting [notes](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-04-21-notes.md)

bc-pi (approved):

RByers (approved):
This looks great to me now, thanks Marcos and Tim!

samuelgoto (approved):
Inclusion rules LGTM++

c2bo (approved):

index.html
```diff
@@ -434,11 +434,115 @@ <h2 id="protocol-registry">
...
+      </li>
+      <li>MUST have undergone security review by the Federated Identity Working
+      Group.
+      </li>
+      <li>MUST have implementation commitment from at least one browser engine,
+      one credential provider/wallet, and one issuer or verifier (depending on
+      the protocol type). Each component MUST be from independent organizations.
+      </li>
+      <li>MUST be defined in a specification which is available publicly at the
+      URL listed in the registry.
```
c2bo, line 480: Seems somewhat redundant to
> MUST have a stable URL that points to a freely and publicly available
    specification.

index.html
```diff
@@ -434,11 +434,112 @@ <h2 id="protocol-registry">
...
+      or a JSON object, of the [=digital credential/exchange protocol=]
+      response structure (i.e., the [=dictionary=] which defines the
+      semantics and validation of the {{DigitalCredential}}'s 
+      {{DigitalCredential/data}} member.
+      </li>
+      <li>MUST define validation rules for members of the request and response
+      structures.
+      </li>
+      <li>MUST have undergone privacy review by the W3C's Privacy Interest
+      Group and Federated Identity Working Group.
```
marcoscaceres, line 470: We should add a note here about who handles this. We should also link to the Privacy group's homepage.

index.html
```diff
@@ -434,11 +434,112 @@ <h2 id="protocol-registry">
...
+      </li>
+      <li>MUST have undergone security review by the Federated Identity Working
+      Group.
+      </li>
+      <li>MUST have implementation commitment from at least one browser engine,
+      one credential provider/wallet, and one issuer or verifier (depending on
+      the protocol type). Each component MUST be from independent organizations.
+      </li>
+      <li>MUST have formally recorded consensus by the Working Group to be
```
marcoscaceres, line 479:
```suggestion
      <li>MUST have formally recorded consensus by the Federated Identity Working Group to be
```

index.html
marcoscaceres, line 470:
```suggestion
      <li>MUST have undergone privacy review by the W3C's <a href="https://www.w3.org/Privacy/IG/">Privacy Interest
      Group</a> and <a href="https://www.w3.org/groups/wg/fedid/">Federated Identity Working Group</a>.
```

index.html
```diff
@@ -434,11 +434,112 @@ <h2 id="protocol-registry">
...
+      {{DigitalCredential/data}} member.
+      </li>
+      <li>MUST define validation rules for members of the request and response
+      structures.
+      </li>
+      <li>MUST have undergone privacy review by the W3C's <a href="https://www.w3.org/Privacy/IG/">Privacy Interest
+      Group</a> and <a href="https://www.w3.org/groups/wg/fedid/">Federated Identity Working Group</a>.
+      </li>
+      <li>MUST have undergone security review by the Federated Identity Working
+      Group.
```
marcoscaceres, line 473:
```suggestion
      <li>MUST have undergone security review by the  <a href="https://www.w3.org/groups/wg/fedid/">Federated Identity Working Group</a>.
```

index.html
```diff
@@ -434,11 +434,111 @@ <h2 id="protocol-registry">
...
+      or a JSON object, of the [=digital credential/exchange protocol=]
+      response structure (i.e., the [=dictionary=] which defines the
+      semantics and validation of the {{DigitalCredential}}'s 
+      {{DigitalCredential/data}} member.
+      </li>
+      <li>MUST define validation rules for members of the request and response
+      structures.
+      </li>
+      <li>MUST have undergone privacy review by the W3C's <a href="https://www.w3.org/Privacy/IG/">Privacy Interest
+      Group</a> and <a href="https://www.w3.org/groups/wg/fedid/">Federated Identity Working Group</a>.
```
marcoscaceres, line 470:
```suggestion
      Group</a> and <a href="https://www.w3.org/groups/wg/fedid/">Federated Identity Working Group</a>.
      <aside class="note" title="Organizing reviews">
        Once an expression of registration is received via GitHub, the registry maintainers will organize the privacy review with the <a href="https://www.w3.org/Privacy/IG/">Privacy Interest Group</a> . Please see the [[[[security-privacy-questionnaire]]] for the kind of questions that will be asked of the protocol you are registering. 
      </aside>
```

index.html
```diff
@@ -423,7 +423,7 @@ <h2>
       </p>
     </section>
     <h2 id="protocol-registry">
-      Registry of protocols for requesting digital credential
+      Registry of protocols
     </h2>
```
marcoscaceres, line 427:
```suggestion
    </h2>
    <p>
      Initiating the registration a protocol is done by <a href="https://github.com/w3c-fedid/digital-credentials/issues">filing an issue</a> in our GitHub repository.
     </p>
```

marcoscaceres:
I made some changes. @timcappalli, please take a look. If all looks good, please merge.

samuelgoto (approved):
Changes LGTM++

timcappalli (approved):

PR #165, opened 2024-09-04
change providers to requests

### LINKED ISSUE #155: Rename "providers" to "presentationRequests" or just  "requests"?
As came up in #71, "providers" might not be the most clear member name to use:

@timcappalli [proposed](https://github.com/WICG/digital-credentials/pull/71#issuecomment-2260675780):

```WebIDL
dictionary DigitalCredentialRequestOptions {
  sequence<DigitalCredentialsPresentationRequest> presentationRequests;
};
```

I'd be ok with just "requests" tho, but `presentationRequests` is also pretty clear.

marcoscaceres:
Closes #155 

- [x] [Modified Web platform tests](https://github.com/WICG/digital-credentials/pull/165)

Implementation commitment:

- [x] WebKit - [issue](https://bugs.webkit.org/show_bug.cgi?id=279933), [issue](https://bugs.webkit.org/show_bug.cgi?id=279931) 
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/digital-credentials/pull/165.html" title="Last updated on Sep 18, 2024, 11:23 PM UTC (48f6041)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/digital-credentials/165/5c07f1f...48f6041.html" title="Last updated on Sep 18, 2024, 11:23 PM UTC (48f6041)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -213,8 +213,8 @@
       The `DigitalCredentialsRequest` dictionary
     </h2>
     <p>
-      The {{DigitalCredentialsRequest}} dictionary represents a [=presentation
-      request]. It is used to specify an [=digital credential/exchange
+      The {{DigitalCredentialsRequest}} dictionary represents a [=digital credential/presentation
+      request=]. It is used to specify an [=digital credential/exchange
       protocol=] and a [=digital credential/request data=], which the user
       agent MAY match against software used by a holder, such as a digital
       wallet.
--- a/index.html
+++ b/index.html
@@ -101,7 +101,7 @@
         </p>
       </dd>
       <dt>
-        <dfn data-dfn-for="digital credential">query</dfn>
+        <dfn data-dfn-for="digital credential">request data</dfn>
       </dt>
       <dd>
         A format that [=verifier=] software or a [=user agent=] uses, via an
@@ -114,7 +114,7 @@
       <dd>
         A format that a [=holder|holder's=] software, such as a digital wallet,
         uses, via an [=digital credential/exchange protocol=], to respond to a
-        [=digital credential/query=] by a [=verifier=].
+        [=digital credential/request data=] by a [=verifier=].
       </dd>
       <dt>
         <dfn data-dfn-for="digital credential">Exchange protocol</dfn>
@@ -198,8 +198,8 @@
     <p>
       The <dfn data-dfn-for="DigitalCredentialRequestOptions">requests</dfn>
       specify an [=digital credential/exchange protocol=] and [=digital
-      credential/query=], which the user agent MAY match against a holder's
-      software, such as a digital wallet.
+      credential/request data=], which the user agent MAY match against a
+      holder's software, such as a digital wallet.
     </p>
     <h2>
       The `DigitalCredentialsRequest` dictionary
@@ -207,8 +207,8 @@
     <p>
       The {{DigitalCredentialsRequest}} dictionary is used to specify an
       [=digital credential/exchange protocol=] and a [=digital
-      credential/query=], which the user agent MAY match against software used
-      by a holder, such as a digital wallet.
+      credential/request data=], which the user agent MAY match against
+      software used by a holder, such as a digital wallet.
     </p>
     <pre class="idl">
     dictionary DigitalCredentialsRequest {
@@ -234,8 +234,8 @@
     </h3>
     <p>
       The <dfn data-dfn-for="DigitalCredentialsRequest">data</dfn> member is
-      the [=digital credential/query=] to be handled by the holder's software,
-      such as a digital wallet.
+      the [=digital credential/request data=] to be handled by the holder's
+      software, such as a digital wallet.
     </p>
     <h2>
       The `DigitalCredential` interface
--- a/index.html
+++ b/index.html
@@ -101,7 +101,7 @@
         </p>
       </dd>
       <dt>
-        <dfn>Presentation request</dfn>
+        <dfn data-dfn-for="digital credential">Presentation request</dfn>
       </dt>
       <dd>
         A presentation request is a request for a digital credential composed
--- a/index.html
+++ b/index.html
@@ -189,7 +189,7 @@
     </h2>
     <pre class="idl">
     dictionary DigitalCredentialRequestOptions {
-      sequence&lt;DigitalCredentialsProvider&gt; requests;
+      sequence&lt;DigitalCredentialsRequest&gt; requests;
     };
     </pre>
     <h3>
@@ -213,7 +213,7 @@
     <pre class="idl">
     dictionary DigitalCredentialsRequest {
       required DOMString protocol;
-      required object request;
+      required object data;
     };
     </pre>
     <h3>
@@ -230,12 +230,12 @@
       custom one.
     </p>
     <h3>
-      The `request` member
+      The `data` member
     </h3>
     <p>
-      The <dfn data-dfn-for="DigitalCredentialsRequest">request</dfn> member
-      is the request to be handled by the holder's software, such as a digital
-      wallet.
+      The <dfn data-dfn-for="DigitalCredentialsRequest">data</dfn> member is
+      the [=digital credential/query=] to be handled by the holder's software,
+      such as a digital wallet.
     </p>
     <h2>
       The `DigitalCredential` interface
--- a/index.html
+++ b/index.html
@@ -189,29 +189,29 @@
     </h2>
     <pre class="idl">
     dictionary DigitalCredentialRequestOptions {
-      sequence&lt;DigitalCredentialsProvider&gt; providers;
+      sequence&lt;DigitalCredentialsProvider&gt; requests;
     };
     </pre>
     <h3>
-      The `providers` member
+      The `requests` member
     </h3>
     <p>
-      The <dfn data-dfn-for="DigitalCredentialRequestOptions">providers</dfn>
+      The <dfn data-dfn-for="DigitalCredentialRequestOptions">requests</dfn>
       specify an [=digital credential/exchange protocol=] and [=digital
       credential/query=], which the user agent MAY match against a holder's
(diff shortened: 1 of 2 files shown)


samuelgoto:
Something that occurred to me that feels a bit off is that we'd have a `requests` and `request` in the API. Here:

```javascript
await navigator.credentials.get({
  digital: {
     requests: [{      <-----------
       protocol: "openid4vp",
       request: {.      <----------- is this weird?
         nonce: "foo",
         client_id_metadata: "bar",
         // ... more stuff
       },
     }]
  }
});
```

marcoscaceres:
Yeah, I spotted that too... I can live with it though. 

How about `.query:`?

marcoscaceres:
Some alternatives:
* `presentationRequests` and `presentationDetails`
* `presentmentRequests` and `requestDetails`

bc-pi:
To add to the a bit odd but not terrible observation, due to the layering and reuse of prior RFCs, the top level thing[1] of a `request` in a signed openid4vp request is also a `request`. See not permanent content at https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#appendix-A.3.2-3 but that would mean something like the below with `requests: ... request: ... request: "..."` which is pretty special. 

```
await navigator.identity.get({
  digital: {
    requests: [{
      protocol: "openid4vp",
      request: {
        request: "eyJhbGciOiJF..."
     }
    }]
  }
});
```

[1] I can't even keep track of what things at what layer are called anymore

bc-pi:
Can someone remind me what the rational for the `requests: [...]` array under `digital` rather than a single request?

bc-pi:
> Can someone remind me what the rational for the `requests: [...]` array under `digital` rather than a single request?

The layers and potential options for various cardinalities is making my little mind hurt. Kinda related is https://github.com/openid/OpenID4VP/issues/248 not to mention that whatever query language(s) in openid4vp can themselves express asking for multiple things and/or combinatorial options.

samuelgoto:
> Can someone remind me what the rational for the requests: [...] array under digital rather than a single request?

Yeah, @bc-pi , that's one of the things that I think we may be worth revisiting: I'm not quite sure anymore we need the level of indirection of `providers`.

To answer your question, originally, `providers` was introduced because we thought that there would be more than 1 `protocol` used (e.g. the **same** request using OpenID4VP, VC API, something coming from ISO, etc) and that wallets would accept "one or the other". A similar level of indirection that we introduced for formats (e.g. mDocs and SD-JWTs).  

At this point, I think it is not clear to me that we would run into this situation: when a verifier makes a request, it doesn't have to connect with a variety of wallets that support different protocols.

marcoscaceres:
I wouldn't mind dropping `.requests` entirely to be honest. CoreIDV on Apple's platform doesn't support requesting multiple, so we would just pick the first known type. 

We should then add a. static `.supportsRequestType()` to the `DigitalCredential` interface, or some such. Otherwise developers won't know if it will immediately fail.

msporny:
@samuelgoto wrote:
> To answer your question, originally, providers was introduced because we thought that there would be more than 1 protocol used (e.g. the same request using OpenID4VP, VC API, something coming from ISO, etc) and that wallets would accept "one or the other". A similar level of indirection that we introduced for formats (e.g. mDocs and SD-JWTs).

Hmm, no, we definitely still have that requirement. For example, in CA DMV, we might have a future where mDoc is requested over a different protocol (or protocol version) than other VCs. We cannot presume that there will only be one protocol.

marcoscaceres:
Right, @msporny, but the future format would need to be known to the browser, so "mdoc" would need to be a registered protocol (i.e., you can't sent up anything as that would expose what is installed on the device, so for user privacy you can't send "wallet-custom-format"). The developer can check and do the filtering on their most preferred **standardized** type there.

marcoscaceres:
Moved the proposal here: https://github.com/WICG/digital-credentials/issues/168

msporny:
> The developer can check and do the filtering on their most preferred **standardized** type there.

Yes, agreed. I wasn't making an argument against having a set of standardized protocols or standardized formats.

It sounded like @samuelgoto was saying: "We don't need to contemplate a set of standardized protocols that a verifier might use to request a credential, because there will only ever be one that a verifier will use to request a credential." ... and that's not true today and is unlikely to be true in the future (as much as we'd all like that to be true).

marcoscaceres:
Ok, talked with @samuelgoto and he also convinced me that we should keep the array. 

for the member names, how about we do `{ requests: [{protocol: "openid4vp", data: {} }] }`.

bc-pi:
> > Can someone remind me what the rational for the requests: [...] array under digital rather than a single request?
> 
> Yeah, @bc-pi , that's one of the things that I think we may be worth revisiting: I'm not quite sure anymore we need the level of indirection of `providers`.
> 
> To answer your question, originally, `providers` was introduced because we thought that there would be more than 1 `protocol` used (e.g. the **same** request using OpenID4VP, VC API, something coming from ISO, etc) and that wallets would accept "one or the other". A similar level of indirection that we introduced for formats (e.g. mDocs and SD-JWTs).
> 
> At this point, I think it is not clear to me that we would run into this situation: when a verifier makes a request, it doesn't have to connect with a variety of wallets that support different protocols.

Thanks for backstory @samuelgoto. My general intuition is that there are too many levels of indirection in all the various layers of all this. But that's easy to say and not so easy to clearly say what should be removed.

bc-pi:
> Ok, talked with @samuelgoto and he also convinced me that we should keep the array.
> 
> for the member names, how about we do `{ requests: [ { protocol: "openid4vp", data: {} }] }`.
> 
> So `requests` and `data`.

I like `requests` and `data`, thanks @marcoscaceres!

samuelgoto (approved):
LGTM

This is a backwards incompatible change, so we are going to need to update the OpenID4VP spec as well as verifiers. It is still possible, but we shouldn't take this lightly.

Can we run this by the community group to make sure that we are on all board with this trade-off?

bc-pi (approved):

index.html
```diff
@@ -189,53 +189,53 @@ <h2>
...
     <h3>
-      The `request` member
+      The `data` member
     </h3>
     <p>
-      The <dfn data-dfn-for="DigitalCredentialsProvider">request</dfn> member
-      is the request to be handled by the holder's software, such as a digital
-      wallet.
+      The <dfn data-dfn-for="DigitalCredentialsRequest">data</dfn> member is
+      the [=digital credential/query=] to be handled by the holder's software,
```
samuelgoto, line 237:
```suggestion
      the [=digital credential/data=] to be handled by the holder's software,
```

index.html
```diff
@@ -114,7 +114,7 @@ <h2>
       <dd>
         A format that a [=holder|holder's=] software, such as a digital wallet,
         uses, via an [=digital credential/exchange protocol=], to respond to a
-        [=digital credential/query=] by a [=verifier=].
+        [=digital credential/request data=] by a [=verifier=].
```
samuelgoto, line 117:
```suggestion
        [=digital credential/request=] by a [=verifier=].
```

index.html
```diff
@@ -101,7 +101,15 @@ <h2>
         </p>
       </dd>
       <dt>
-        <dfn data-dfn-for="digital credential">query</dfn>
+        <dfn>Presentation request</dfn>
+      </dt>
+      <dd>
+        A presentation request is a request for a digital credential composed
```
samuelgoto, line 107:
```suggestion
        A presentation request is a request for a [=digital credential=] composed
```

index.html
```diff
@@ -101,7 +101,15 @@ <h2>
         </p>
       </dd>
       <dt>
-        <dfn data-dfn-for="digital credential">query</dfn>
+        <dfn>Presentation request</dfn>
```
marcoscaceres, line 104:
```suggestion
        <dfn data-dfn-for="digital credential">Presentation request</dfn>
```

index.html
```diff
@@ -189,53 +197,54 @@ <h2>
...
-      The `DigitalCredentialsProvider` dictionary
+      The `DigitalCredentialsRequest` dictionary
     </h2>
     <p>
-      The {{DigitalCredentialsProvider}} dictionary is used to specify an
-      [=digital credential/exchange protocol=] and a [=digital
-      credential/query=], which the user agent MAY match against software used
-      by a holder, such as a digital wallet.
+      The {{DigitalCredentialsRequest}} dictionary represents a [=presentation
+      request]. It is used to specify an [=digital credential/exchange
```
marcoscaceres, line 217:
```suggestion
      The {{DigitalCredentialsRequest}} dictionary represents a [=digital credential/presentation
      request=]. It is used to specify an [=digital credential/exchange
```

samuelgoto:
```suggestion
      request=]. It is used to specify an [=digital credential/exchange
```

samuelgoto (approved):
LGTM

This is a backwards incompatible change, but a good one in case we still have time. Lets run this by the CG before merging.

samuelgoto:
Just for transparency, the other alternatives that we considered for `data` was `payload`, `params` and `body`, but `data` seemed best.

timcappalli:
2024-09-18 call: good to merge once we have links

marcoscaceres:
Updated the tests https://github.com/WICG/digital-credentials/pull/165

marcoscaceres:
Filed issues on WebKit side...

marcoscaceres:
Going ahead and merging, as we have tests and implementation commitments and it's mostly a cosmetic change.

samuelgoto:
> Filed issues on WebKit side...

Chromium bug filed:

https://g-issues.chromium.org/issues/370982995

PR #173, opened 2024-09-20
Update explainer to include feature detection and new syntax

### LINKED ISSUE #170: Update example to show feature detection
Now that we're back to using `navigator.credentials`, our simple feature detection mechanism of `if ('identity' in navigator)` won't work. Presumably we'd recommend `if ('DigitalCredential' in window)` instead, right? Perhaps we should add this to the example in the explainer to make it clear? 

I know of one developer who wasn't sure how best to feature-detect our origin trial and so instead looked for a Chrome Android UA string with Chrome version >= 128. Obviously that's not what we want people doing since the API will become available on other browsers too. Our examples should document our best-practice recommendations.

timcappalli:
Closes #170

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -37,33 +37,58 @@ Here is an example of how the  the API might be used in practice:
 The API needs to be initiated through a user gesture, such as a button click:
 
 ```html
-<button onclick="requestLicense()">Request Driver's license<button>
+<button onclick="requestCredential()">Request Driver's license<button>
 ```
 
 
 ```javascript
-async function requestLicense() {
-  const oid4pv = {
-    // Protocol extensibility:
-    protocol: "oid4vp", // An example of an OpenID4VP request to wallets. // Based on https://github.com/openid/OpenID4VP/issues/125
-    data: {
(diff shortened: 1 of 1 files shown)


OR13 (approved):
I wonder if adding fetch would improve this example.

timcappalli:
> I wonder if adding fetch would improve this example.

the full example that includes fetch is linked below the code snippet.

explainer.md
```diff
@@ -37,33 +37,58 @@ Here is an example of how the  the API might be used in practice:
...
-      requests: [oid4pv],
-    },
-  });
-  // To be decrypted on the server...
-  const encryptedData = digitalCredential.data;
-}
+async function requestCredential() {
+
+  // Check for Digital Credentials API support
+  if (typeof window.DigitalCredential !== 'undefined') {
```
samuelgoto, line 48: Maybe return early here? e.g.

```suggestion
  if (typeof window.DigitalCredential !== 'undefined') {
    // API not supported, return undefined.
    return;
  }
```

samuelgoto (approved):
LGTM

PR #179, opened 2024-10-08
Require JSON serialization checks coming in

### LINKED ISSUE #125: [spec] DigitalCredentialRequestOptions and DigitalCredential objects must be JSON serializable
In [WebAuthn L3](https://github.com/w3c/webauthn/pull/1703), we introduced new methods (`toJSON`, `parseCreationOptionsFromJSON`, and `parseRequestOptionsFromJSON`) for serializing responses and for deserializing requests.

This both improves the developer experience and also allows us to reference these methods directly in the CTAP specification for hybrid transports, instead of defining the encoding and structure there.

timcappalli:
Closes #125

- [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/53526) - and tested by IDL harness. 

Implementation commitment:

- [x] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=295018)
- [ ] [Chromium](https://crbug.com/427885787)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/179.html" title="Last updated on Jul 2, 2025, 3:18 AM UTC (e566aaf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/179/cd2c274...e566aaf.html" title="Last updated on Jul 2, 2025, 3:18 AM UTC (e566aaf)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -231,9 +231,6 @@
       the [=digital credential/request data=] to be handled by the holder's
       software, such as a digital wallet.
     </p>
-    <p>
-      The object MUST only contain [=JSON type=] values.
-    </p>
     <h2>
       The `DigitalCredential` interface
     </h2>
--- a/index.html
+++ b/index.html
@@ -231,6 +231,9 @@
       the [=digital credential/request data=] to be handled by the holder's
       software, such as a digital wallet.
     </p>
+    <p>
+      The object MUST only contain [=JSON type=] values.
+    </p>
     <h2>
       The `DigitalCredential` interface
     </h2>
@@ -252,6 +255,7 @@
     <pre class="idl">
     [Exposed=Window, SecureContext]
     interface DigitalCredential : Credential {
+      [Default] object toJSON();
       readonly attribute DOMString protocol;
       readonly attribute object data;
     };
@@ -306,6 +310,9 @@
       {{DigitalCredentialRequestOptions/requests}} member.
       </li>
       <li>If |requests| is empty, [=exception/throw=] a {{TypeError}}.
+      </li>
+      <li>If |requests| does not contain [=JSON type=] values,
+        [=exception/throw=] a {{TypeError}}.
       </li>
       <li>
         <aside class="issue">
--- a/index.html
+++ b/index.html
@@ -308,8 +308,13 @@
       </li>
       <li>If |requests| is empty, [=exception/throw=] a {{TypeError}}.
       </li>
-      <li>If |requests| does not contain [=JSON type=] values,
-        [=exception/throw=] a {{TypeError}}.
+      <li>
+        [=List/For each=] |request| of |requests|:
+        <ol>
+          <li>
+            [=serialize a JavaScript value to a JSON string|Serialize=] |request| to a JSON string.
+          </li>
+        </ol>
       </li>
       <li>
         <aside class="issue">
--- a/index.html
+++ b/index.html
@@ -312,7 +312,7 @@
         [=List/For each=] |request| of |requests|:
         <ol>
           <li>
-            [=serialize a JavaScript value to a JSON string|Serialize=] |request| to a JSON string.
+            [=serialize a JavaScript value to a JSON string|Serialize=] |request| to a JSON string. [=exception/throw|Re-throw=] any [=exception=].
           </li>
         </ol>
       </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -231,6 +231,9 @@ <h3>
       the [=digital credential/request data=] to be handled by the holder's
       software, such as a digital wallet.
     </p>
+    <p>
+      The object MUST only contain [=JSON type=] values.
+    </p>
```
marcoscaceres, line 236: We probably don't need this. Serialization will throw automatically if it's not the case. 

```suggestion
```

index.html
```diff
@@ -307,6 +311,9 @@ <h3>
       </li>
       <li>If |requests| is empty, [=exception/throw=] a {{TypeError}}.
       </li>
+      <li>If |requests| does not contain [=JSON type=] values,
+        [=exception/throw=] a {{TypeError}}.
+      </li>
```
marcoscaceres, line 316: This is not right here... you need to: 

1. [=List/for each=] |request| of  |requests|:
1.1. [=serialize a JavaScript value to a JSON string|Serialize=] |request| a JSON string. Rethrow any exceptions.

marcoscaceres (changes requested):
Couple of small things...

index.html
marcoscaceres: Yes, but it's in the algorithm. It's redundant to state it twice.

marcoscaceres: Where it should be defined is part of the registry rules. That is, for your protocol to be registered, it might be JSON stringifiable. That would let new protocols know that it's a requirement.

index.html
marcoscaceres: @timcappalli, we also would need this on the way out, when `readonly attribute object data;` is set... though there might not be enough spec text in the current spec to add this right now. However, it would be good to add something.

samuelgoto: > readonly attribute object data

The benefit of also doing this on the way back is that we could make this integrate well with `fetch()` ... arguably, we could also this in a separate PR ...

> This is not right here... you need to:

Yeah, we'd still want to do this, I think.

marcoscaceres: No. This is because if the wallet returns some bogus value that is not a JSON *object* (e.g., a string, or a number, will TypeError at this point), then that can't map to `.data` (because it's required to be an *object*). Thus you need to JSON parse the response _before_ it is stored in `.data`... then the developer calling `.toJSON()` walks both the DigitalCredential instance _and_ the `. data`. structure. 

Make sense?

index.html
```diff
@@ -307,6 +308,14 @@ <h3>
       </li>
       <li>If |requests| is empty, [=exception/throw=] a {{TypeError}}.
       </li>
+      <li>
+        [=List/For each=] |request| of |requests|:
+        <ol>
+          <li>
+            [=serialize a JavaScript value to a JSON string|Serialize=] |request| to a JSON string.
```
marcoscaceres, line 315: Note that "serialize a JavaScript value to a JSON string" can throw, so that needs to be dealt with here. 

```suggestion
            [=serialize a JavaScript value to a JSON string|Serialize=] |request| to a JSON string. [=exception/throw|Re-throw=] any [=exception=].
```

For example, a  cyclic structures:

```
const recursiveObj = {};
recursiveObj.self = recursiveObj;
```

marcoscaceres:
@timcappalli, see also https://github.com/WICG/digital-credentials/pull/179#discussion_r1841349351

marcoscaceres:
Blocked on tests and browser bugs.

marcoscaceres:
Blocked on tests.

index.html
```diff
@@ -541,6 +550,14 @@ <h3>
       </li>
       <li>If |requests| is empty, [=exception/throw=] a {{TypeError}}.
       </li>
+      <li>
```
mohamedamir: I added the same logic to issuance as well.

marcoscaceres (approved):

marcoscaceres:
Still blocked on a test.

marcoscaceres:
[Implemented in WebKit](https://github.com/WebKit/WebKit/pull/47472/files#top) and sent over a [test](https://github.com/web-platform-tests/wpt/pull/53526).

PR #193, opened 2024-12-11
Add initial intro text

### LINKED ISSUE #182: [spec] Add initial intro text
The intro is currently "TBW". Need something prior to CGR1.

timcappalli:
Closes #182

Very basic intro text for CGR1

### DIFF
--- a/index.html
+++ b/index.html
@@ -64,8 +64,31 @@
     <h2>
       Introduction
     </h2>
+    <p>
+      This document defines an API enabling a website to request presentation of a verifiable digital credential (VDC).
+    </p>
+    <p>
+      The API design is agnostic to both credential presentation exchange protocols and credential formats. However, this document includes a registry of credential presentation exchange protocols to ensure interoperability.
+    </p>
+    <p>
+      The API is designed to support the following key properties:
+      <ul>
+        <li>
+          By separating the act of requesting from the specific protocol, we can enable flexibility and adaptability in both the protocol and credential formats. This way, the pace of changes in browsers won't hinder progress or block new developments.
+        </li>
+        <li>
+          Require request transparency, enabling user-agent inspection for risk analysis.
+        </li>
+        <li>
+          Assume response opacity (encrypted responses), enabling verifiers and holders to control where potentially sensitive PII is exposed.
+        </li>
+        <li>
+          Websites cannot access any information without user interaction. This API ensures that sites cannot silently query for digital credentials or communicate with wallet providers without the user's active participation and confirmation of each action.
+        </li>
+      </ul>
+    </p>
     <p id="credential-type-examples">
-      There are many types of digital credential that can be presented using
+      There are many types of verifiable digital credentials that can be presented using
       this API. <dfn data-lt="credential type examples">Examples of these
       types</dfn> include:
     </p>


index.html
```diff
@@ -85,6 +108,18 @@ <h2>
...
+    <h3>
+      Multiple User Agents
+    </h3>
+    <p>
+      In the verifiable digital credentials ecosystem, the digital credential provider, 
+      often known as an identity wallet, functions as a user agent. 
+    </p>
+    <p>
+      When integrated with the Digital Credentials API, 
+      there are two user agents involved: the Web Platform and the digital credential provider.
```
MasterKale, line 120: Is "Web Platform" here synonymous with "browser"? Browsers are also called "clients" in similar contexts (e.g. WebAuthn), are other runtimes considered part of "Web Platform" that may prevent a simpler term here?

index.html
```diff
@@ -85,6 +108,18 @@ <h2>
       Cases</a>
       </li>
     </ul>
+    <h3>
+      Multiple User Agents
```
samuelgoto, line 112: I'd encourage avoiding using the term User Agents for wallets ... it is a fairly overloaded term in Web Platform, and from a security threat modeling perspective, the Wallet has a very different position than the Browser, so I'd avoid using the same term to describe both.

timcappalli, line 112: Feedback from 2024-12-11 call: user agent has a very specific meaning in the web platform. We are essentially trying "agent of the user". Either that say or find a new word.

timcappalli: Also 2024-12-11 remove this for now and address it in its own section around roles / entities.

MasterKale (approved):

index.html
```diff
@@ -64,8 +64,31 @@
...
+      This document defines an API enabling a website to request presentation of a verifiable digital credential (VDC).
+    </p>
+    <p>
+      The API design is agnostic to both credential presentation exchange protocols and credential formats. However, this document includes a registry of credential presentation exchange protocols to ensure interoperability.
+    </p>
+    <p>
+      The API is designed to support the following key properties:
+      <ul>
+        <li>
+          By separating the act of requesting from the specific protocol, we can enable flexibility and adaptability in both the protocol and credential formats. This way, the pace of changes in browsers won't hinder progress or block new developments.
```
TallTed, line 77:
```suggestion
          Keep the act of requesting separate from the specific protocol, thereby enabling flexibility and adaptability in both the protocol and credential formats. This way, the pace of change in browsers won't hinder progress nor block new developments.
```

index.html
```diff
@@ -64,8 +64,31 @@
...
+        </li>
+        <li>
+          Require request transparency, enabling user-agent inspection for risk analysis.
+        </li>
+        <li>
+          Assume response opacity (encrypted responses), enabling verifiers and holders to control where potentially sensitive PII is exposed.
+        </li>
+        <li>
+          Websites cannot access any information without user interaction. This API ensures that sites cannot silently query for digital credentials or communicate with wallet providers without the user's active participation and confirmation of each action.
```
TallTed, line 86:
```suggestion
          Forbid websites from accessing any information without user interaction. This API ensures that sites cannot silently query for digital credentials nor communicate with wallet providers without the user's active participation and confirmation of each action.
```

index.html
```diff
@@ -64,8 +64,31 @@
...
+        </li>
+      </ul>
+    </p>
     <p id="credential-type-examples">
-      There are many types of digital credential that can be presented using
+      There are many types of verifiable digital credentials that can be presented using
```
TallTed, line 91:
```suggestion
      There are many types of verifiable digital credential that can be presented using
```

TallTed (changes requested):
Editorial fixes (grammar, punctuation, etc.)

index.html
timcappalli: @TallTed can you explain this one? Singular sounds very weird when read aloud. If I were to replace VDCs with "documents", just to try a different word, I don't think "document" would be correct ("...there are many types of document that can be presented using...")

TallTed: Sorry for delay. 

`"...there are many types of document that can be presented using..."` would be correct.

`"...there are many types of documents that can be presented using..."` would not.

Try changing `types` to `kinds` and `document` to `pizza`. I expect "There are many kinds of `pizzas` that you can order" and "You can order many kinds of `pizzas`" will both sound odd to you. I expect "There are many kinds of `pizza` that you can order" and "You can order many kinds of `pizza`" will both sound right.

If you really want the plural `credentials`, I might suggest `Verifiable digital credentials of many types can be presented...`

index.html
```diff
@@ -64,8 +64,31 @@
     <h2>
       Introduction
     </h2>
+    <p>
+      This document defines an API enabling a website to request presentation of a verifiable digital credential (VDC).
```
marcoscaceres, line 68: Automagically adds the (VDC) and auto titles it globally: 
 
```suggestion
      This document defines an API enabling a website to request presentation of a <dfn data-abbr>Verifiable Digital Credential</dfn>.
```

index.html
```diff
@@ -64,8 +64,31 @@
...
+    </p>
+    <p>
+      The API design is agnostic to both credential presentation exchange protocols and credential formats. However, this document includes a registry of credential presentation exchange protocols to ensure interoperability.
```
marcoscaceres, line 71:
```suggestion
      The API design is agnostic to both credential presentation exchange protocols and credential formats. However, this document includes a [[[##protocol-registry]] to ensure interoperability.
```

index.html
```diff
@@ -64,8 +64,31 @@
...
+    </p>
+    <p>
+      The API is designed to support the following key properties:
```
marcoscaceres, line 74: The word "properties" is too ambiguous.

index.html
```diff
@@ -64,8 +64,31 @@
...
+      <ul>
+        <li>
+          Keep the act of requesting separate from the specific protocol, thereby enabling flexibility and adaptability in both the protocol and credential formats. This way, the pace of change in browsers won't hinder progress nor block new developments.
```
marcoscaceres, line 77: This is not guaranteed and is presumptuous that "browsers ... hinder progress" (which is not ideal).  I prefer not to repeat the discussion of the role of user agent, keeping  users safe, IPC restrictions, etc. 

> " This way, the pace of change in browsers won't hinder progress nor block new developments." 

```suggestion
          Keep the act of requesting separate from the specific protocol, thereby enabling flexibility and adaptability in both the protocol and credential formats. 
```

Also, this is wrong with the respect to the rest of the platform... just because this bit could be seen to move faster (it won't), the rest of the platform still adheres to the development that _quote_ "hinder progress" _unquote_. 

This API is even affected right now... for example, AbortController handling is not interoperable right now.

index.html
```diff
@@ -64,8 +64,31 @@
...
+        </li>
+        <li>
+          Require request transparency, enabling user-agent inspection for risk analysis.
+        </li>
+        <li>
+          Assume response opacity (encrypted responses), enabling verifiers and holders to control where potentially sensitive PII is exposed.
```
marcoscaceres, line 83:
```suggestion
          Assume response opacity (encrypted responses), enabling verifiers and holders to control where potentially sensitive personally identifiable information is exposed.
```

index.html
```diff
@@ -64,8 +64,31 @@
...
+        </li>
+        <li>
+          Forbid websites from accessing any information without user interaction. This API ensures that sites cannot silently query for digital credentials nor communicate with wallet providers without the user's active participation and confirmation of each action.
```
marcoscaceres, line 86:
```suggestion
          Forbid websites from accessing any information without user interaction, ensuring that sites cannot silently query for digital credentials nor communicate with wallet providers without the user's active participation and confirmation of each action.
```

index.html
```diff
@@ -64,8 +64,31 @@
...
+        </li>
+      </ul>
+    </p>
     <p id="credential-type-examples">
-      There are many types of digital credential that can be presented using
+      Verifiable digital credentials of many types can be presented using
```
marcoscaceres, line 91:
```suggestion
      [=Verifiable credentials=] of many types can be presented using
```

index.html
timcappalli: goals?

index.html
```diff
@@ -64,8 +64,31 @@
     <h2>
       Introduction
     </h2>
+    <p>
+      This document defines an API enabling a website to request presentation of a <dfn data-abbr>verifiable digital credential</dfn> (VDC).
```
samuelgoto, line 68:
```suggestion
      This document defines an API enabling a website to request presentation of a <dfn data-abbr>digital credential</dfn> (VDC).
```

Is "verifiable" necessary here? Can't we assume that a property of a digital credential is that it is verifiable? I'm a little worried this will be read as equivalent to a W3C VC (which is not, e.g. in the case of an MDoc?).

index.html
samuelgoto: +1 to "goals", seems more common in these types of docs.

index.html
samuelgoto: I think removing this would be OK ... I think, perhaps, a nicer (and more accepted.) way to put this is in terms of extensibility ... let me try to rephrase this in a suggestion:

```suggestion
          Keep the act of requesting separate from the specific protocol, thereby enabling the extensibility of the protocol and credential formats to be done outside of the browser.
```

index.html
timcappalli: verifiable digital credentials is being used as the umbrella term in many corners of the industry to cover all flavors including W3C VCs, mdocs, SD-JWT VCs, etc. 

If there is concern, we can add a more formal definition.

index.html
marcoscaceres: Suggestion sounds good.

index.html
```diff
@@ -64,8 +64,31 @@
...
+    </p>
+    <p>
+      The API design is agnostic to both credential presentation exchange protocols and credential formats. However, this document includes a [[[#protocol-registry]]] to ensure interoperability.
```
RByers, line 71:
```suggestion
      The API design is agnostic to both credential presentation exchange protocols and credential formats. However, to promote interoperability this document includes a [[[#protocol-registry]]].
```

IMHO we're very far from being able to "ensure interoperability" 😄

index.html
```diff
@@ -64,8 +64,31 @@
...
+    </p>
+    <p>
+      The API is designed to support the following goals:
```
RByers, line 74: Perhaps also worth adding some other key things we know make this attractive over custom schemes?
 - Enables platform-provided credential selection UX for when multiple wallet applications have credentials which match a request
 - Enables platforms to potentially provide secure cross-device requests with proximity checks

RByers (commented):
Looks good just a couple suggestions

index.html
Ryang03: > > Verifiable digital credentials is being used as the umbrella term in many corners of the industry to cover all flavors including W3C VCs, mdocs, SD-JWT VCs, AnonCreds, etc.
> 
> Ah, interesting, I didn't quite get that impression (but I'm certainly not as well connected). I get the impression that everybody seems to be aligning more on "digital credentials" than "verifiable digital credentials". For example, is the OpenID4VP WG using "verifiable digital credentials"?

Hey Sam - we have started using VDC as the go to term. Mostly because it adequately describes the characteristics of what we are looking for. There are many existing digital credentials that are no cryptographically verifiable (think traditional login credentials, etc.) and the term VC is very tightly associated with a specific data model. VDC is distinct and reasonably descriptive. That doesn't mean it HAS to be used here, but we would encourage its use. See this for more on it:

https://www.nist.gov/blogs/cybersecurity-insights/digital-identities-getting-know-verifiable-digital-credential-ecosystem

RByers (approved):
LGTM, thanks!

index.html
marcoscaceres:
```suggestion
      This document defines an API enabling a website to request presentation of a  [=digital credentials=].
```

marcoscaceres: filed https://github.com/WICG/digital-credentials/issues/198 in case we want to enhance out definition of a "digital credential" to explicitly talk about being **verifiable**.

index.html
marcoscaceres: I'm still not loving the "the pace of change in browsers won't hinder progress" being there.

marcoscaceres: I think it should just be: 

```suggestion
          Keep the act of requesting separate from the specific protocol, thereby enabling the extensibility of the protocol and credential formats. 
```

index.html
```diff
@@ -64,8 +64,37 @@
...
+      The API design is agnostic to both credential presentation exchange protocols and credential formats. However, to promote interoperability this document includes a [[[#protocol-registry]]].
+    </p>
+    <p>
+      The API is designed to support the following goals:
+      <ul>
+        <li>
+          Keep the act of requesting separate from the specific protocol, thereby enabling flexibility and adaptability in both the protocol and credential formats. This way, the pace of change in browsers won't hinder progress nor block new developments.
+        </li>
+        <li>
+          Require request transparency, enabling user-agent inspection for risk analysis.
```
marcoscaceres, line 80:
```suggestion
          Require request to be unencrypted, enabling user-agent inspection for risk analysis.
```

index.html
```diff
@@ -64,8 +64,37 @@
...
+        </li>
+        <li>
+          Assume response opacity (encrypted responses), enabling verifiers and holders to control where potentially sensitive personally identifiable information is exposed.
+        </li>
+        <li>
+          Forbid websites from accessing any information without user interaction, ensuring that sites cannot silently query for digital credentials nor communicate with wallet providers without the user's active participation and confirmation of each action.
```
marcoscaceres, line 86:
```suggestion
          Require user interaction for requests , ensuring that sites cannot silently query for digital credentials nor communicate with wallet providers without the user's active participation and confirmation of each action.
```

marcoscaceres: This should really be in the privacy considerations section, if not already there.

index.html
```diff
@@ -64,8 +64,37 @@
...
+        </li>
+        <li>
+          Enable platform-provided credential selection UX when multiple wallet applications have credentials which match a request.
```
marcoscaceres, line 89:
```suggestion
          Enable platform-provided credential selection UX when multiple wallet applications have credentials that match a request.
```

marcoscaceres (changes requested):
Still insisting that browsers are not the issue with extensibility. 

Otherwise, looks ok. 

We should consider linking things "request" to [=digital credential/presentation request|request=] and so on.

index.html
samuelgoto: That works for me

marcoscaceres (approved):

Issue #168, opened 2024-09-12
Add a way to check what protocols are supported


marcoscaceres:
It's currently not possible to check what protocols the browser supports so developers won't know if calling .get() will immediately fail (as it requires user activation).

As such, we should then add a. static setlike`.supportedProtocols` With the sequence, a developer can easily filter with standard Set methods.

### Proposed API addition

```WebIDL
interface DigitalCredentialsSupportedProtocols {
  readonly setlike<DOMString>; // user agent pre-populates
};

partial interface DigitalCredential {
   readonly attribute DigitalCredentialsSupportedProtocols supportedProtocols;
}
```

Usage:

```JS
if (DigitalCredential.supportedProtocols.has("openid4vp")) {
  // let's make an openid4vp request
}

// Or you can do...
const supported = Array.from(DigitalCredential.supportedProtocols).filter(typesTheRPSupports);

// Or whatever developers want:
for (const supported in DigitalCredential.supportedProtocols.values()) {
   // do things with supported
}

const requests = [...DigitalCredential.supportedProtocols].map(toRequestWeKnowHowToMake);

// or even... 
switch (true) {
   case DigitalCredential.supportedProtocols.has("openid4vp"):
      makeOpenIDRequest(data);
      break;
   case DigitalCredential.supportedProtocols.has("whatever"):
      makeWhateverRequest(data);
      break;
   default:
     throw TypeError("Oh noes! they don't support our favorite protocol!")
}
```

That gives a ton of flexibility. You can even use it will all the new fancy JavaScript set comparison operations:

```JS
// DigitalCredential.supportedProtocols is a Set
const supportedProtocols = DigitalCredential.supportedProtocols;

// Example Set of protocols to compare
const protocolsToCheck = new Set(['openid4vp', 'someOtherProtocol']);

// Union: Combining both sets
const union = supportedProtocols.union(protocolsToCheck);
console.log(union); // Set { 'openid4vp', 'someOtherProtocol', ... }

// Intersection: Getting common protocols between the sets
const intersection = supportedProtocols.intersection(protocolsToCheck);
console.log(intersection); // Set { 'openid4vp' }

// Difference: Getting protocols supported by DigitalCredential but not in protocolsToCheck
const difference = supportedProtocols.difference(protocolsToCheck);
console.log(difference); // Set { ... } (protocols supported by DigitalCredential but not in protocolsToCheck)

// Symmetric Difference: Getting protocols that are in either set, but not in both
[...]

PR #221, opened 2025-04-18
Add userAgentAllowsProtocol static method

### LINKED ISSUE #219: Add a means for developers to check protocols allowed by user agents
Based on discussion at the [2025-04-11 hybrid meeting](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-04-11-hybrid-notes.md#minutes-afternoon) and in #200, this issue proposes the creation of a new static method called `isProtocolSupportedByClient`, which takes a protocol identifier string as input, and returns a promise that resolves with a boolean. 

This only represents whether the client (e.g. browser) understand the request enough to pass it southbound to the platform. It does NOT represent whether a credential provider/wallet is available and can support the protocol.

```webidl
partial interface DigitalCredential {
  static Promise<boolean> isProtocolSupportedByClient(DOMString protocol);
};
```

As an aside, as I said in the F2F, I think we will regret the ergonomics of this (vs a more general getClientCapabilities() method) long term, just as we did with WebAuthn 6+ years later, but I'd like to get this moving forward as this is important for developers.

timcappalli:
Closes #219 

- [X] Modified Web platform tests (https://github.com/web-platform-tests/wpt/pull/53089)

Implementation commitment:

- [X] WebKit ([link to issue](https://bugs.webkit.org/show_bug.cgi?id=294352))
- [X] Chromium ([link to issue](https://issues.chromium.org/issues/424208782))
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/221.html" title="Last updated on Jun 25, 2025, 11:48 PM UTC (063b6b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/221/c5ac0ef...063b6b6.html" title="Last updated on Jun 25, 2025, 11:48 PM UTC (063b6b6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -496,9 +496,9 @@
       The {{DigitalCredential/clientAllowsProtocol()}} method allows digital
       credential verifiers and issuers to determine which 
       [=digital credential/exchange protocols=] the client supports for
-      interactions with credential providers.
+      interactions with [=credential providers=].
     </p>
-    <p>
+    <p class="note">
       This method does not convey [=digital credential/request protocol=] or 
       [=digital credential/issuance protocol=] support in the [=client device=]
       or [=credential provider=].
@@ -514,6 +514,26 @@
         indicate whether the browser supports distributing requests with that
         protocol to underlying platform or provider.
     </p>
+    <h4>
+      clientAllowsProtocol algorithm
+    </h4>
+    <p>
+      When this method is invoked, the user agent MUST execute the following
+      algorithm:
+    </p>
+    <ol class="algorithm">
+      <li>
+        If |protocol| contains characters other than lowercase ASCII letters,
+        digits, or hypens, throw a "{{ConstraintError}}" {{DOMException}}.
+      </li>
+      <li>
+        If |protocol| is known and allowed by the user agent, 
+        return <code>true</code>.
+      </li>
+      <li>
+        Else, return <code>false</code>.
+      </li>
+    </ol>
     <h3>
       [[\DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)
       internal method
--- a/index.html
+++ b/index.html
@@ -462,7 +462,7 @@
     interface DigitalCredential : Credential {
       readonly attribute DOMString protocol;
       [SameObject] readonly attribute object data;
-      static boolean clientAllowsProtocol(DOMString protocol);
+      static boolean userAgentAllowsProtocol(DOMString protocol);
     };
     </pre>
     <p>
@@ -490,12 +490,13 @@
     </h2>
     <aside class="issue" data-number="65"></aside>
     <h3>
-      The `<dfn>clientAllowsProtocol()</dfn>` method
+      The `<dfn>userAgentAllowsProtocol()</dfn>` method
     </h3>
     <p>
-      The {{DigitalCredential/clientAllowsProtocol()}} method allows digital
+      The {{DigitalCredential/userAgentAllowsProtocol()}} method allows digital
       credential verifiers and issuers to determine which 
-      [=digital credential/exchange protocols=] and [=digital credential/issuance protocols=] the client supports for
+      [=digital credential/exchange protocols=] and
+      [=digital credential/issuance protocols=] the user agent supports for
       interactions with [=credential providers=].
     </p>
     <p class="note">
@@ -515,7 +516,7 @@
         protocol to underlying platform or provider.
     </p>
     <h4>
-      clientAllowsProtocol algorithm
+      userAgentAllowsProtocol algorithm
     </h4>
     <p>
       When this method is invoked, the user agent MUST execute the following
--- a/index.html
+++ b/index.html
@@ -619,8 +619,7 @@
     </p>
     <p>
       This method accepts a [=DOMString=] that uniquely specifies the
-      [=digital credential/exchange protocol=] and returns a [=Promise=] that
-      resolves to a Boolean value.
+      [=digital credential/exchange protocol=] and returns a Boolean value.
     </p>
     <pre class="idl">
       partial interface DigitalCredential {
--- a/index.html
+++ b/index.html
@@ -499,8 +499,9 @@
       interactions with credential providers.
     </p>
     <p>
-      This method does not convey protocol support in the underlying platform
-      or credential provider.
+      This method does not convey [=digital credential/request protocol=] or 
+      [=digital credential/issuance protocol=] support in the [=client device=]
+      or [=credential provider=].
     </p>
     <p>
         User agents SHOULD NOT vary the response value based on any information
--- a/index.html
+++ b/index.html
@@ -462,6 +462,7 @@
     interface DigitalCredential : Credential {
       readonly attribute DOMString protocol;
       [SameObject] readonly attribute object data;
+      static boolean clientAllowsProtocol(DOMString protocol);
     };
     </pre>
     <p>
@@ -488,6 +489,30 @@
       Integration with Credential Management API
     </h2>
     <aside class="issue" data-number="65"></aside>
+    <h3>
+      The `<dfn>clientAllowsProtocol()</dfn>` method
+    </h3>
+    <p>
+      The {{DigitalCredential/clientAllowsProtocol()}} method allows digital
+      credential verifiers and issuers to determine which 
+      [=digital credential/exchange protocols=] the client supports for
+      interactions with credential providers.
+    </p>
+    <p>
+      This method does not convey protocol support in the underlying platform
+      or credential provider.
+    </p>
+    <p>
+        User agents SHOULD NOT vary the response value based on any information
+        about availability of hardware, presence or configuration of software,
+        wallets, credential providers, or digital credentials, or user
+        configuration or preferences. If the response value varied, the user
+        agent would introduce risks both of fingerprinting and of silently
+        revealing other details about user behavior or configuration. The
+        response value SHOULD vary only by user agent major version and
+        indicate whether the browser supports distributing requests with that
+        protocol to underlying platform or provider.
+    </p>
     <h3>
       [[\DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)
       internal method
@@ -594,38 +619,6 @@
       <dfn class="export" data-dfn-for="DigitalCredential">[[\discovery]]</dfn>
       whose value is "remote".
     </p>
-    <h3>
-      `clientAllowsProtocol()` Method
-    </h3>
-    <p>
-      Digital credential verifiers and issuers use this static method to
-      determine which [=digital credential/exchange protocols=] the client
-      supports for interactions with credential providers.
-    </p>
-    <p>
-      This method does not convey protocol support in the underlying platform
-      or credential provider.
-    </p>
-    <p>
-        User agents SHOULD NOT vary the response value based on any information
-        about availability of hardware, presence or configuration of software,
-        wallets, credential providers, or digital credentials, or user
-        configuration or preferences. If the response value varied, the user
-        agent would introduce risks both of fingerprinting and of silently
-        revealing other details about user behavior or configuration. The
-        response value SHOULD vary only by user agent major version and
-        indicate whether the browser supports distributing requests with that
-        protocol to underlying platform or provider.
-    </p>
-    <p>
-      This method accepts a [=DOMString=] that uniquely specifies the
-      [=digital credential/exchange protocol=] and returns a Boolean value.
-    </p>
-    <pre class="idl">
-      partial interface DigitalCredential {
-        static boolean clientAllowsProtocol(DOMString protocol);
-      };
-    </pre>
     <section class="informative">
       <h3>
         User permission
--- a/index.html
+++ b/index.html
@@ -607,8 +607,8 @@
       or credential provider.
     </p>
     <p>
-      This method accepts a single argument, a [=DOMString=] that specifies the
-      [=digital credential/exchange protocol=], and returns a [=Promise=] that
+      This method accepts a [=DOMString=] that specifies the
+      [=digital credential/exchange protocol=] and returns a [=Promise=] that
       resolves to a Boolean value.
     </p>
     <pre class="idl">
--- a/index.html
+++ b/index.html
@@ -607,6 +607,9 @@
       or credential provider.
     </p>
     <p>
+        User agents SHOULD NOT vary the response value based on any information about availability of hardware, presence or configuration of software, wallets or credentials, or user configuration or preferences. If the response value varied, the user agent would introduce risks both of fingerprinting and of silently revealing other details about user behavior or configuration. The response value SHOULD vary only by user agent major version and indicate whether the browser supports distributing requests with that protocol to underlying platform or provider.
+    </p>
+    <p>
       This method accepts a [=DOMString=] that uniquely specifies the
       [=digital credential/exchange protocol=] and returns a [=Promise=] that
       resolves to a Boolean value.
--- a/index.html
+++ b/index.html
@@ -515,9 +515,6 @@
         indicate whether the browser supports distributing requests with that
         protocol to underlying platform or provider.
     </p>
-    <h4>
-      userAgentAllowsProtocol algorithm
-    </h4>
     <p>
       When this method is invoked, the user agent MUST execute the following
       algorithm:
--- a/index.html
+++ b/index.html
@@ -607,7 +607,7 @@
       or credential provider.
     </p>
     <p>
-        User agents SHOULD NOT vary the response value based on any information about availability of hardware, presence or configuration of software, wallets or credentials, or user configuration or preferences. If the response value varied, the user agent would introduce risks both of fingerprinting and of silently revealing other details about user behavior or configuration. The response value SHOULD vary only by user agent major version and indicate whether the browser supports distributing requests with that protocol to underlying platform or provider.
+        User agents SHOULD NOT vary the response value based on any information about availability of hardware, presence or configuration of software, wallets, credential providers, or digital credentials, or user configuration or preferences. If the response value varied, the user agent would introduce risks both of fingerprinting and of silently revealing other details about user behavior or configuration. The response value SHOULD vary only by user agent major version and indicate whether the browser supports distributing requests with that protocol to underlying platform or provider.
     </p>
     <p>
       This method accepts a [=DOMString=] that uniquely specifies the
--- a/index.html
+++ b/index.html
@@ -499,7 +499,7 @@
       interactions with [=credential providers=].
     </p>
     <p class="note">
-      This method does not convey [=digital credential/request protocol=] or 
+      This method does not convey [=digital credential/exchange protocol=] or 
       [=digital credential/issuance protocol=] support in the [=client device=]
       or [=credential provider=].
     </p>
--- a/index.html
+++ b/index.html
@@ -529,9 +529,6 @@
       <li>
         Return `true` if the user agent allows |protocol|, `false` otherwise.
       </li>
-      <li>
-        Else, return <code>false</code>.
-      </li>
     </ol>
     <h3>
       [[\DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)
--- a/index.html
+++ b/index.html
@@ -490,7 +490,7 @@
     </h2>
     <aside class="issue" data-number="65"></aside>
     <h3>
-      The `<dfn>userAgentAllowsProtocol()</dfn>` method
+      The <dfn>userAgentAllowsProtocol()</dfn> method
     </h3>
     <p>
       The {{DigitalCredential/userAgentAllowsProtocol()}} method allows digital
--- a/index.html
+++ b/index.html
@@ -504,7 +504,7 @@
       or [=credential provider=].
     </p>
     <p>
-        User agents SHOULD NOT vary the response value based on any information
+        User agents MUST NOT vary the response value based on any information
         about availability of hardware, presence or configuration of software,
         wallets, credential providers, or digital credentials, or user
         configuration or preferences. If the response value varied, the user
--- a/index.html
+++ b/index.html
@@ -527,8 +527,7 @@
         digits, or hypens, throw a "{{ConstraintError}}" {{DOMException}}.
       </li>
       <li>
-        If |protocol| is known and allowed by the user agent, 
-        return <code>true</code>.
+        Return `true` if the user agent allows |protocol|, `false` otherwise.
       </li>
       <li>
         Else, return <code>false</code>.
--- a/index.html
+++ b/index.html
@@ -607,7 +607,15 @@
       or credential provider.
     </p>
     <p>
-        User agents SHOULD NOT vary the response value based on any information about availability of hardware, presence or configuration of software, wallets, credential providers, or digital credentials, or user configuration or preferences. If the response value varied, the user agent would introduce risks both of fingerprinting and of silently revealing other details about user behavior or configuration. The response value SHOULD vary only by user agent major version and indicate whether the browser supports distributing requests with that protocol to underlying platform or provider.
+        User agents SHOULD NOT vary the response value based on any information
+        about availability of hardware, presence or configuration of software,
+        wallets, credential providers, or digital credentials, or user
+        configuration or preferences. If the response value varied, the user
+        agent would introduce risks both of fingerprinting and of silently
+        revealing other details about user behavior or configuration. The
+        response value SHOULD vary only by user agent major version and
+        indicate whether the browser supports distributing requests with that
+        protocol to underlying platform or provider.
     </p>
     <p>
       This method accepts a [=DOMString=] that uniquely specifies the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -400,6 +400,28 @@ <h3>
       <dfn class="export" data-dfn-for="DigitalCredential">[[\discovery]]</dfn>
       whose value is "remote".
     </p>
+    <h3>
+      `isProtocolSupportedByClient()` Method
+    </h3>
+    <p>
+      Digital credential verifiers and issuers use this static method to
+      determine which [=digital credential/exchange protocols=] the client
+      understands.
```
RByers, line 409: I think what's important here is not whether the client "understands" the protocol but whether it will permit it to be passed on to the platform. Chromium current plans to pass on anything without necessarily understanding most things. What if we said "the client supports sending to credential providers"?

RByers (commented):
Thanks for doing this, I think it probably is important and valuable given WebKit's plans to not support some popular protocols. Should we perhaps define this to call an internal function, which then we also call from the main processing parts of the spec and return a well-defined error for when the result is 'false'? @marcoscaceres @mohamedamir WDYT?

index.html
timcappalli: Yeah, I actually had that originally and can't remember why I changed to understands. I'll switch it back.

index.html
```diff
@@ -400,6 +400,28 @@ <h3>
       <dfn class="export" data-dfn-for="DigitalCredential">[[\discovery]]</dfn>
       whose value is "remote".
     </p>
+    <h3>
+      `isProtocolSupportedByClient()` Method
```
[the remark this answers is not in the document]
timcappalli: I think that would be a different discussion and likely another static method. Please open an issue.

index.html
```diff
@@ -400,6 +400,28 @@ <h3>
...
+    </h3>
+    <p>
+      Digital credential verifiers and issuers use this static method to
+      determine which [=digital credential/exchange protocols=] the client
+      supports sending to credential providers.
```
MasterKale, line 409: This calls out verifiers **and** issuers as users of this method. I don't have a problem with this goal, but does it make sense for this method to return `true` to an **issuer** when a specified protocol is presentation-only?

I'm loathe to suggest this, but might we need to have separate "`isPresentationProtocolSupportedByClient()`" and "`isIssuanceProtocolSupportedByClient()`"? Or perhaps a second argument of `"presentation"`/`"issuance"` to the currently proposed method to help provide a more nuanced support query capability?

MasterKale:
```
await isProtocolSupportedByClient("openid4vp", "presentation"); // true
await isProtocolSupportedByClient("annexc", "issuance"); // false
```

index.html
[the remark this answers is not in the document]
c2bo: I agree this should be another issue/discussion, but just as a general reaction:
Filtering on a credential type level surprises me quite a bit and I'd like to better understand why that has come up. I don't think DC API should implement or encourage any kind of filtering on that (credential/presentation) level at all.

index.html
TallTed, line 409:
```suggestion
      supports for interactions with credential providers.
```

index.html
```diff
@@ -400,6 +400,28 @@ <h3>
...
+    </p>
+    <p>
+      This method does not convey protocol support in the underlying platform
+      or credential provider.
+    </p>
+    <p>
+      This method accepts a single argument, a [=DOMString=] that specifies the
+      [=digital credential/exchange protocol=], and returns a [=Promise=] that
```
TallTed, line 417:
```suggestion
      This method accepts a single argument &mdash; a [=DOMString=] that specifies the
      [=digital credential/exchange protocol=] &mdash; and returns a [=Promise=] that
```

TallTed (changes requested):
Minor text changes for clarity

MasterKale (approved):
This seems like a good compromise across verifiers, browsers, and platforms 🤔

TallTed (approved):

marcoscaceres:
Hey all, please see https://github.com/w3c-fedid/digital-credentials/issues/168 ... I proposed a different way of doing this there that is, IMO, more developer friendly and avoid the "support" question... that the browser recognizes the protocol string does not imply "support" for it. 

I suggest we refactor this pull request to align more fully with https://github.com/w3c-fedid/digital-credentials/issues/168

marcoscaceres (changes requested):
Please see https://github.com/w3c-fedid/digital-credentials/issues/168

index.html
```diff
@@ -400,6 +400,28 @@ <h3>
...
+      Digital credential verifiers and issuers use this static method to
+      determine which [=digital credential/exchange protocols=] the client
+      supports for interactions with credential providers.
+    </p>
+    <p>
+      This method does not convey protocol support in the underlying platform
+      or credential provider.
+    </p>
+    <p>
+      This method accepts a [=DOMString=] that specifies the
```
c2bo, line 416:
```suggestion
      This method accepts a [=DOMString=] that uniquely specifies the
```

npdoty:
Is this intended to only reveal information on what the browser version supports passing through and specifically not to reveal any details of what wallets are available, what the user has configured, what credentials they have, whether they have another device that supports credentials, etc.? (I know I raise this question every time it comes up, but we haven't written down the answer, so I keep asking.)

If so, we should add normative requirements that indicate that.

timcappalli:
> Is this intended to only reveal information on what the browser version supports passing through and specifically not to reveal any details of what wallets are available, what the user has configured, what credentials they have, whether they have another device that supports credentials, etc.? (I know I raise this question every time it comes up, but we haven't written down the answer, so I keep asking.)
> 
> If so, we should add normative requirements that indicate that.

Essentially: "can you pass a request using protocol XYZ to the next component"

There is no possible way for it to reveal any details about wallets or credentials. It just returns a boolean given a protocol identifier.

npdoty:
Browsers could modify the response based on some knowledge of wallet availability or configuration. I think we should just specify that the UA shouldn't use that kind of knowledge for a customized value, they should only describe browser-level support without user/device/app/wallet configuration detail.

timcappalli:
> Browsers could modify the response based on some knowledge of wallet availability or configuration

Not sure I follow. The browser is the only party handling this request and response.

npdoty:
As in, a browser vendor trying to implement this section could try to use their knowledge of system/app/wallet/user-configuration properties to modify the return value, and I think this API should tell the browser vendor implementer not to do that.

timcappalli:
Gotcha. I'm not sure how we write something like that given each browser has their own method of determining what the response should be. Do you have some suggested text?

npdoty:
I can try to propose something (tomorrow, likely) that would set some SHOULD NOT/MUST NOT conditions to try to capture what I think is the agreement.

index.html
```diff
@@ -400,6 +400,28 @@ <h3>
...
+    <p>
+      Digital credential verifiers and issuers use this static method to
+      determine which [=digital credential/exchange protocols=] the client
+      supports for interactions with credential providers.
+    </p>
+    <p>
+      This method does not convey protocol support in the underlying platform
+      or credential provider.
+    </p>
+    <p>
```
npdoty, line 415:
```suggestion
    <p>
        User agents SHOULD NOT vary the response value based on any information about availability of hardware, presence or configuration of software, wallets or credentials, or user configuration or preferences. If the response value varied, the user agent would introduce risks both of fingerprinting and of silently revealing other details about user behavior or configuration. The response value SHOULD vary only by user agent major version and indicate whether the browser supports distributing requests with that protocol to underlying platform or provider.
    </p>
    <p>
```

npdoty: This is just a starting point proposed text for trying to capture the idea that this is just a convenience method indicating browser version support, rather than details to the site about what the user has configured.

timcappalli: That sounds great @npdoty. thanks!

MasterKale:
From WG meeting @ May 28: The name of the method was chosen for brevity, but the intent behind it is closer to "canRequestForProtocolReachWallets()." The current `isProtocolSupportedByClient()` name is lower priority and can be changed. Arguments were made for and against keeping it an asynchronous method; we will continue to discuss this in future meetings.

MasterKale:
I believe I've found some guidance on how to determine when to define a method as asynchronous vs synchronous. The W3C's Web Platform Design Principles suggest to me that it's best to go with async by default:

https://www.w3.org/TR/design-principles/#async-by-default

> Due to current limitations, some specific areas of the web platform lack asynchronous support. Therefore having support for synchronous methods in addition to asynchronous methods can be beneficial for usability. Consider these cases an exception, and have a clear path for deprecation of the synchronous methods as web platform capabilities evolve.

We should probably take this into account as we continue to discuss the shape of this method.

MasterKale:
cc @marcoscaceres on this ⬆️

npdoty:
Since this method isn't intended to invoke user permission or a detailed request to another resource (in fact, I believe we are specifically prohibiting that, because it's just supposed to be a browser support check, not a check of available resources), I think it makes more sense to stay synchronous:
https://www.w3.org/TR/design-principles/#synchronous

c2bo (approved):

index.html
```diff
@@ -400,6 +400,38 @@ <h3>
...
+        agent would introduce risks both of fingerprinting and of silently
+        revealing other details about user behavior or configuration. The
+        response value SHOULD vary only by user agent major version and
+        indicate whether the browser supports distributing requests with that
+        protocol to underlying platform or provider.
+    </p>
+    <p>
+      This method accepts a [=DOMString=] that uniquely specifies the
+      [=digital credential/exchange protocol=] and returns a Boolean value.
+    </p>
```
marcoscaceres: We don't need this as it's already "what WebIDL says". 

```suggestion
```

Progdrasil (approved):

index.html
```diff
@@ -400,6 +400,38 @@ <h3>
...
+    <pre class="idl">
+      partial interface DigitalCredential {
```
marcoscaceres: Move this to the main interface.

timcappalli:
To Do: Add test (e.g. `com.banana.protocol` = `false`)

index.html
```diff
@@ -326,6 +327,30 @@ <h3>
...
       object types.
     </p>
+    <h3>
+      The `<dfn>clientAllowsProtocol()</dfn>` method
+    </h3>
+    <p>
+      The {{DigitalCredential/clientAllowsProtocol()}} method allows digital
+      credential verifiers and issuers to determine which 
+      [=digital credential/exchange protocols=] the client supports for
+      interactions with credential providers.
```
marcoscaceres: As we don't have a notion on a "client", it doesn't make sense to talk about a "client". 

```suggestion
      The {{DigitalCredential/allowsProtocol()}} method enables
      a [=verifier=] to determine if the [=user agent=] is capable of
      forwarding [=request data=] to a [=holder=] based solely on the               
      [=digital credential/exchange protocol=] or [=digital credential/issuance protocol=] string.
```

index.html
```diff
@@ -301,6 +301,7 @@ <h2>
     interface DigitalCredential : Credential {
       readonly attribute DOMString protocol;
       [SameObject] readonly attribute object data;
+      static boolean clientAllowsProtocol(DOMString protocol);
```
marcoscaceres: See rationale for change below...

```suggestion
      static boolean allowsProtocol(DOMString protocol);
```

index.html
```diff
@@ -326,6 +327,30 @@ <h3>
       credential's response data. It contains the subset of JSON-parseable
       object types.
     </p>
+    <h3>
+      The `<dfn>clientAllowsProtocol()</dfn>` method
```
```suggestion
      The static `<dfn>allowsProtocol()</dfn>` method
```

index.html
```diff
@@ -326,6 +327,30 @@ <h3>
...
+    </h3>
+    <p>
+      The {{DigitalCredential/clientAllowsProtocol()}} method allows digital
+      credential verifiers and issuers to determine which 
+      [=digital credential/exchange protocols=] the client supports for
+      interactions with credential providers.
+    </p>
+    <p>
+      This method does not convey protocol support in the underlying platform
+      or credential provider.
```
```suggestion
      This method does not convey [=digital credential/request protocol=] or [=digital credential/issuance protocol=] support in the underlying platform or [=holder=].
```

index.html
```diff
@@ -326,6 +327,30 @@ <h3>
...
+    </p>
+    <p>
```
marcoscaceres, line 507: This reads like a note... so maybe should be a note? 

What's missing here is the actual algorithm for processing the string. The main thing is that it needs to enforce a well defined identity protocol string structure, as per https://github.com/w3c-fedid/digital-credentials/issues/270

index.html
```diff
@@ -326,6 +327,30 @@ <h3>
...
+        User agents SHOULD NOT vary the response value based on any information
+        about availability of hardware, presence or configuration of software,
+        wallets, credential providers, or digital credentials, or user
+        configuration or preferences. If the response value varied, the user
+        agent would introduce risks both of fingerprinting and of silently
```
marcoscaceres: There's a mix of privacy considerations and processing behavior here... the whole "value SHOULD vary only by user agent major version" is not needed in my opinion: browser versions are not tied to engine versions, so this doesn't really make sense. In fact, a browser may use an entirely different engine on one OS, but still have the same version number (i.e., version numbers are mostly just marketing). 

The algorithm should determine the behavior of what is returned here, and let's keep it simple. 

1. Check for string structure, as defined by " identity protocol string" (let's rename this maybe too) - that is, we need clear rules for what to do when the string is malformed, as we discussed on today's call (i.e., why does "org.iso.mdoc" return false across user agents... because "." is not allowed)  
2. Return true if use agent allows/knows the identity protocol string.
3. otherwise, return false.

marcoscaceres (changes requested):
This is looking a lot better but it's still missing a lot of details. @mohamedamir and I did a review together and found a few issues that we need to clarify.

index.html
timcappalli: We do have a client. Addressed in #271

index.html
timcappalli: Most of this text was requested by @npdoty

timcappalli: alg added in https://github.com/w3c-fedid/digital-credentials/pull/221/commits/033a6aa1776cf6fcad092d3ab2ddcb24c6ecc582

index.html
marcoscaceres: Right, but that adds a ton of infrastructure that we need to check against implementations. 

We definitely have "clients" in WebKit for DC APIs, but the way #271 is spec'ed doesn't match the implementation (we need to rewrite #271 as it also needs a coordinator for this those clients).

At the same time, clients are infrastructure, not something that necessarily play a role in checking the string, so I think we should drop "client". 

@mohamedamir, as second implementation, would you agree?

index.html
```diff
@@ -464,6 +465,51 @@ <h2>
...
+      clientAllowsProtocol algorithm
+    </h4>
+    <p>
+      When this method is invoked, the user agent MUST execute the following
+      algorithm:
+    </p>
+    <ol class="algorithm">
+      <li>
+        If |protocol| contains characters other than lowercase ASCII letters,
+        digits, or hypens, throw a "{{ConstraintError}}" {{DOMException}}.
```
marcoscaceres: I think this should just return false. If we use the exception, the exception needs to be `InvalidCharacterError`. https://webidl.spec.whatwg.org/#invalidcharactererror 

```suggestion
        If |protocol| is not a [=protocol identifier=], return false.
```

index.html
```diff
@@ -464,6 +465,51 @@ <h2>
...
+      </li>
+      <li>
+        If |protocol| is known and allowed by the user agent, 
+        return <code>true</code>.
```
marcoscaceres: To be "allowed" implies that it is known, so just: 

```suggestion
        Return `true` if the user agent allows |protocol|, `false` otherwise.
```

The return type is boolean, so this is just

index.html
```diff
@@ -464,6 +465,51 @@ <h2>
...
+      </li>
+      <li>
+        Else, return <code>false</code>.
+      </li>
```
```suggestion
```

index.html
marcoscaceres: Right. I'm not saying not to include @npdoty's suggestions - just that they are not in the right place or order.

index.html
mohamedamir: Chrome implementation matches #271 in its current shape. Which parts @marcoscaceres don't match WebKit's implementation?

I am wondering however why do we need to add this to the spec? (why do we need to add distinction between a `client` and a `user agent`?

Do we have precedence in some other spec for this notion of a `client`?

I tried to look into https://www.w3.org/TR/credential-management-1/ and couldn't find a counterpart for the client concept for example.

timcappalli: Spec terminology doesn't always match the code in an implementation. It often describes higher level / logical roles and relationships. 

> Do we have precedence in some other spec for this notion of a client?

Yes, WebAuthn. Most of this text is from WebAuthn and applies in the exact same way. Without a `client` and `client device`, we cannot describe how cross-device flows are invoked with the CLIENT to Authenticator Protocol.

The `client` and `client device` terms need to stay.

mohamedamir: Gotcha!
Thanks Tim for the explanation! (and excuse my ignorance here about the WebAuthn spec)

I think given the text is aligned with what's in the Webauthn spec, and I see your point that it's more higher level, this should be fine to land then!

index.html
```diff
@@ -464,6 +465,51 @@ <h2>
       Integration with Credential Management API
     </h2>
     <aside class="issue" data-number="65"></aside>
+    <h3>
+      The `<dfn>clientAllowsProtocol()</dfn>` method
+    </h3>
+    <p>
+      The {{DigitalCredential/clientAllowsProtocol()}} method allows digital
+      credential verifiers and issuers to determine which 
+      [=digital credential/exchange protocols=] the client supports for
```
```suggestion
      [=digital credential/exchange protocols=] and [=digital credential/issuance protocols=] the client supports for
```

index.html
```diff
@@ -464,6 +465,51 @@ <h2>
...
+      interactions with [=credential providers=].
+    </p>
+    <p class="note">
+      This method does not convey [=digital credential/request protocol=] or
```
```suggestion
      This method does not convey [=digital credential/exchange protocol=] or 
```

index.html
```diff
@@ -464,6 +465,51 @@ <h2>
...
+    <p>
+        User agents SHOULD NOT vary the response value based on any information
+        about availability of hardware, presence or configuration of software,
+        wallets, credential providers, or digital credentials, or user
+        configuration or preferences. If the response value varied, the user
+        agent would introduce risks both of fingerprinting and of silently
+        revealing other details about user behavior or configuration. The
+        response value SHOULD vary only by user agent major version and
+        indicate whether the browser supports distributing requests with that
+        protocol to underlying platform or provider.
```
mohamedamir: Should we stick to referring to client here for example?
```suggestion
        revealing other details about user behavior or configuration. The
        response value SHOULD vary only by client's major version and
        indicate whether the client supports distributing requests with that
        protocol to underlying platform or provider on the client device.
```

index.html
timcappalli: @mohamedamir / @MasterKale what's your take on this? exception or just `false` for strings with invalid characters?

mohamedamir: I don't feel strongly.
I think returning false is simpler, especially, from the developer's point of view, malformed protocols or disallowed protocols are equally actionable.
Therefore, I would vote for simplicity.,

MasterKale: Hmm, is there a compelling reason to restrict protocol names to ascii? Why not allow e.g. some wallet in Japan to support a hypothetical `"認証プロト"` protocol identifier for some domestic use case? I think we should define this API to allow any UTF-8 string to be passed in; put another way I'd rather see `false` than an exception.

marcoscaceres: The API takes a DOMString (UTF-8), so if we need to expand the definition of "protocol identifier", we can do that. However, this seems rather far fetched as it would introduce a whole bunch of issues to deal with (e.g., punctuation in other languages, etc.). 

Protocols are implemented (in WebKit at least) as C++ enumerations - so we have to be careful that whatever we allow won't break the browser's IDL parsers (which translate everything into C++).

Screenshot 2025-06-18 at 10 44 22 AM

Remember that this is for developers, who are used to coding using English constructs (`if, while, for, navigator.credentials.get(), try, catch` etc.).

index.html
marcoscaceres: No, because we haven't defined "client" yet. 

We should do that only _after_ we define what a client. That's a big topic, and something we should do in a separate PR.

index.html
johannhof: FWIW I put a placeholder in #283 to discuss this in the privacy considerations - I agree that we should probably not be too verbose here.

index.html
```diff
@@ -464,6 +465,47 @@ <h2>
...
+      [=digital credential/exchange protocols=] and [=digital credential/issuance protocols=] the client supports for
+      interactions with [=credential providers=].
+    </p>
+    <p class="note">
+      This method does not convey [=digital credential/exchange protocol=] or 
+      [=digital credential/issuance protocol=] support in the [=client device=]
+      or [=credential provider=].
+    </p>
+    <p>
+        User agents SHOULD NOT vary the response value based on any information
```
johannhof: Why is this not a MUST NOT?

marcoscaceres: agree... this should be a MUST NOT

index.html
marcoscaceres: I still don't think we need to mention client here, even it it's to do with clients. The clients are an implementation detail.

index.html
MasterKale: Alright, I won't make this a hill to die on then. Thanks for the explanation.

@timcappalli just to highlight this part of my earlier comment then:

> ...put another way I'd rather see `false` than an exception

I'm not sure if a conclusion was reached on "exception vs false" yet.

index.html
npdoty: I was just volunteering some placeholder text to capture what I thought was the general intent, but I'm happy for it to be refined, or for the implications to be discussed further in the privacy considerations. 

Some normative requirement is important to set expectations about interoperability, to document the consensus that it shouldn't reflect other kinds of availability.

index.html
timcappalli: Yeah, false makes sense.

index.html
timcappalli: changed

index.html
MasterKale: An actor not being part of the name makes for an odd method name:

```
DigitalCredential.allowsProtocol()
```

It'll lead to questions like, "what allows the protocol? How does a 'DigitalCredential' allow it?" But as we all know it's the browser/client/user agent/whatever we want to call it that is ultimately deciding, at the point of invocation of this method, whether a request of the specified protocol will be allowed to make it past the browser. How is the "client" an implementation detail if it's the primary reason this method is being defined, because of the client's active role in facilitating use of DC API?

marcoscaceres: The clients play no part here (hence my objection to the inclusion of the client in the name); only the user agent. Why this method return immediately. I could live with:

* `DigitalCredential.userAgentAllowsProtocol()`

mohamedamir: But according to Tim's definition in #271, the client is part of the user-agent.

>  A [=Digital Credentials API Client=] is an intermediary entity typically implemented in the user agent (in whole, or in part)

and that definition is very aligned with the [WebAuthn Client](https://www.w3.org/TR/webauthn-2/#webauthn-client) 

i.e. 
```mermaid
graph TD
    subgraph Client Device
        subgraph User Agent
            subgraph Client
            end
        end
    end
```

@marcoscaceres Why do you think this is fundamentally different here?

index.html
MasterKale: For sake of progress, a definition of "client" has been proposed in #271 for us to talk about at the next WG meeting.

index.html
```diff
@@ -464,6 +465,48 @@ <h2>
       Integration with Credential Management API
     </h2>
     <aside class="issue" data-number="65"></aside>
+    <h3>
+      The `<dfn>userAgentAllowsProtocol()</dfn>` method
```
```suggestion
      The <dfn>userAgentAllowsProtocol()</dfn> method
```

index.html
```diff
@@ -464,6 +465,48 @@ <h2>
...
+        configuration or preferences. If the response value varied, the user
+        agent would introduce risks both of fingerprinting and of silently
+        revealing other details about user behavior or configuration. The
+        response value SHOULD vary only by user agent major version and
+        indicate whether the browser supports distributing requests with that
+        protocol to underlying platform or provider.
+    </p>
+    <h4>
+      userAgentAllowsProtocol algorithm
+    </h4>
```
```suggestion
```

marcoscaceres (approved):
LGTM

hlflanagan:
Discussed on [DC API Series B call, 25 June 2025](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-06-25-DCAPI-B-notes.md)

PR #238, opened 2025-05-13
Add a11y guidance for user agents

### LINKED ISSUE #229: Horizontal Review - Accessibility
## Description

Work through [this questionnaire](https://w3c.github.io/apa/fast/checklist.html) then [request a review via GitHub](https://github.com/w3c/a11y-request/issues/new/choose) from [APA](https://www.w3.org/WAI/APA/)

MasterKale:
This PR adds informative guidance to user agents encouraging them to mediate presentation and issuance requests in a way that supports a wide range of accessible inputs and outputs.

The intent is to submit this content in the horizontal review request to the [a11y-request repo](https://github.com/w3c/a11y-request).

Partially closes #229

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

### DIFF
--- a/index.html
+++ b/index.html
@@ -963,8 +963,14 @@
       <h2>
         Accessibility Considerations
       </h2>
-      <p class="issue" title="Work in Progress">
-        This section is a work in progress as this document evolves.
+      <p>
+        [=User agents=] SHOULD mediate presentation and issuance requests in a
+        way that offers support for a wide range of user input methods, including
+        input hardware like keyboards, pointing devices, touch screens, voice
+        recognition, etc. Additionally, user agents SHOULD support accessible
+        output methods during these requests to ensure that prompts for
+        [=user consent=] during such requests can be meaningfully communicated to
+        the widest range of end-users.
       </p>
     </section>
     <section id="idl-index"></section>


index.html
```diff
@@ -963,8 +963,14 @@ <h3>
...
       </h2>
-      <p class="issue" title="Work in Progress">
-        This section is a work in progress as this document evolves.
+      <p>
+        [=User agents=] SHOULD mediate presentation and issuance requests in a
+        way that offers support for a wide range of user input methods, including
+        input hardware like keyboards, pointing devices, touch screens, voice
+        recognition, etc. Additionally, user agents SHOULD support accessible
+        output methods during these requests to ensure that prompts for
+        [=user consent=] during such requests can be meaningfully communicated to
```
timcappalli, line 972: We've traditionally said that the browser only handles permission, not consent.

wseltzer:
Discussed 11 June [minutes](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-06-11-DCAPI-B-notes.md)

TallTed:
@mohamedamir --

> "credential chooser" aggregates the credentials from multiple "holders".

I think, rather --

"credential chooser" aggregates one or more credentials from one or more "holders", from which the user can choose which credential to use for a given purpose.

MasterKale:
> Matt: also what is the plan with this PR?

My plan has been to incorporate Marcos' suggestions into the next iteration. I'll update this PR by our next B-series meeting on Wednesday.

index.html
```diff
@@ -86,6 +86,13 @@
           date: "2025-02-25",
           publisher: "W3C"
         },
+        "credman-accessibility": {
+          title: "5.3. Credential Selection",
+          href:" https://w3c.github.io/webappsec-credential-management/#user-mediated-selection",
+          authors: ["Nina Satragno", "Marcos Cáceres"],
+          date: "2024-08-13",
+          publisher: "W3C"
+        }
```
marcoscaceres: Oopsy... there's a different way to do this... 

```suggestion
```

Reminds me I need to add support for citing sections to respec.

index.html
```diff
@@ -1893,8 +1900,16 @@ <h4>
...
       </h2>
-      <p class="issue" title="Work in Progress">
-        This section is a work in progress as this document evolves.
+      <p>
+        Credential presentation and issuance facilitated by the Digital Credentials API have many
+        of the same accessibility considerations as detailed in [[[credman-accessibility]]] of the
+        Credential Management API.
+      </p>
+      <p>
+        [=credential chooser | Credential choosers=] have additional responsibilities. Contents
```
```suggestion
        [=Credential choosers=] have additional responsibilities. Contents
```

index.html
```diff
@@ -1893,8 +1900,16 @@ <h4>
...
+      <p>
+        Credential presentation and issuance facilitated by the Digital Credentials API have many
+        of the same accessibility considerations as detailed in [[[credman-accessibility]]] of the
+        Credential Management API.
+      </p>
+      <p>
+        [=Credential choosers=] have additional responsibilities. Contents
+        displayed to users during these ceremonies, including modal dialogs and QR codes, should be
+        accessible to screen readers. Interactive elements of a ceremony should also be
+        keyboard-accessible whether the user wishes to proceed with the ceremony or abort it.
```
marcoscaceres: Remember, these are "considerations", not so much conformance requirements (well, they are... but these are "things to consider when creating the experience")... so, this should be written as "Things to consider", and then what to expect: 


```suggestion
Implementers need to consider accessibility when designing [=credential choosers=] specifically for this API. Content presented to users in modal dialogs during [=issuance=] or [=presentation=]  — including text, QR codes, and other visual media — SHOULD be accessible to assistive technologies, including screen readers.
</p>
<p>
Interactive elements — including those that allow the user to continue or cancel — MUST be operable in a device-independent manner. This includes support for a range of input modalities, such as keyboard navigation, switch control, and voice input.
</p>
```

This aligns with what I found playing around with Chrome's and Safari's implementations using VoiceOver and tabbing around with the keyboard.

index.html
```diff
@@ -1893,8 +1900,16 @@ <h4>
       <h2>
         Accessibility Considerations
       </h2>
-      <p class="issue" title="Work in Progress">
-        This section is a work in progress as this document evolves.
+      <p>
+        Credential presentation and issuance facilitated by the Digital Credentials API have many
+        of the same accessibility considerations as detailed in [[[credman-accessibility]]] of the
+        Credential Management API.
+      </p>
```
marcoscaceres, line 2273: The problem is that Cred Man doesn't detail any "accessibility considerations" (right now). So if someone goes looking, they would be disappointed.

```suggestion
```

I suggest we remove this, until such time as Cred Man actually gets accessibility considerations.

MasterKale: Makes sense, I've removed this ✌️

index.html
MasterKale: Thanks for digging into this, these changes make sense to me

index.html
```diff
@@ -2056,8 +2056,18 @@ <h4>
       <h2>
         Accessibility Considerations
       </h2>
-      <p class="issue" title="Work in Progress">
-        This section is a work in progress as this document evolves.
+      <p>
+        Implementers need to consider accessibility when designing [=credential choosers=]
```
mohamedamir, line 2267: "credential choosers" are out of scope for this API, this it is the responsibility of the underlying platform to display them.
Not sure if this actually fits here.

mohamedamir:
> @marcoscaceres @mohamedamir @timcappalli The accessibility section has been marked `class="informative"`, so Respec says the section is non-normative:
> Does the inclusion of SHOULD and MUST in the current version of the text mean this class should be dropped accordingly?

You are right, if we are going to stick with SHOULD and MUST we should drop this class IMHO.

index.html
MasterKale: Okay, good point. This should really be about user agents, that might put some UI before invoking the credential chooser, making that UI accessible. Could a simple replacement here work?

marcoscaceres: I'm ok with keeping [=credential choosers=] here... the API still needs to present a credential chooser because Cred Man explicitly requires it as part of the get a credential steps.

index.html
```diff
@@ -2055,15 +2055,25 @@ <h4>
...
+        specifically for this API. Content presented to users in modal dialogs
+        during [=digital credential/issuance request|issuance=]
+        or [=digital credential/presentation requests|presentation=] - including text, QR codes,
+        and other visual media - SHOULD be accessible to assistive technologies,
+        including screen readers.
```
```suggestion
        specifically for this API. The content of modal dialogs
        presented during [=digital credential/issuance request|issuance=]
        or [=digital credential/presentation requests|presentation=], which can include text, QR codes,
        and other visual media, SHOULD be labelled and exposed to assistive technologies.
```

index.html
```diff
@@ -2055,15 +2055,25 @@ <h4>
...
+      </p>
+      <p>
+        Interactive elements - including those that allow the user to continue or cancel
+        - MUST be operable in a device-independent manner. This includes support for a range
+        of input modalities, such as keyboard navigation, switch control, and voice input.
```
marcoscaceres: The second part is overly prescriptive. It reads like, for example, it MUST support voice input, etc. 

```suggestion
        Interactive elements, particularly those that allow the user to continue or abort [=digital credential/issuance request|issuance=] or [=digital credential/presentation requests|presentation=] requests, 
        MUST be operable in a device-independent manner. 
```

marcoscaceres (commented):
I think this is nearly there... just tightened up the text a little bit.

index.html
MasterKale: I dug into the CredMan definition that [=credential choosers=] ends up getting linked to, and it's to a section on User Mediated Selection that talks about credential choosers as something that User Agents implement:

https://www.w3.org/TR/credential-management-1/#user-mediated-selection

> When responding to a call to [get()](https://www.w3.org/TR/credential-management-1/#dom-credentialscontainer-get) on an origin which requires [user mediation](https://www.w3.org/TR/credential-management-1/#user-mediated), user agents MUST ask the user for permission to share credential information. This SHOULD take the form of a **credential chooser** which presents the user with a list of credentials that are available for use on a site, allowing them to select one which should be provided to the website, or to abort the request entirely.

This must be what @marcoscaceres is referring to above (I'm making this connection more for my benefit.) So perhaps it's okay to keep the mention of credential choosers here. @mohamedamir might this change your mind about this particular section of accessibility guidance?

MasterKale:
Thanks for the review @marcoscaceres, I've incorporated those changes. There's only one remaining conversation above, otherwise this PR is ready for @mohamedamir and @timcappalli's review.

index.html
mohamedamir: _(Sorry for being late, fall through the cracks during my vacation)_

I think @marcoscaceres is referring this [step](https://www.w3.org/TR/credential-management-1/#algorithm-request:~:text=Let%20result%20be%20the%20result%20of%20asking%20the%20user%20to%20choose%20a%20Credential%2C%20given%20options%20and%20credentials.) in the algorithm.

But @marcoscaceres didn't we discuss that the algorithm needs to be updated since at its current state, the algorithm doesn't accurately match the DC flow?

I am sorry, but I still think adding Credential Chooser here is confusing and doesn't full represent the user agent role in DC calls.

mohamedamir: I have looked at the algorithm again, and I see how Cred Man actually references a credential chooser (that's different from the credential chooser that Android displays for example). Although effectively, we don't display it in DC, but I am fine keeping it as long as it links to the Cred Man one.

PR #249, opened 2025-05-28
Add guidance on how to handle multiple presentation request entries

### LINKED ISSUE #220: Supporting multiple protocols in a single DC API call
Right now the fact that `digital.requests` is an array communicates that it SHOULD be possible to specify multiple presentation requests. And in the past we've talked about how it would be nicer ergonomics to enable Verifiers to specify a "protocolv2" structure of a request alongside an existing "protocolv1" structure after a major protocol update. But what about the case of the _same_ fundamental request for info defined as two separate protocols?

I think we could service both scenarios if multiple items in `requests` were processed as OR's, as in "protocolv1 OR protocolv2", "protocolv1 OR fooproto."

With the recent news about other presentation protocols coming onto the scene sooner than later, we should probably try to figure this out sooner.

MasterKale:
This PR adds initial text on A) how user agents MUST handle the presence of multiple presentation requests in `options.digital.requests`, and B) how verifiers SHOULD populate `options.digital.requests` to handle getting back a corresponding presentation response.

Closes #220.

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

### DIFF
--- a/index.html
+++ b/index.html
@@ -482,6 +482,28 @@
           Details of how to actually get the [=digital credential=] are
           forthcoming.
         </aside>
+        <!-- Goal: Normative guidance on how multiple requests will only return ONE response with N credentials -->
+        <ol class="algorithm" >
+          <li>
+            <aside class="issue">
+              The following guidance on handling multiple requests will need to be relocated once
+              this entire section of the algorithm is defined.
+            </aside>
+            When multiple [=digital credential/presentation requests=] are present
+            in {{DigitalCredentialRequestOptions/requests}},
+            [=user agents=] MUST only return one
+            [=digital credential/presentation responses | presentation response=] for one
+            of these requests.
+            Multiple credentials MAY be present in this one response depending on the capabilities
+            of the corresponding [=digital credential/exchange protocol=].
+            <aside class="note">
+              Verifiers SHOULD order entries in {{DigitalCredentialRequestOptions/requests}}
+              by descending preference to help [=user agents=] determine
+              which [=digital credential/presentation responses | presentation response=]
+              to return.
+            </aside>
+          </li>
+        </ol>
       </li>
       <li>Return a {{DigitalCredential}}.
       </li>


index.html
```diff
@@ -482,6 +482,27 @@ <h3>
...
+          <li>
+            <aside class="issue">
+              The following guidance on handling multiple requests will need to be relocated once
+              this entire section of the algorithm is defined.
+            </aside>
+            When multiple [=digital credential/presentation requests=] are present
+            in {{DigitalCredentialRequestOptions/requests}},
+            [=user agents=] MUST only return one
+            [=digital credential/presentation responses | presentation response=] for one
+            of these requests.
```
```suggestion
            [=user agents=] MUST return a maximum of one
            [=digital credential/presentation responses | presentation response=] for each
            of these requests.
```

index.html
```diff
@@ -482,6 +482,27 @@ <h3>
...
+            Multiple credentials MAY be present in this one response depending on the capabilities
+            of the corresponding [=digital credential/exchange protocol=].
```
TallTed: This needs further refinement. I'm trying to capture the concept before getting the language just right. There will also need to be some language about what happens when a response that contains credentials A, B, and C is sent after receiving individual requests for credentials A, B, C, D and E, and then a new request for credential B is received. Does the first response cause the new request for B to be ignored? Or might a new response containing B, D, and E be sent? Or something else?
```suggestion
            Multiple credentials MAY be present in one such response depending on the capabilities
            of the corresponding [=digital credential/exchange protocol=].
            Each credential SHOULD be present in only one such response, such that a response
            containing multiple credentials is considered to be a response to all individual
            requests for one of the credentials it contains.
```

index.html
```diff
@@ -497,6 +497,27 @@ <h3>
...
+            <aside class="note">
```
marcoscaceres: Notes shouldn't include RFC2119 keywords (i.e., get rid of the SHOULD). 

I don't think this even makes sense, to be honest. How they are presented is up to the user agent, not up to verifier.

MasterKale: If we imagine a Verifier that is trying to handle a wallet ecosystem that's in between foo-proto-v1 and foo-proto-v2, how might they express a preference for getting a foo-proto-v2 response when a wallet supports it? I'm suggesting here that we approach it by allowing Verifiers to sorting the list of requests by protocol, from most preferred to least preferred.

index.html
MasterKale: This is slightly different behavior than what I'm trying to define here. I'm trying to capture the idea that, while a verifier may specify multiple requests using different protocols for each, the response the verifier gets back will only be for one of those requests.

The suggested wording sounds like it'd open the door to allowing multiple requests to get back one response each. Am I misunderstanding the intent of these proposed changes?

TallTed: Perhaps this?
```suggestion
            [=user agents=] MUST NOT return more than one
            [=digital credential/presentation responses | presentation response=], regardless
            of the number of requests received.
```

index.html
TallTed: Tweaked to remove the RFC2119
```suggestion
            Multiple credentials might be present in one such response depending on the capabilities
            of the corresponding [=digital credential/exchange protocol=].
            Each credential is meant to be present in only one such response, such that a response
            containing multiple credentials is considered to be a response to all individual
            requests for any of the credentials it contains.
```

npdoty:
I think we should indicate to the verifier that the multiple requests (and similarly for the request language) should be semantically compatible/equivalent, or the user/user agent/wallet are easily going to be confused, or might be seen as abuse. Browsers can try to catch really extreme cases that violate that, but it probably wouldn't be easy to test.

MasterKale:
From WG meeting:

- Add a note to Verifiers that requests should be semantically similar
- Review text to see how it might clarify that we're talking about multiple requests within a **single** invocation of DC API

wseltzer:
Discussed [9 July](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-07-09-DCAPI-B-notes.md).

index.html
marcoscaceres: You can't express that preference, because that would allow targeting particular wallets. The way to express it is to just pass the request either exclusively for "-v2" or pass all of them... but it's up to the user to pick the wallet and credential they want. 

For example: I, as a user, have "Super Awesome Wallet" installed (its my favorite app - but only supports V1) and "Crappy V2 Wallet: omg it's sooo bad" (has the V2 credential)... I, as user, don't care that the issuer wants "-v2", because "Super Awesome Wallet" has both my credential and provides such an awesome experience.

index.html
MasterKale: > ...such that a response containing multiple credentials is considered to be a response to all individual requests for any of the credentials it contains.

@TallTed but the way I see it, the presence of credentials in the response (or lack thereof) is a product of the protocol and requested credential format's capabilities. I'm only trying to highlight that the singular response that should come back should be a response structured for whatever protocol was ultimately leveraged for presentation.

I don't want to try and prescribe here how protocols or doc formats downstream should work, just that the user agent should make sure only one response comes through. The main problem I'm trying to solve with this PR is that the VDC ecosystem will be between OID4VP and Annex C for a while - RPs should be able to specify requests for mDLs using both in a single DC API call to support the widest range of browsers. Only one OID4VP **or** Annex C response should be returned.

index.html
MasterKale: Alright, I see why this can't be in here. I'll remove it.

index.html
```diff
@@ -1969,6 +1983,14 @@ <h4>
...
           handling multiple requests and responses for credential presentation.
         </p>
+        <p>
+          Verifiers may wish to define a single credential request
+          using multiple [=digital credential/exchange protocols=],
+          for example to achieve maximum compatibility across [=user agent | user agents=]
+          or when migrating to a newer version of a protocol. When doing so,
+          verifiers SHOULD ensure that all requests are semantically equivalent to avoid confusing
+          the user, the user agent, and the wallet during processing of the request.
+        </p>
```
MasterKale, line 2210: @npdoty I added this text under a different section to try and cover your suggestion that multiple requests should be semantically similar. What do you think?

index.html
MasterKale: @TallTed I tweaked the text a bit into something I think is a midpoint between our positions. What do you think?

TallTed: @MasterKale — There are a lot of variations of these few lines, and I cannot tell which is what, near 1.5 months after my last. Where should I look to see your tweaked text? And which positions do you think this tweaked text is between?

index.html
TallTed: @MasterKale — As I read [the above](https://github.com/w3c-fedid/digital-credentials/pull/249#discussion_r2304998966), it seems that the intent is a lot like HTTP's content negotiation, where a `GET` request includes an `Accept:` header which contains some number of content types, each with a "quality" ranking which is really a "preference" ranking. The HTTP server has its own set of possible content types, each with its own "quality" ranking. The two sets are melded, winding up with a new "quality" ranked list of media types, and the one with the highest "quality" is generally expected to be returned to the client. The client is not expected to issue multiple `GET` requests to brute-force its way to the first one the server will deliver as requested (and not only because the server may deliver an unrequested media type in response to any `GET`, which may mean that the client receives something it doesn't know how to deal with at all, and depending how that brute-force routine was written, the client might stop there with no useful data *or* it might continue down its list of preferred media types until it actually receives what it asked for).

This is approximately the understanding I had after a FedID call some weeks ago.

I think the text should start by making clear the RP is making a single DC API call which somehow conveys information similar to the HTTP `Accept:` header, in this case including all of the RP's protocol, document type, and ranking preferences. Probably, the permutations of protocol and document type pairings should be treated as the media types, and the ranking preferences remain as ranking; the server should respond to the highest ranked pair that it supports (or do the same calculations as with the HTTP `Accept:` header, and return the resulting highest pair, which may or may not be the highest ranked pair in the RP's requeset).

Note that the HTTP specifications do not mandate "one and only one response to any given HTTP request", but there is no reason why a short sentence saying so couldn't be appended to the HTTP specs, probably as an optional thing that could be switched on or off depending on deployment needs.

I submit that this DC API spec should borrow heavily from [the Content Negotiation section](https://datatracker.ietf.org/doc/html/rfc2295) of the HTTP specs, and add that short "one and only one response" sentence to the resulting DC API spec.

MasterKale: > and ranking preferences

Based on [feedback from @marcoscaceres](https://github.com/w3c-fedid/digital-credentials/pull/249#discussion_r2199383904) I don't think it's feasible anymore to try and offer Verifiers the ability to "order by preference." This probably weakens the connection between what I'm proposing here and the prior art you're citing in HTTP's content negotiation mechanisms...

index.html
MasterKale: Oh apologies, this review doesn't appear in the Files changed tab anymore because I made enough changes to everything. I'll resolve this and tag you in a new comment.

index.html
```diff
@@ -771,6 +771,20 @@ <h3>
...
+        <ol class="algorithm" >
+          <li>
+            <aside class="issue">
+              The following guidance on handling multiple requests will need to be relocated once
+              this entire section of the algorithm is defined.
+            </aside>
+            When multiple [=digital credential/presentation requests=] are present
+            in {{DigitalCredentialRequestOptions/requests}},
+            [=user agents=] MUST NOT return more than one
+            [=digital credential/presentation responses | presentation response=].
```
MasterKale: @TallTed Here's the latest version of the sentence that we were [discussing in a previous comment chain](https://github.com/w3c-fedid/digital-credentials/pull/249#discussion_r2195765174).

index.html
MasterKale: In fact I took the sorting-by-preference out completely so I'm going to mark this resolved for now.

index.html
```diff
@@ -771,6 +771,20 @@ <h3>
...
+            Multiple credentials MAY be present in this one response depending on the capabilities
+            of the corresponding [=digital credential/exchange protocol=].
```
TallTed: @MasterKale — Responding to https://github.com/w3c-fedid/digital-credentials/pull/249/files#r2305337096 requires more lines than were connected to that comment...
```suggestion
            When multiple [=digital credential/presentation requests=] are present
            in a {{DigitalCredentialRequestOptions/request}},
            a [=user agent=] MUST NOT return more than one
            [=digital credential/presentation responses | presentation response=].
            However, multiple credentials MAY be included in this single response, depending on the capabilities
            of the corresponding [=digital credential/exchange protocol=].
```

index.html
TallTed: Response to full sentence is in https://github.com/w3c-fedid/digital-credentials/pull/249/files#r2308148399

index.html
```diff
@@ -777,11 +777,12 @@ <h3>
               The following guidance on handling multiple requests will need to be relocated once
               this entire section of the algorithm is defined.
             </aside>
-            When multiple [=digital credential/presentation requests=] are present
-            in {{DigitalCredentialRequestOptions/requests}},
-            [=user agents=] MUST NOT return more than one
-            [=digital credential/presentation responses | presentation response=].
-            Multiple credentials MAY be present in this one response depending on the capabilities
+            When multiple [=digital credential/presentation request | presentation requests=]
+            are present in |requests|,
```
```suggestion
            are present in a |request|,
```

MasterKale: I stuck with `requests` here to try and correlate the guidance to the processing step above that refers to it as `requests`:

Screenshot 2025-08-28 at 4 44 52 PM

index.html
```diff
@@ -780,6 +780,21 @@ <h3>
...
+              The following guidance on handling multiple requests will need to be relocated once
+              this entire section of the algorithm is defined.
+            </aside>
+            When multiple [=digital credential/presentation request | presentation requests=]
+            are present in |requests|,
+            a [=user agent=] MUST NOT return more than one
+            [=digital credential/presentation response=].
+            However, multiple credentials MAY be included in this single response
+            depending on the capabilities
+            of the corresponding [=digital credential/exchange protocol=].
```
```suggestion
            A [=user agent=] MUST NOT return more than one
            [=digital credential/presentation response=] to |requests|,
            even when |requests| contains multiple
            [=digital credential/presentation request | presentation requests=].
            However, multiple credentials MAY be included in this single response
            depending on the capabilities
            of the corresponding [=digital credential/exchange protocol=].
```

index.html
TallTed: Please see [revised suggestion](https://github.com/w3c-fedid/digital-credentials/pull/249/files#r2310836253).

wseltzer:
Discussed 3 September [minutes](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-09-03-DCAPI-B-notes.md#dc-api-prs-and-issues).

index.html
```diff
@@ -786,6 +786,21 @@ <h3>
...
         </aside>
+        <ol class="algorithm" >
+          <li>
+            <aside class="issue">
+              The following guidance on handling multiple requests will need to be relocated once
+              this entire section of the algorithm is defined.
+            </aside>
+            When multiple [=digital credential/presentation request | presentation requests=]
+            are present in |requests|,
+            a [=user agent=] MUST NOT return more than one
```
marcoscaceres: A user agent can't do this regardless, because the WebIDL for a digital credential requires an object (i.e., it can't be a sequence).

index.html
```diff
@@ -786,6 +786,21 @@ <h3>
...
+            [=digital credential/presentation response=].
+            However, multiple credentials MAY be included in this single response
```
marcoscaceres: This doesn't make sense as a normative requirement, as we can't put normative requirements on formats (only on user agents). This might be useful as a non-normative note though.

index.html
```diff
@@ -1983,6 +1998,14 @@ <h4>
           We need to describe concerns, tradeoffs and possible mitigations of
           handling multiple requests and responses for credential presentation.
         </p>
+        <p>
+          Verifiers may wish to define a single credential request
+          using multiple [=digital credential/exchange protocols=],
+          for example to achieve maximum compatibility across [=user agent | user agents=]
```
```suggestion
          for example to achieve maximum compatibility across [=user agents=]
```

index.html
```diff
@@ -1983,6 +1998,14 @@ <h4>
...
+          or when migrating to a newer version of a protocol. When doing so,
+          verifiers SHOULD ensure that all requests are semantically equivalent to avoid confusing
```
marcoscaceres: As above... we can't put conformance requirements on verifiers, but we can suggest best practices in non-normative ways.

TallTed: How about —
```suggestion
          verifiers are expected to ensure that all requests are semantically equivalent to avoid confusing
```

index.html
TallTed: I don't read this as putting requirements on any format (which aren't mentioned here anyway) nor protocol (which are), but rather putting requirements on user agents *subject to* the limitations of the <strike>format</strike> protocol in use, i.e., the user agent MAY include multiple credentials in a single response _if the <strike>format</strike> protocol in use supports doing so_.

index.html
TallTed: Excellent! That should let us simplify the phrasing here substantially. I'll do so [lower, as a revision of a previous suggestion for this block](https://github.com/w3c-fedid/digital-credentials/pull/249/files#r2350117592).

index.html
TallTed: Incorporating all the comments by @marcoscaceres. Too many lines changed in too many ways, so GitHub can't properly highlight the changes. I hope you have a better diff tool at hand.
```suggestion
            A [=user agent=] cannot return more than one
            [=digital credential/presentation response=] to |requests|,
            even when |requests| contains multiple
            [=digital credential/presentation request | presentation requests=],
            because the WebIDL for a digital credential requires an object
            (i.e., it can't be a sequence).
            However, a [=user agent=] MAY include multiple credentials in this
            single response if the [=digital credential/exchange protocol=]
            in use supports doing so.
```

hlflanagan:
Discussed as part of AOB: https://github.com/w3c-fedid/meetings/blob/main/2025/2025-12-02-DCAPI-A-notes.md#aob

index.html
MasterKale: Thanks @TallTed for proposing alternative wording from all of Marcos' comments. I've gone ahead and incorporated this text 🙌

marcoscaceres:
There might be some text we can bring over from here to #420. There’s still something to be said, even if non-normatively, about not expecting multiple credentials (unless the format allows for it).

Issue #247, opened 2025-05-28
Privacy Considerations: Ensuring user choice over wallet involvement (avoid wallet surveillance)


johannhof:
#245 covers information leakage to websites, but we should also consider how information exposure to wallets works

Users may need to perform credential exchanges in sensitive situations where they would expect their user agent to gather their permission before proceeding with a specific wallet. This is to avoid leaking this information to the wallet, and to give the user choice over which wallet to use for a specific purpose.

johannhof:
I believe it's fair to close this since this now has text in the privacy considerations (added in #253)

Issue #243, opened 2025-05-22
Privacy Considerations: Addressing "unnecessary" usage of the API


johannhof:
This seems to be one of the key Privacy topics we need to address for this API - how can DC protect users against illegitimate and/or unnecessary usage of the API. The spec cites an obvious case of "requesting a driver's license to log into a movie rating website", but we will likely see more nuanced examples going forward.

From what I understand, the definition of illegitimate can often be subject to government regulation. Obviously, this is something that DC implementers should defer to and provide the capabilities to support, and we could bolster the spec to more clearly define this requirement.

Where legitimacy is not clearly regulated, things get more interesting - to what degree can and should a user agent enforce its own protections?

It looks like this has been discussed several times as part of the incubation process, e.g. in #35, #44 and #136. I don't think we can come to an entirely satisfying conclusion in the short term, but we should make an effort to reflect the state of the discussion and make sensible recommendations and requirements that we can agree on.

CC @npdoty @martinthomson @samuelgoto @bvandersloot-mozilla

johannhof:
Another challenge here: Some regulatory environments may decide to filter / authorize verifiers at the Wallet level - this seems like a requirement that would be very difficult to reconcile at the browser level.

zoracon:
Linking a similar issue I raised: https://github.com/w3c-fedid/digital-credentials/issues/317#issuecomment-3110649454

I was told none of this is able to get solved by the DC API. But what I am trying to get to the bottom of is what _can_. So instead of opening more issues that will get closed, maybe here is a place to discuss what is possible for the DC API to facilitate when it comes to verifier abuse and unnecessary requests.

kdenhartog:
Following on from @zoracon question of "what can" be done, I'd like to also add are we placing restrictions on protocols or features that allow for the circumvention of unnecessary usage of the API? For example, are browsers expected to restrict access to protocols which are too permissive in their willingness to share credentials?

marcoscaceres:
This issue asked for guidance on what constitutes "unnecessary" credential requests and how user agents should protect against them.

The spec now has a full **"Unnecessary Requests for Credentials"** section:

> "Unnecessary credential requests are a key privacy risk to the entire digital credentials ecosystem."

This section covers intentional abuse, unnecessary requests without malicious intent, excessive data collection, and includes dedicated sub-sections for both government-issued and non-government-issued credentials with specific mitigations.

[...]

Issue #226, opened 2025-04-25
privacy review for registry inclusion


npdoty:
Requirements we should clarify for conducting privacy review for protocols to be included in a registry.
 

- [x] review would be done by the [Privacy Working Group](https://www.w3.org/groups/wg/privacy/), not the Privacy Interest Group (no longer exists).
- [ ] privacy review needs to be completed, and issues need to be meaningfully addressed, in concordance with the recommendations of the threat model/privacy considerations document (conducting a privacy review and ignoring all the raised issues isn't the point)

I would also add something about process, although I'm not sure this would have as broad consensus in the group at the moment.
- protocols listed in the registry should be developed through an open process allowing impacted stakeholders to participate, review and provide feedback on the potential implications for privacy and other human rights

(The registry inclusion update PR got merged before I could list out these issues.)

wseltzer:
Discussed 11 June [minutes](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-06-11-DCAPI-B-notes.md)

npdoty:
(Just being consistent on using `privacy-tracker` for a privacy issue that we are tracking and could be of relevance to external privacy review stage. `-needs-resolution` is for external reviewers indicating a particular level of importance; those kinds of decisions can come later and based on external review.)

hlflanagan:
Discussed [2026-02-18 FedID WG call](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-02-18-DCAPI-B-notes.md#privacy-review-for-registry-inclusion-226). Further discussion needed.

hlflanagan:
Discussed on [1 April 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-04-01-DCAPI-B-notes.md#privacy-review-for-registry-inclusion-226) call. We need more people focused on this (and related) issues.

PR #251, opened 2025-05-30
Privacy Considerations: Add initial Introduction, Spectrum of Privacy sections

johannhof:
This is some initial text for the introductory part.

@npdoty @simoneonofri I'd love to get your review. In the interest of fast progress, I'd like to make sure we can merge these PRs iteratively so if you think there are things that should be mentioned but are not in this PR, let's add those in follow-ups. This is still meant to be WIP, but let's not pile up a huge PR.

### DIFF
--- a/index.html
+++ b/index.html
@@ -47,7 +47,22 @@
       edDraftURI: "https://w3c-fedid.github.io/digital-credentials/",
       group: "fedid",
       github: "https://github.com/w3c-fedid/digital-credentials",
-      localBiblio: {},
+      localBiblio: {
+        "credential-considerations": {
+          title: "User considerations for credentials on the Web",
+          href: "https://github.com/w3c/credential-considerations/blob/main/credentials-considerations.md",
+          authors: ["Nick Doty", "Rick Byers"],
+          date: "2025-03-26",
+          publisher: "W3C"
+        },
+        "threat-model-decentralized-identities": {
+          title: "Threat Model for Decentralized Identities",
+          href: "https://github.com/w3c-cg/threat-modeling/blob/main/models/decentralized-identities.md",
+          authors: ["Simone Onofri"],
+          date: "2025-05-28",
+          publisher: "W3C"
+        },
+      },
       xref: {
         profile: "web-platform",
       },
@@ -857,31 +872,78 @@
         <p>
           This section is a work in progress as this document evolves.
         </p>
-        <p>
-          The documents listed below outline various privacy considerations for
-          Digital Credentials, both broadly and for presentation on the web.
-          Their contents will be integrated into this document gradually.
-        </p>
-        <ul>
-          <li>
-            <a href=
-            "https://github.com/WICG/digital-credentials/blob/main/horizontal-reviews/security-privacy.md">
-            TAG Security and Privacy Considerations Questionnaire (WIP)</a>
-          </li>
-          <li>
-            <a href="https://github.com/w3cping/credential-considerations">W3C
-            Privacy Interest Group (PING) Credentials Considerations
-            repository</a> which contains in-progress list of concerns and
-            implications of identity-verification and credential presentation,
-            for consideration in Internet and Web standardization
-          </li>
-          <li>
-            <a href=
-            "https://github.com/w3c-cg/threat-modeling/blob/main/models/decentralized-identities.md">
-            Threat Model for Decentralized Identities</a>
-          </li>
-        </ul>
       </div>
+      <p>
+      The Digital Credentials API integrates into a complex ecosystem with multiple technology
+      layers and various participants (including but not limited to Verifiers, Holders and Issuers),
+      each of which have to consider different aspects of user privacy.
+
+      This specification does not attempt to exhaustively list all considerations
+      for the different participants. We would like to refer these parties to a
+      variety of other resources that explore the digital credentials threat
+      model more holistically:
+      </p>
+      <ul>
+        <li>
+          <a href="">W3C Privacy Working Group Credentials Considerations</a> [[credential-considerations]]
+        </li>
+        <li>
+          <a href="">
+          Threat Model for Decentralized Identities</a> [[threat-model-decentralized-identities]]
+        </li>
+        <li>
+          <a data-cite="vc-data-model#privacy-considerations">
+          VC Data Model Privacy Considerations</a>
+        </li>
+        <li>
+          ...
+        </li>
+      </ul>
+      <p>
+        Instead, this specification aims to describe where the Digital Credentials
+        API itself is responsible for meeting user expectations of Privacy, and
+        how developers and user agents implementing or integrating with the API
+        are to consider these responsibilities.
+      </p>
+      <p>
+        The privacy considerations for digital credentials are not static. They
+        will evolve over time as the ecosystem matures, and may be informed by
+        the behavior of other actors in the ecosystem, improvements in other
+        layers of the stack, new threats to user privacy, as well as changing
+        societal norms and regulations.
+      </p>
+      <p>
+        It is expected that the various groups involved in the design and implementation
+        of the Digital Credentials API actively monitor the evolving privacy
+        landscape and participate in the corresponding evolution of the API.
+      </p>
+      <section data-cite="vc-data-model#spectrum-of-privacy">
+        <h3>
+          Spectrum of Privacy
+        </h3>
+        <p>
+          The Digital Credentials API serves a variety of use cases with different
+          grades of data disclosure and individual users with different preferences
+          depending on the context that they are in. Notably, the privacy properties
+          of a credential exchange mediated by this API could be mandated by
+          the legal and regulatory environment of an individual user.
+        </p>
+        <p>
+          This means that some users may not want, or be allowed,
+          to use the most privacy-preserving means of exchanging credential information.
+          Nonetheless, user agents need to serve users with an experience
+          that is private by default and protect them from harm.
+        </p>
+        <p>
+          Because of this spectrum of preferences and use cases, it may be difficult for a user
+          agent to discern whether a user means to expose their personal information
+          or is being tricked into doing so. It is thus the user agent's responsibility
+          to ensure that every user understands what data they are sharing and who
+          will participate in the exchange of information, before the exchange
+          begins. It is further the user agent's responsibility to ensure that
+          all parties have a legitimate reason for facilitating the exchange.
+        </p>
+      </section>
       <section>
         <h3>
           Unnecessary Requests for Credentials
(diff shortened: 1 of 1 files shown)


timcappalli (approved):
LGTM! Thanks Johann!

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
-          Explain how selective disclosure and unlinkable disclosure help
-          preserve privacy as well as their limitations in doing so.
+        <p>
+          Because of this spectrum of preferences and use cases, it may be difficult for a user
+          agent to discern whether a user means to expose their personal information
+          or is being tricked into doing so. It is thus the user agent's responsibility
+          to ensure that every user understands what data they are sharing and who
+          will participate in the exchange of information, before the exchange
+          begins. It is further the user agent's responsibility to ensure that
+          all parties have a legitimate reason for facilitating the exchange.
```
mohamedamir: It seems to me like this is asking for too much from the user agent. IMHO, the user agent's responsibility is to make sure the user is aware of what data is being exchanged, but how would the user agent ensure there is a legitimate for the exchange?

Similar to the camera API for example, the user agent makes sure the camera stream isn't share without clear user permission, but it doesn't ensure if there is a legitimate reason  for sharing it.

johannhof: Hm, yeah, I think this is a fair perspective, although I think this is a bit of a special consideration for DC given the concern raised about abuse, overuse and exclusion (see #243). I do believe that there are different opinions on this particular point.

In any case, it's probably something that should be discussed in #243 and related PRs / issues, so I'm happy to remove the text for now just for the sake of getting the introduction in, and we can continue the conversation and potentially add it back in later.

mohamedamir: Fair enough, happy to either remove in this PR, or in a follow-up once when there is consensus!
I just usually prefer to compare this to Camera API, where we have comparable privacy bar IMHO!

mohamedamir (approved):

npdoty:
Does this PR intend to delete all the existing placeholder sections without replacing them? The substantive text adds more to the Introduction and a Spectrum of Privacy section (per the VCDM spec), which I understand, but doesn't explain why it removes most of the placeholder sections.

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
+          <a data-cite="vc-data-model#privacy-considerations">
+          VC Data Model Privacy Considerations</a>
+        </li>
+        <li>
+          ...
+        </li>
+      </ul>
+      <p>
+        Instead, this specification aims to describe where the Digital Credentials
+        API itself is responsible for meeting user expectations of Privacy, and
```
npdoty: I think the goal is not just to describe how the API might meet expectations of privacy, but to describe how user agent implementers satisfy the [user agent duties](https://w3ctag.github.io/privacy-principles/#dfn-user-agent-duties) and make the web a trustworthy platform. The goal is protecting user privacy, not just meeting expectations (which are hard to determine, and may not satisfy those protections in cases where users no longer expect to have any privacy).

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
+        how developers and user agents implementing or integrating with the API
+        are to consider these responsibilities.
+      </p>
```
npdoty, line 908: It would be useful to define responsibilities of different parties. But I don't think the browser API design can just disregard the privacy impacts when we hope other parties will do it but know they won't.

index.html
npdoty: I'm not sure this is part of the Spectrum of Privacy concept (as VCDM defines it). But I do think that part of addressing the risks of abuse of government-issued ID on the web is helping to ensure the use is appropriate, not just transparency of what information will be shared in a particular transaction.

index.html
simoneonofri: In general, I agree with keeping the scope of the Digital Credentials API within what the API can do for the benefit of the user. There will be several threats inherited on the one hand from the threat model for Digital Credentials in general (already indicated), and others (at least from what we are seeing in the Security Considerations) from the [Threat Model for the Web](https://github.com/simoneonofri/threat-model-web) - which I will move to the W3C organization as soon as possible.

index.html
simoneonofri: @npdoty Which parts exactly?

index.html
simoneonofri: The spectrum is useful for assessment and disclosure of the type of data being released. In the meantime, I would include it to establish a common vocabulary.

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
+          Explain how the API could be used to unnecessarily request digital
+          credentials from individuals such as requesting a driver's license to
+          log into a movie rating website and how the ecosystem can mitigate
+          this risk.
         </p>
       </section>
       <section>
         <h3>
-          Transmission of Personally Identifiable Information
+          User Permission and Transparency
```
simoneonofri: I agree to keep the placeholders for the threats. Once we have the content, we can give it a structure.

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
-        <p class="issue" title="Work in progress">
-          Explain how the API could be used to unnecessarily request digital
-          credentials from individuals such as requesting a driver's license to
-          log into a movie rating website and how the ecosystem can mitigate
-          this risk.
-        </p>
-      </section>
-      <section>
-        <h3>
-          Over Collection of Data
```
npdoty: Data minimization designs for Web APIs are traditional protections to consider. It would be an important quality for considering the protocols in the registry, but would also be important for the API design, whether it's easier for sites to ask for everything, or more minimal specific data elements.

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
-        </h3>
-        <p class="issue" title="Work in progress">
-          Explain how the API could be used to request more data than necessary
-          for a transaction and how the ecosystem can mitigate that over
-          collection.
-        </p>
-      </section>
-      <section>
-        <h3>
-          Individual Consent
```
npdoty: Consent seems important and germane. There isn't consensus in the group yet on whether consent is gathered by the site, the browser or the wallet, and that desperately needs to be documented. Where consent will be obtained determines how information flows between the parties. (If consent is handled by the wallet, then all the context inforemation needs to be passed along to the wallet, for example.)

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
+        societal norms and regulations.
+      </p>
+      <p>
+        It is expected that the various groups involved in the design and implementation
+        of the Digital Credentials API actively monitor the evolving privacy
+        landscape and participate in the corresponding evolution of the API.
+      </p>
+      <section data-cite="vc-data-model#spectrum-of-privacy">
         <h3>
-          Data Retention
```
npdoty: Retention remains relevant. Especially significant for evaluation of the privacy characteristics of the request protocols, as those may include intent-to-retain indications.

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
+          or is being tricked into doing so. It is thus the user agent's responsibility
+          to ensure that every user understands what data they are sharing and who
+          will participate in the exchange of information, before the exchange
+          begins. It is further the user agent's responsibility to ensure that
+          all parties have a legitimate reason for facilitating the exchange.
         </p>
       </section>
       <section>
         <h3>
-          Phoning Home
```
npdoty: This is a [critical privacy property of digital ID](https://nophonehome.com/). It's relevant for the web API in terms of how/when wallets are notified about a request, evaluation of the request and issuance protocols to see whether presentation will enable phone-home, and whether revocation properties are included in such a way that might lead to phone home.

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
+        <p>
+          This means that some users may not want, or be allowed,
+          to use the most privacy-preserving means of exchanging credential information.
+          Nonetheless, user agents need to serve users with an experience
+          that is private by default and protect them from harm.
         </p>
-      </section>
-      <section>
-        <h3>
-          Selective and Unlinkable Disclosure
```
npdoty: Selective disclosure and unlinkable presentation are key privacy properties for any digital ID system. I would expect the privacy considerations section to detail the requirements for request protocols to enable selective disclosure to take place, and a way for linkable and unlinkable presentations to be distinguished in order to help the user identify tracking capabilities that they might not otherwise realize.

index.html
```diff
@@ -857,111 +872,94 @@ <h2>
...
+          The Digital Credentials API serves a variety of use cases with different
+          grades of data disclosure and individual users with different preferences
+          depending on the context that they are in. Notably, the privacy properties
+          of a credential exchange mediated by this API could be mandated by
+          the legal and regulatory environment of an individual user.
         </p>
-      </section>
-      <section>
-        <h3>
-          Compliance with Privacy Regulations
```
npdoty: I anticipate that we will want to document how sites complying with eIDAS, GDPR or other regulations will be able to use this API, and how. I don't feel strongly whether it's in this document or a separate legal/implementation explainer sort of document.

index.html
johannhof: Thanks Nick, I think that's a much better way to say it. I'll update the PR.

index.html
johannhof: > But I don't think the browser API design can just disregard the privacy impacts when we hope other parties will do it but know they won't.

That's a bit of a cynical way to look at it - I still think we should frame concerns from the perspective of this API. I agree that when there is reasonable doubt / concern whether another party will live up to our expectations of user privacy then that's important to discuss. However, and maybe this is more of an editorial point, I don't believe that this should usually be in the form of a separate section in the Privacy considerations. I'd like to structure those around the main themes that the DC API will have to care about, and then frame the responsibilities of the DC API in those areas based on what it may or may not expect from others.

index.html
johannhof: Sure - I can add them back in. Just from my own editorial perspective, I'd really prefer tracking these as issues instead of (or at least in addition to) inline placeholders. Otherwise we lose the ability to discuss them with the group.

@simoneonofri @npdoty - it would be great if you could file issues with the `privacy-considerations` tag if you particularly care about keeping any of the current placeholders as top-level topics.

index.html
johannhof: Okay, but I don't think it's worth retaining (har har) this as a separate section, and instead we should probably figure out if we want to include it in a section on evaluating protocols, which I hope we can track in #226

index.html
johannhof: That is totally fair and correct, but it seems to me that this is a case for #247 then, and doesn't need it's own separate section?

index.html
johannhof: We should consider the protocol evaluation bit, for sure, in its own section, but I'm not sure what we would say about the API design piece. It doesn't feel like something that applies to the DC API.

index.html
johannhof: I added some consent vs. permission text to https://github.com/w3c-fedid/digital-credentials/pull/253

I believe that the vast majority of this group, and implementers, don't see the DC API in a position where it can enforce legal consent, and the relevant information should be handed down by the protocol. I elaborated a bit more on that in the PR, but feel free to suggest changes / additions.

Just in order to avoid blocking this PR on all those removal discussions I'm going to restore these headers but this is another one where I'd push back on a separate top-level section.

index.html
johannhof: Yeah, I agree that we should encourage user agents to take advantage of selective / unlinkable disclosures to frame their UI in a way that shows users the value of these properties and clearly communicates what will *not* be shared as part of an exchange that makes use of them.

johannhof:
I addressed the actionable feedback that I got on this. Given editor approval, I'm going to merge this in the interest of rapid iteration - again, leaving the WIP sign up and happy to revisit any part of what's been written here. I believe we'll get to the more in-depth discussion in other sections anyway. :)

wseltzer:
Discussed 2 June [minutes](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-06-02-DCAPI-A-notes.md)

index.html
npdoty: As I noted, phoning home is not just a risk from wallets, and user choice over wallets (which is extremely unlikely) doesn't seem like a good solution anyway. I would rather not blame the user for not having done extensive due diligence over their wallet, even if they actually had free choice over it, which we know they will not.

index.html
npdoty: It's come up several times in the Digital Credentials API already. For example, should a site be able to request multiple credentials in a single request, multiple versions of a credential, multiple credentials from multiple wallets? Should the site indicate which parts of the request are optional and which are mandatory? Should we design the API to have simpler paths to request certain individual fields, rather than entire documents?

johannhof: I filed #286 to track this

Issue #136, opened 2024-06-26
Describe trust signals in protocol registry


npdoty:
[At the last meeting](https://github.com/WICG/digital-credentials/wiki/2024-06-21-Hybrid-Meeting-Notes#request-types-credential-types-registries-etc), there was recognition of a requirement that may apply to EUDI cases of the wallet needing to confirm the registration/approval of the verifier and a desire to also enable other kinds of trustmarks, attestations or verifications from third-parties that could help a user/wallet know whether to provide that information. (This also came up as a recommended approach in #59.)

This could potentially be done: 
1) at the protocol layer, with an additional profile of how to communicate this information in client metadata (OAuth client metadata, in this case, metadata communicated about the server/verifier), 
2) as well-known information about an origin,
3) or even as an API parameter.

Requirements may include indicating what data will be requested, for what purposes and for how long, who has verified the basis of the request, and then some proof of verification or attestation (a signature, basically) from that third-party (whether a local government data protection authority or some independent organization doing reviews).

There could be other information about the verifier that the verifier should, wants to or needs to communicate, including declared privacy policy information, an endpoint to access or request deletion of credential data previously collected, or who to contact regarding complaints or abuse of the system. As noted, there might be some similarities to past attempts at machine-readable privacy policy documents, but this would be specific to a particular sensitive use case, facilitate compliance with regulations, and be deployed in contexts with more regulator involvement.

jogu:
As far as I know these things are all defined at the OID4VP level already, and work fine with the draft of the OID4VP browser API profile.

I am not sure there is anything for the browser API group to do.

npdoty:
I did re-read the OpenID4VP Editor's Draft and RFC 7591 again before opening this issue to confirm that while it might be possible to use those to communicate some of this information, it didn't seem to be profiled or standardized. I've looked now at the PR for a browser profile, and while that describes some subset of these uses, I still don't see the detail on how that information would actually be communicated or examples of those attestations or trustmarks. If you have pointers on something else I should look at, or if I should open issues for clarification elsewhere, let me know.

jogu:
[...]

PR #253, opened 2025-06-02
Add initial text for permission & transparency

johannhof:
Something to get us started - includes a potentially somewhat controversial claim about Consent vs. Permission for @npdoty to discuss with the group.

@simoneonofri would love to get your review here as well

***
<a href="https://pr-preview.s3.amazonaws.com/johannhof/digital-credentials/pull/253.html" title="Last updated on Jun 10, 2025, 4:33 AM UTC (9528a30)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/253/5ce4855...johannhof:9528a30.html" title="Last updated on Jun 10, 2025, 4:33 AM UTC (9528a30)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1017,14 +1017,119 @@
       </section>
       <section>
         <h3>
-          Transmission of Personally Identifiable Information
+          User Permission and Transparency
         </h3>
-        <p class="issue" title="Work in progress">
-          Explain that the API does enable the transmission of personally
-          identifiable information and that it does its best to ensure there is
-          informed consent by the individual, but that the consent might be
-          provided due to exhaustion or not understanding what PII is being
-          transmitted and how to mitigate those concerns.
+        <p class="issue" title="Work in progress"></p>
+        <p>
+          The Digital Credentials API enables the capability to share highly
+          personal, sensitive and at-risk user information with websites via
+          credentials, potentially granting the ability to track users online
+          and offline, through permanent unique irrevocable cross-context
+          identifiers. It also reveals parts of the user's browsing activity as
+          well as their intent to identify to specific websites to wallets. A
+          crucial responsibility of the user agent in a credential request is
+          to gather permission by the user to proceed with the exchange of
+          information.
+        </p>
+        <p>
+          Important context that is needed for a user to make a decision about
+          proceeding with a credential exchange includes:
+        </p>
+        <ul>
+          <li>The origin of the verifier that requests the credential.
+          </li>
+          <li>The information that is being requested, or would be revealed by
+          responding to the request.
+          </li>
+          <li>Why the information is being requested.
+          </li>
+          <li>What the information will be used for.
+          </li>
+          <li>Whether presenting this information will enable tracking.
+          </li>
+          <li>How the information will be shared or retained.
+          </li>
+          <li>Which wallet(s) can be used to fulfill the credential request.
+          </li>
+          <li>Which credential would be used to share the requested
+          information.
+          </li>
+          <li>Any evaluations and attestations of this information, if
+          available.
+          </li>
+        </ul>
+        <p>
+          It is advised that user agents in their implementation ensure that
+          this information is fully disclosed to the user before an exchange of
+          any user-related information occurs.
+        </p>
+        <p class="issue" data-number="252">
+          Should these be normative in the spec?
+        </p>
+        <p class="issue" data-number="44">
+          Should the API be designed so the site can provide in-context
+          explanations?
+        </p>
+        <h4>
+          Integrating Multiple User Agents
+        </h4>
+        <p>
+          Depending on the technical architecture of the user's system, it is
+          likely that the definition of a "user agent" includes multiple
+          cooperating layers of the software stack, such as a browser and the
+          operating system. The greatest priority for these layers has to be a
+          safe and well-informed user permission experience. As such,
+          integration can be vital for user safety. Some layers may hold
+          information that is inaccessible to others, such as the availability
+          of a user's credentials. Overprompting or prompting without
+          sufficient context could lead to (exploitable) confusion and prompt
+          blindness.
+        </p>
+        <p>
+          For this reason user agents are encouraged to integrate software
+          layers for an ideal user experience when prompting for permission, if
+          they consider it safe to do so. This could happen, for example, if a
+          browser trusts the API contract of an operating system to show an
+          appropriate prompt and thus does not show a prompt itself.
+        </p>
+        <h4>
+          Permission Prior to Wallet Selection
+        </h4>
+        <p>
+          As part of the user permission flow, the user agent needs to ensure
+          that users retain the choice over whether to forward a credential
+          request to a wallet, and which wallet to select. This is due to the
+          information disclosure that happens as part of the request, and the
+          ability for wallets to retain or share this information at time of
+          the request.
+        </p>
+        <h4>
+          Permission vs. Consent
+        </h4>
+        <p>
+          The permission mediated by the user agent is not consent, which can
+          vary among different legal and regulatory environments and may need
+          to be collected by the digital wallet before sharing information with
+          the verifier, or by the verifier itself before initiating the
+          request. With frameworks and regulations for obtaining consent still
+          being developed, the API aims to enable the exchange of the necessary
+          information, which could include:
+        </p>
+        <ul>
+          <li>The privacy policy of the verifier receiving the credential.
+          </li>
+          <li>The purpose of the verifier for requesting the information.
+          </li>
+          <li>Assertions of the verifier's legimitacy and registration for
+          accessing the credential, such as <a href=
+          "https://github.com/eu-digital-identity-wallet/eudi-doc-architecture-and-reference-framework/blob/main/docs/annexes/annex-2/annex-2-high-level-requirements.md#a2327-topic-27---registration-of-pid-providers-providers-of-qeaas-pub-eaas-and-non-qualified-eaas-and-relying-parties">
+            EUDI access and registration certificates</a>.
+          </li>
+        </ul>
+        <p>
+          As more of this information becomes available in a structured format,
+          we expect user agents and this specification to leverage it to
+          improve the user permission experience as well.
         </p>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -1040,21 +1040,12 @@
           <li>The information that is being requested, or that would be
           revealed by responding to the request.
           </li>
-          <li>Why the information is being requested.
-          </li>
-          <li>What the information will be used for.
-          </li>
           <li>Whether presenting this information will enable tracking.
-          </li>
-          <li>How the information will be shared or retained.
           </li>
           <li>Which wallet(s) can be used to fulfill the credential request.
           </li>
           <li>Which credential would be used to share the requested
           information.
-          </li>
-          <li>Any evaluations and attestations of this information, if
-          available.
           </li>
         </ul>
         <p>
@@ -1119,6 +1110,13 @@
           <li>The privacy policy of the verifier receiving the credential.
           </li>
           <li>The purpose for which the verifier is requesting the information.
+          </li>
+          <li>What the information will be used for.
+          </li>
+          <li>How the information will be shared or retained.
+          </li>
+          <li>Any evaluations and attestations of this information, if
+          available.
           </li>
           <li>Assertions of the verifier's legimitacy and registration for
           accessing the credential, such as <a href=
--- a/index.html
+++ b/index.html
@@ -1021,25 +1021,24 @@
         </h3>
         <p class="issue" title="Work in progress"></p>
         <p>
-          The Digital Credentials API enables the capability to share highly
-          personal, sensitive and at-risk user information with websites via
-          credentials, potentially granting the ability to track users online
-          and offline, through permanent unique irrevocable cross-context
-          identifiers. It also reveals parts of the user's browsing activity as
-          well as their intent to identify to specific websites to wallets. A
-          crucial responsibility of the user agent in a credential request is
-          to gather permission by the user to proceed with the exchange of
-          information.
+          The Digital Credentials API enables the sharing of highly personal,
+          sensitive, and at-risk user information with websites via credentials,
+          potentially granting the ability to track users online and offline,
+          through permanent, unique, irrevocable, cross-context identifiers. It
+          also reveals parts of the user's browsing activity as well as their
+          intent to identify to specific websites and/or wallets. One crucial
+          responsibility of the user agent in a credential request is to gather
+          permission from the user to proceed with the exchange of information.
         </p>
         <p>
-          Important context that is needed for a user to make a decision about
-          proceeding with a credential exchange includes:
+          Important context details that are needed for a user to make an informed
+          decision about proceeding with a credential exchange include the following:
         </p>
         <ul>
           <li>The origin of the verifier that requests the credential.
           </li>
-          <li>The information that is being requested, or would be revealed by
-          responding to the request.
+          <li>The information that is being requested, or that would be
+          revealed by responding to the request.
           </li>
           <li>Why the information is being requested.
           </li>
@@ -1060,8 +1059,8 @@
         </ul>
         <p>
           It is advised that user agents in their implementation ensure that
-          this information is fully disclosed to the user before an exchange of
-          any user-related information occurs.
+          the details listed are fully disclosed to the user before an exchange
+          of any user-related information occurs.
         </p>
         <p class="issue" data-number="252">
           Should these be normative in the spec?
@@ -1074,51 +1073,52 @@
           Integrating Multiple User Agents
         </h4>
         <p>
-          Depending on the technical architecture of the user's system, it is
-          likely that the definition of a "user agent" includes multiple
+          Depending on the technical architecture of a user's system, it is
+          likely that the definition of a "user agent" will include multiple
           cooperating layers of the software stack, such as a browser and the
-          operating system. The greatest priority for these layers has to be a
-          safe and well-informed user permission experience. As such,
+          operating system. The greatest priority for these layers has to be
+          a safe and well-informed user permission experience. As such,
           integration can be vital for user safety. Some layers may hold
-          information that is inaccessible to others, such as the availability
-          of a user's credentials. Overprompting or prompting without
-          sufficient context could lead to (exploitable) confusion and prompt
-          blindness.
+          information that is inaccessible by other layers, such as the
+          availability of a user's credentials. Overprompting or prompting
+          without sufficient context could lead to (exploitable) confusion
+          and prompt blindness.
         </p>
         <p>
-          For this reason user agents are encouraged to integrate software
-          layers for an ideal user experience when prompting for permission, if
-          they consider it safe to do so. This could happen, for example, if a
+          For this reason, user agents prompting for permission are encouraged
+          to integrate software layers for an ideal user experience, if they
+          consider it safe to do so. This could happen, for example, if a
           browser trusts the API contract of an operating system to show an
-          appropriate prompt and thus does not show a prompt itself.
+          appropriate prompt, and thus does not show a prompt itself.
         </p>
         <h4>
           Permission Prior to Wallet Selection
         </h4>
         <p>
           As part of the user permission flow, the user agent needs to ensure
-          that users retain the choice over whether to forward a credential
+          that users retain the power to choose whether to forward a credential
           request to a wallet, and which wallet to select. This is due to the
           information disclosure that happens as part of the request, and the
-          ability for wallets to retain or share this information at time of
+          ability of wallets to retain or share this information at the time of
           the request.
         </p>
         <h4>
           Permission vs. Consent
         </h4>
         <p>
-          The permission mediated by the user agent is not consent, which can
-          vary among different legal and regulatory environments and may need
-          to be collected by the digital wallet before sharing information with
-          the verifier, or by the verifier itself before initiating the
-          request. With frameworks and regulations for obtaining consent still
-          being developed, the API aims to enable the exchange of the necessary
-          information, which could include:
+          The permission mediated by the user agent is not consent, which has
+          specific legal definitions which can vary among different legal and
+          regulatory environments and may need to be collected by the digital
+          wallet before sharing information with the verifier, or by the
+          verifier itself before initiating the request. With frameworks and
+          regulations for obtaining consent still being developed, this API
+          aims to enable the exchange of the necessary information, which
+          could include the following:
         </p>
         <ul>
           <li>The privacy policy of the verifier receiving the credential.
           </li>
-          <li>The purpose of the verifier for requesting the information.
+          <li>The purpose for which the verifier is requesting the information.
           </li>
           <li>Assertions of the verifier's legimitacy and registration for
           accessing the credential, such as <a href=


index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+        </h4>
+        <p>
+          Depending on the technical architecture of the user's system, it is
+          likely that the definition of a "user agent" includes multiple cooperating
+          layers of the software stack, such as a browser and the operating system,
+          and potentially the wallet application. The greatest priority for these
+          layers has to be a safe and well-informed user permission experience. As
+          such, integration can be vital for user safety. Some layers may hold
+          information that is inaccessible to others, such as the availability of a
+          user's credentials. Overprompting or prompting without sufficient context
```
c2bo: I think this clarification is really important - at the end of the day it shouldn't be important which layer fulfils which part. In the overall flow we must make sure to provide the user with the necessary information to make an informed decision on what data they are sharing with whom for what purpose.

In the beginning of the document, the term "user agent" is introduced with a reference to the whatwg. Not sure about this, but would it make sense to introduce "user agent" with this meaning earlier on or reference this section to avoid confusion?

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          sensitive and at-risk user information with websites via credentials, and
+          reveals parts of the user's browsing activity as well as their intent to
+          authenticate with specific websites to wallets.
+          A crucial responsibility of the user agent in a credential request
+          is to gather permission by the user to proceed with the exchange of
+          information.
+        </p>
+
+        <p>
+          The most critical pieces of context that are needed for a user to make a decision
```
npdoty: Critical context certainly also includes: 

- why information is being requested, 
- what it's being used for, 
- how the data will be shared or retained, 
- whether presentation will enable tracking,
- evaluations and attestations of this information.

See: [Adding another permission? A guide](https://github.com/w3c/adding-permissions) and [User considerations for credentials on the Web: In-context explanations](https://github.com/w3c/credential-considerations/blob/main/credentials-considerations.md#in-context-explanations).

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
-          Explain that the API does enable the transmission of personally
-          identifiable information and that it does its best to ensure there is
-          informed consent by the individual, but that the consent might be
-          provided due to exhaustion or not understanding what PII is being
-          transmitted and how to mitigate those concerns.
+        </p>
+        <p>
+          The Digital Credentials API enables the capability to share highly personal,
+          sensitive and at-risk user information with websites via credentials, and
+          reveals parts of the user's browsing activity as well as their intent to
```
npdoty: This language gets at the sensitivity, but we might also note:
the ability to track users online and offline, and permanent unique irrevocable cross-context identifiers

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          authenticate with specific websites to wallets.
```
npdoty: I'm not sure we should use authenticate -- presenting ID should not be for logging in to websites.

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+        <p>
+          It is expected that user agents in their implementation ensure that
+          this information is fully disclosed to the user before an exchange of
+          any user-related information occurs.
+        </p>
+
+        <p class="issue" data-number="252">
+          Should these be normative in the spec?
+        </p>
+
```
```suggestion
< p class="issue" data-number="44">
   API should be designed so the site can provide in-context explanations
</p>
```

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          layers has to be a safe and well-informed user permission experience. As
+          such, integration can be vital for user safety. Some layers may hold
+          information that is inaccessible to others, such as the availability of a
+          user's credentials. Overprompting or prompting without sufficient context
+          could lead to (exploitable) confusion and prompt blindness.
+        </p>
+        <p>
+          For this reason user agents are encouraged to integrate software layers for an ideal user
+          experience when prompting for permission, if they consider it safe to do so.
+          This could happen, for example, if a browser trusts the API contract of an
```
npdoty: I'm not sure this hypothetical is useful. In particular, it seems to actively encourage backsliding on one of the few privacy properties that implementers have agreed on: that the user should confirm before any information is revealed on to the wallet. (As opposed to, we can avoid asking the user for permission because we hope some other layer of the platform will make sure nothing bad happens.)

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          operating system to show an appropriate prompt and thus does not show a prompt itself.
+        </p>
+
+        <h4>
+          User Choice over Wallets
```
npdoty: I think "user choice over wallets" means something else to most people here (about interoperability and the ability to choose the wallet you use to store or present credentials).

Maybe 
```suggestion
          Permission Prior to Wallet Selection
```

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          request.
+        </p>
+
+        <h4>
+          Permission vs. Consent
+        </h4>
+        <p>
+          The permission mediated by the user agent is not consent, which can vary among
+          different legal and regulatory environments and is expected to be collected by
+          the digital wallet before sharing information with the verifier.
```
npdoty: Wallets are poorly positioned to gather user consent. They aren't the party that's requesting or using the information and only have indirect communication with that party (the verifier). The different legal and regulatory environments vary not based on the wallet (I can use my wallet for credential requests no matter where I am or where the website is) but based on other information that is more specific to the user and to the verifier.

"expected to be collected" is also pretty shocking -- it sounds like what we mean is that the wallet should collect consent (because no one else will), but we know they won't.

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+
+          The exchange protocol is responsible for including relevant information for consent,
```
npdoty: If we agree that the wallet should get the user's informed consent (rather than the verifier, or the user agent with help from the verifier), then yes we should document that model and the exchange protocol would need to include additional information communicated from the verifier so that the wallet software can present it to the user for consent. So far, I have not seen any exchange protocol willing to include those properties. Some exchange protocols have specifically removed fields like `purpose` that could be used for this from recent versions.

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+
+        <ul>
+        <li>
+          The privacy policy of the verifier receiving the credential.
+        </li>
+        <li>
+          The purpose of the verifier for requesting the information.
+        </li>
+        <li>
+          Assertions of the verifier's legimitacy and registration for accessing
```
npdoty: An issue for this topic more generally than just EUDI certificates:
https://github.com/w3c-fedid/digital-credentials/issues/136
(Originally titled "define a well-known way for a verifier to indicate registration, validation, trustmark assurances or other necessary info")

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          of user disclosure.
+        </li>
+        <li>
+          Auditing and approving wallet applications only if they meet a certain standard
+          of user disclosure.
+        </li>
+        </ul>
+
+        <p>
+          All of these options have drawbacks, e.g. increasing effort and process to
```
npdoty: Drawbacks for whom?

It seems like many of these involve increased effort for specification authors and user agent implementers. In the [priority of constituencies](https://www.w3.org/TR/design-principles/#priority-of-constituencies), user needs come before decreasing effort involved in specification design.

index.html
npdoty: It seems quite an important part of technical architecture design that we describe which layer fulfills which part of a responsibility to the user.

For example, many people are coerced into the particular wallet that they use, or don't have a choice about the issuer of their credentials or the issuer's policies, or may use a device that restricts their choice about installed software. There might be multiple user agents involved, but the threat model should also consider that some may not be trusted in the same way by the user.

johannhof: Yeah, I can look into that! (maybe in a separate PR)

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+        </li>
+        <li>
+          Assertions of the verifier's legimitacy and registration for accessing
+          the credential, such as <a href="https://github.com/eu-digital-identity-wallet/eudi-doc-architecture-and-reference-framework/blob/main/docs/annexes/annex-2/annex-2-high-level-requirements.md#a2327-topic-27---registration-of-pid-providers-providers-of-qeaas-pub-eaas-and-non-qualified-eaas-and-relying-parties">EUDI access and registration certificates</a>.
+        </li>
+        </ul>
+
+        <p>
+          However, in practice, it is difficult for user agents to assert that this disclosure
+          is universally being made to users, and to judge the quality of the presentation to the user.
```
```suggestion
          is ever being made to users, or to judge the quality of the presentation to the user, or even to detect when the web content directly misrepresents what the user sees in the wallet.
```

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          User agents could improve independently ensure that a sensible level of transparency is
```
npdoty: This sentence doesn't parse.

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          upheld through a number of interventions:
+        </p>
+
+        <ul>
+        <li>
+          Requiring relevant fields such as a Privacy Policy to be passed in the
```
npdoty: Privacy policy links would actually be the lowest priority and least helpful to users trying to make a determination about presenting a credential.

```suggestion
          Requiring relevant fields such as purpose, sharing and retention to be passed in the
```

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+        </li>
+        </ul>
+
+        <p>
+          All of these options have drawbacks, e.g. increasing effort and process to
+          participate in credential exchanges on the Web, and are dependent on the evolution of
+          other initiatives and legal or regulatory frameworks in this problem space.
+          We expect future efforts in the relevant working groups to
+          ensure that the Digitial Credentials API empowers all user agents to collaborate
+          effectively to improve user transparency, permission and consent.
```
npdoty: I don't know what this means. 

Also, permission and consent typically aren't the kinds of concepts that you improve; do you mean something more specific -- increasing the likelihood that users are able to give informed consent, making permission requests more comprehensible or less burdensome, etc.?

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          users retain the choice over whether to forward a credential request to
+          a wallet, and which wallet to select.
+
+          This is due to the information disclosure that happens as part of the request,
+          and the ability for wallets to retain or share this information at time of the
+          request.
+        </p>
+
+        <h4>
+          Permission vs. Consent
```
npdoty, line 1097: If this is to be a key distinction and justification for user agents not taking steps to ensure informed consent, then we should define permission and consent separately and explain what responsibility is handled for each. I anticipate that few users will understand this distinction in any case, but also that few readers of the spec will understand the distinction if we don't define it.

index.html
```diff
@@ -953,14 +953,141 @@ <h3>
...
+          The most critical pieces of context that are needed for a user to make a decision
+          about proceeding with a credential exchange are:
+        </p>
+
+        <ul>
+        <li>
+          The origin of the verifier that requests the credential.
+        </li>
+        <li>
+          The information that is being requested.
```
```suggestion
          The information that is being requested or would be revealed by responding to the request.
```

npdoty (commented):
I've tried to note some of the particular areas of inconsistency, existing open issues and questions regarding the justifications in this section. If the group wants to describe the possible interventions or the privacy/security risks, we can open more issues and link to those issues in a section like this.

As I've described repeatedly for the group, I don't think this approach of 'consent gathered by a wallet app that isn't directly involved in the communication to the user' is promising or even feasible. But if the group does want to commit to it, we should be clear about that, and detail more explicitly the requirements that would entail, on verifiers, user agents, exchange protocols and wallet software to realize consent and minimize security threats.

index.html
TallTed: Removing errant space from @npdoty's suggestion
```suggestion
<p class="issue" data-number="44">
   API should be designed so the site can provide in-context explanations
</p>
```

index.html
johannhof: Yeah, I guess I was struggling for a verb here maybe "identify to" is better

index.html
johannhof: Hm, we specifically call out in this PR that wallets should generally not be involved before user confirmation. I think we should trust the user agent to integrate where it's possible while following the normative and informative spec text.

index.html
johannhof: Yeah, happy to go with that

index.html
johannhof: Sure, I think there could be obligations with the verifier as well (it all really depends, which is kind of the point), and I can clarify that.

We should discuss this question again with the group, but I've not heard from any implementer, nor other members of this group, that it would be appropriate to make consent a concern of this API.

@martinthomson has charted a realistic path to documenting the requirements for protocols, and potentially other parties, in #255. This could include the transmission of what we understand is information required for wallets to get consent, such as data purpose.

index.html
johannhof: I don't think we're in a position to determine this in a way that it could be a strong requirement, though I guess the group could have an opinion on the right model, and we should discuss it. I should soften the language here a bit and include a link to an issue for the group to clarify if we have a stronger recommendation.

index.html
johannhof: I was thinking of the verifiers, not the specification designer, but I don't feel strongly about this sentence.

index.html
johannhof: > Yeah, I can look into that! (maybe in a separate PR)

(This was in response to @c2bo)

> It seems quite an important part of technical architecture design that we describe which layer fulfills which part of a responsibility to the user.
> 
> For example, many people are coerced into the particular wallet that they use, or don't have a choice about the issuer of their credentials or the issuer's policies, or may use a device that restricts their choice about installed software. There might be multiple user agents involved, but the threat model should also consider that some may not be trusted in the same way by the user.

Again, I believe we talk about this when discussing "choice over wallets". I disagree that we need to discuss the entire stack when this is essentially a question of whether wallets are trusted user agents (I agree they're generally not, but also think there may be instances where the user agent could consider them to be).

index.html
johannhof: Thanks, will add it.

index.html
johannhof: Ah yeah, sorry, will fix.

index.html
johannhof: Sure, that seems fair :)

index.html
johannhof: Right, I guess I meant both - A lot of the frameworks in this ecosystem are still being written. The DC API and its spec should support these frameworks in achieving these goals. But also, to your point, we can improve permission UX over time, especially when protocols end up including e.g. purpose information (or if we would require them to via the registry review) - this is all dependent on ecosystem evolution, and not something that the spec can define today.

johannhof: I can try to reframe this!

index.html
johannhof: 👍

index.html
johannhof: I'm not sure I fully agree that this is most critical for a permission prompt (which is a confirmation that the request may be delegated to the selected wallet), so we should probably move this into an issue and add an inline issue here.

johannhof: Actually, thinking about this a bit more, I believe that we can list these but maybe qualify the language a bit - this is guidance for user agents after all, not a normative text.

index.html
johannhof: Thanks, but I'm turning this into a question I think :)

index.html
johannhof: That's what I'm trying to do in this section, but it may be a bad attempt. Happy to iterate on this.

index.html
johannhof: FWIW I'll remove the "potentially even the wallet application" from the PR for now - I agree it sounds a lot like arbitrary 3rd party wallets could be included and its omission doesn't hurt.

johannhof:
@npdoty This is ready for another review - I addressed your comments, I believe, and rewrote the permission vs. consent section.

Note that I rebased on top of #254 (which needs to be merged), causing some noise at the top of the doc - please ignore it and just start from the Permission & Transparency section.

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+          </li>
+          <li>Which credential would be used to share the requested
+          information.
+          </li>
+          <li>Any evaluations and attestations of this information, if
+          available.
+          </li>
+        </ul>
+        <p>
+          It is advised that user agents in their implementation ensure that
```
RByers, line 1052: When you say "user agent" here you're not including the wallet, right? I think of this API as sharing some (but not all) of the "user agency" responsibility between the browser, OS and wallet. In some sense it's actually a core part of the design that the browser is not the only "user agent" because we're intentionally blinding ourselves to the PII that only the wallet knows (a key requirement we heard from many stakeholders who felt the wallet ecosystem was better positioned to respect the user's interests than the browser). I'd actually go so far as to argue that this is the primary difference between the DC API and FedCM - with FedCM the browser is the user's identity agent, and with DC it's the wallet acting as the agent but with the browser providing some extra defense-in-depth to help mitigate risks of buggy, poorly designed or abusive wallet apps. 

I ask because I think some of the above make sense to expose through the browser/OS (and we are) but some (like the purpose) are probably better handled by the wallet. How do we want to talk about that split? I like the text below that talks about it in terms of the OS. One general principle we might want to apply is that when the browser can be relatively confident that some other layer (OS / wallet) is going to meet these criteria than it can rely on that rather than risk double-prompting.

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+        <p>
+          The Digital Credentials API enables the capability to share highly
+          personal, sensitive and at-risk user information with websites via
+          credentials, potentially granting the ability to track users online
+          and offline, through permanent unique irrevocable cross-context
+          identifiers. It also reveals parts of the user's browsing activity as
+          well as their intent to identify to specific websites to wallets. A
+          crucial responsibility of the user agent in a credential request is
+          to gather permission by the user to proceed with the exchange of
+          information.
```
```suggestion
          The Digital Credentials API enables the sharing of highly personal,
          sensitive, and at-risk user information with websites via credentials,
          potentially granting the ability to track users online and offline,
          through permanent, unique, irrevocable, cross-context identifiers. It
          also reveals parts of the user's browsing activity as well as their
          intent to identify to specific websites and/or wallets. One crucial
          responsibility of the user agent in a credential request is to gather
          permission from the user to proceed with the exchange of information.
```

index.html
RByers: I don't understand how a layer of the stack which doesn't even have access to the data being communicated could meaningfully collect consent for it's transfer and use. To me it seems like saying it should be the browser who collects consent is arguing for a different architecture more like FedCM than like the DC API and where EUDI has gone. It makes sense to me that we should have solutions both for where the user & issuer want to rely on trust of the browser to hold credentials and those where they want to rely on slicing that trust off into a separate app which operates under different constraints.

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+        </p>
+        <p>
+          Important context that is needed for a user to make a decision about
+          proceeding with a credential exchange includes:
```
```suggestion
          Important context details that are needed for a user to make an informed
          decision about proceeding with a credential exchange include the following:
```

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+        </p>
+        <ul>
+          <li>The origin of the verifier that requests the credential.
+          </li>
+          <li>The information that is being requested, or would be revealed by
+          responding to the request.
```
```suggestion
          <li>The information that is being requested, or that would be
          revealed by responding to the request.
```

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+          information.
+          </li>
+          <li>Any evaluations and attestations of this information, if
+          available.
+          </li>
+        </ul>
+        <p>
+          It is advised that user agents in their implementation ensure that
+          this information is fully disclosed to the user before an exchange of
+          any user-related information occurs.
```
```suggestion
          It is advised that user agents in their implementation ensure that
          the details listed are fully disclosed to the user before an exchange
          of any user-related information occurs.
```

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+          Depending on the technical architecture of the user's system, it is
+          likely that the definition of a "user agent" includes multiple
+          cooperating layers of the software stack, such as a browser and the
+          operating system. The greatest priority for these layers has to be a
+          safe and well-informed user permission experience. As such,
+          integration can be vital for user safety. Some layers may hold
+          information that is inaccessible to others, such as the availability
+          of a user's credentials. Overprompting or prompting without
+          sufficient context could lead to (exploitable) confusion and prompt
+          blindness.
```
```suggestion
          Depending on the technical architecture of a user's system, it is
          likely that the definition of a "user agent" will include multiple
          cooperating layers of the software stack, such as a browser and the
          operating system. The greatest priority for these layers has to be
          a safe and well-informed user permission experience. As such,
          integration can be vital for user safety. Some layers may hold
          information that is inaccessible by other layers, such as the
          availability of a user's credentials. Overprompting or prompting
          without sufficient context could lead to (exploitable) confusion
          and prompt blindness.
```

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+        </p>
+        <p>
+          For this reason user agents are encouraged to integrate software
+          layers for an ideal user experience when prompting for permission, if
+          they consider it safe to do so. This could happen, for example, if a
+          browser trusts the API contract of an operating system to show an
+          appropriate prompt and thus does not show a prompt itself.
```
```suggestion
          For this reason, user agents prompting for permission are encouraged
          to integrate software layers for an ideal user experience, if they
          consider it safe to do so. This could happen, for example, if a
          browser trusts the API contract of an operating system to show an
          appropriate prompt, and thus does not show a prompt itself.
```

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+        <h4>
+          Permission Prior to Wallet Selection
+        </h4>
+        <p>
+          As part of the user permission flow, the user agent needs to ensure
+          that users retain the choice over whether to forward a credential
+          request to a wallet, and which wallet to select. This is due to the
+          information disclosure that happens as part of the request, and the
+          ability for wallets to retain or share this information at time of
+          the request.
```
```suggestion
          As part of the user permission flow, the user agent needs to ensure
          that users retain the power to choose whether to forward a credential
          request to a wallet, and which wallet to select. This is due to the
          information disclosure that happens as part of the request, and the
          ability of wallets to retain or share this information at the time of
          the request.
```

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+          Permission vs. Consent
+        </h4>
+        <p>
+          The permission mediated by the user agent is not consent, which can
+          vary among different legal and regulatory environments and may need
+          to be collected by the digital wallet before sharing information with
+          the verifier, or by the verifier itself before initiating the
+          request. With frameworks and regulations for obtaining consent still
+          being developed, the API aims to enable the exchange of the necessary
+          information, which could include:
```
```suggestion
          The permission mediated by the user agent is not consent, which has
          specific legal definitions which can vary among different legal and
          regulatory environments and may need to be collected by the digital
          wallet before sharing information with the verifier, or by the
          verifier itself before initiating the request. With frameworks and
          regulations for obtaining consent still being developed, this API
          aims to enable the exchange of the necessary information, which
          could include the following:
```

index.html
```diff
@@ -1017,14 +1017,119 @@ <h3>
...
+        </p>
+        <ul>
+          <li>The privacy policy of the verifier receiving the credential.
+          </li>
+          <li>The purpose of the verifier for requesting the information.
```
```suggestion
          <li>The purpose for which the verifier is requesting the information.
```

TallTed (changes requested):
Language and markup tweaks for clarity

index.html
johannhof: Yeah I wasn't really thinking of wallets here, but I should also say that the list got a lot longer based on @npdoty's suggestions. I agree that some of the above isn't necessarily up to the browser / OS user agent to disclose.

I'm trying to describe the split by separating between permission and consent, where I think there are a few things that a browser (or trusted OS) should absolutely show for virtually all types of credential exchanges as part of the permission flow, and other pieces of disclosure that might slightly depend on the type of credential that's shared and local regulation, which is better placed with the wallet.

We could move some of these list items into the list below, in the consent section, do you have any suggestions?

RByers: Makes sense, thanks. Yeah I'm struggling with how to define the boundary precisely. Personally I don't think there's a clear line, instead I think the job for most of these is primarily the wallets but the browser/OS exist as a fallback / defence-in-depth out of fear that wallets may not always mitigate strongly enough the risk of abusive websites. But maybe that's not right. Certainly it would be cleaner if we could define a clear boundary in separations of concerns between the UA and wallet. I supposed there are a couple things which are certainly clear: we're saying wallet and credential selection is something that the platform should mediate so the user has total control over when communication occurs with a particular wallet app at all. Perhaps everything else is primarily the wallet's responsibility?

npdoty: Are there any indications that wallets are interested or capable in taking on almost all the responsibility of explaining the context to the user, when they aren't involved in any of the communication leading up to the transaction? Are there any existing wallet implementations that do this? Or any sign that groups defining exchange protocols are interested in passing through all of this context information in order to handle consent-gathering, despite the clear evidence I've seen so far that shows that they are specifically removing those capabilities?

johannhof: Nick, I'm very supportive of the argument that user agents should strive to deliver the most complete permission experience to their users with the greatest useful amount of context they can give. And we will work on this, both in the spec and implementation.

However, I think Rick is right that there are pieces of information listed here, such as retention policies and purpose of use, that naturally live with the verifier and the wallet and are usually part of legal / regulatory requirements for disclosure of information as part of consent, and I don't think a browser / OS should just define its own homegrown variant of a legal concept like consent.

But IMO we should absolutely require protocols to support the transfer of information needed for the verifier and/or the wallet to facilitate consent - preventing any parties in the exchange from performing their legal duties sounds like a terrible idea.

For many government issued credentials (which I guess is what we're mostly concerned about here), the hope is that these governments also put in strong regulation around how users consent to their exchange. If that is not the case, I believe we should (and I will in my next PR) encourage browsers to significantly increase the friction in their permission flow.

johannhof: I updated this and moved a few line items to the lower consent section.

index.html
```diff
@@ -1017,14 +1017,117 @@ <h3>
...
+          Permission vs. Consent
+        </h4>
+        <p>
+          The permission mediated by the user agent is not consent, which has
+          specific legal definitions which can vary among different legal and
+          regulatory environments and may need to be collected by the digital
+          wallet before sharing information with the verifier, or by the
+          verifier itself before initiating the request. With frameworks and
+          regulations for obtaining consent still being developed, this API
+          aims to enable the exchange of the necessary information, which
```
mohamedamir, line 1106: I struggle to understand what is meant by this?
Would this part of the protocol request?
More fields will be added to the API request in the future to capture such information?

index.html
```diff
@@ -1017,14 +1017,117 @@ <h3>
...
+        </p>
+        <h4>
+          Permission vs. Consent
+        </h4>
+        <p>
+          The permission mediated by the user agent is not consent, which has
+          specific legal definitions which can vary among different legal and
+          regulatory environments and may need to be collected by the digital
+          wallet before sharing information with the verifier, or by the
+          verifier itself before initiating the request. With frameworks and
```
mohamedamir, line 1104: What does "user agent" mean in this context?
I can read two things:
1- It collects permission
2- It's not a digital wallet.

Although right above this, it mentioned the digital wallet is part of the user agent.

index.html
```diff
@@ -1017,14 +1017,117 @@ <h3>
...
+          <li>Assertions of the verifier's legimitacy and registration for
+          accessing the credential, such as <a href=
+          "https://github.com/eu-digital-identity-wallet/eudi-doc-architecture-and-reference-framework/blob/main/docs/annexes/annex-2/annex-2-high-level-requirements.md#a2327-topic-27---registration-of-pid-providers-providers-of-qeaas-pub-eaas-and-non-qualified-eaas-and-relying-parties">
+            EUDI access and registration certificates</a>.
+          </li>
+        </ul>
+        <p>
+          As more of this information becomes available in a structured format,
+          we expect user agents and this specification to leverage it to
+          improve the user permission experience as well.
```
mohamedamir, line 1130: Ah, I see
So this doesn't exist today, but the user agents should collectively consume that information once available!
Fair enough!
Thank you!

index.html
mohamedamir: Sorry for commenting too soon!
This is explained in the last paragraph!

mohamedamir (approved):
Looks great!
Thanks Johann.

The term "user agent" is very overloaded here, maybe we should think in the future thing about a better naming convention by clearly splitting the responsibilities between different layers in the system.

index.html
mohamedamir: In addition, if the user ends up storing their sensitive information in an untrusted wallet, it's too late anyway!
I think mechanisms like Play store policies can go along way here to protected against that.
It's important however to define those policies. 
What is expected from a trustworthy digital wallet?

johannhof: Yeah, I think I agree with that. I also agree this should be defined but I'd prefer for that to happen in a different document that we can link out from here :)

johannhof:
Thanks @mohamedamir! I filed #263 for the user agent considerations.

@npdoty FYI, I'll merge this PR given editor approval, but you should absolutely feel free to revisit this text and we can work on changes / improvements / additions.

npdoty:
This was merged before we finished the ongoing conversations, or before the group discussed it. I've opened https://github.com/w3c-fedid/digital-credentials/issues/290 to summarize the WG discussion and what we need to do to clarify the terms and responsibilities.

PR #260, opened 2025-06-06
Add initial privacy considerations section for protocol requirements

johannhof:
I added some descriptions of privacy requirements we probably want normatively listed in the protocol review requirements. @martinthomson @npdoty @simoneonofri - would like to get your eyes on this

I think this is a good way to cover topics like selective disclosure, unlinkability, etc. while making it clear that they're not directly under the control of the DC API. WDYT @npdoty?

***
<a href="https://pr-preview.s3.amazonaws.com/johannhof/digital-credentials/pull/260.html" title="Last updated on Jun 18, 2025, 5:33 PM UTC (eba4bdb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/260/bd989f6...johannhof:eba4bdb.html" title="Last updated on Jun 18, 2025, 5:33 PM UTC (eba4bdb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1054,7 +1054,7 @@
           purpose of revocation where possible.
         </p>
         <p class="issue" data-number="280">
-          We should discuss whether and unlinkable revocation techniques are
+          We should discuss whether unlinkable revocation techniques are
           practical enough to be required normatively.
         </p>
         <h5>
--- a/index.html
+++ b/index.html
@@ -1003,6 +1003,19 @@
           unlinkability was achieved in practice. Nonetheless, protocols are
           requested to limit linkability wherever possible.
         </p>
+        <p>
+          Note that unlinkability is exclusively a consideration for attributes
+          that can not be linked to a specific user identity. Inherently linkable
+          attributes such as names, driver's license numbers or phone numbers do
+          not benefit from unlinkability.
+        </p>
+        <p>
+          Through the Digital Credentials API, the user agent can help verifiers
+          and wallets exchange unlinkable attributes, but it can not guarantee
+          that no linkable information is passed between verifiers and wallets.
+          It is recommended that user agents account for this fact in their
+          user permission experience.
+        </p>
         <p class="issue" data-number="279">
           Which level of unlinkability is the goal for this API? Can we
           normatively enforce support for any particular unlinkability features
@@ -1026,8 +1039,11 @@
           application owns this decision. While some wallets can be considered
           user agents, it is generally recommended that the user agent
           implementing the Digital Credentials API designs its permission
-          experience to prevent exposure of a request to the wallet application
-          before user confirmation.
+          experience to prevent <a href=
+          "#permission-prior-to-wallet-selection">exposure of a request to the
+          wallet application</a> before user confirmation (keeping in mind
+          <a href="#multiple-user-agents">considerations for integrating
+          multiple cooperating user agents</a>).
         </p>
         <p>
           Protocols are required to support mechanisms that allow issuers,
@@ -1062,8 +1078,9 @@
         </h5>
         <p>
           User understanding and participation are non-negotiable properties of
-          a credential exchange. The protocol is expected to help all involved parties 
-          enable user participation by providing the information vital for informed permission and/or consent.
+          a credential exchange. The protocol is expected to help all involved
+          parties enable user participation by providing the information vital
+          for informed permission and/or consent.
         </p>
         <h5>
           Support for verifier authorization
@@ -1217,7 +1234,7 @@
           Should the API be designed so the site can provide in-context
           explanations?
         </p>
-        <h4>
+        <h4 id="multiple-user-agents">
           Integrating Multiple User Agents
         </h4>
         <p>
--- a/index.html
+++ b/index.html
@@ -654,7 +654,7 @@
           Once an expression of registration is received via GitHub, the
           registry maintainers will organize the privacy review with the
           <a href="https://www.w3.org/groups/wg/privacy/">Privacy Working
-          Group</a> . Please see the [[[[security-privacy-questionnaire]]] for
+          Group</a> . Please see the [[[security-privacy-questionnaire]]] for
           the kind of questions that will be asked of the protocol you are
           registering.
         </aside>
@@ -960,12 +960,12 @@
         <p>
           The protocol registry for the Digital Credentials API is designed to
           ensure that, among other requirements, supported protocols facilitate
-          specific privacy-enhancing capabilities. Protocols must undergo
-          privacy review by the W3C's <a href=
+          specific privacy-enhancing capabilities. Protocols are required to
+          undergo privacy review by the W3C's <a href=
           "https://www.w3.org/groups/wg/privacy/">Privacy Working Group</a>.
         </p>
         <h4>
-          Exchange Protocol Requirements for User Privacy
+          Exchange Protocol Considerations for User Privacy
         </h4>
         <h5>
           Selective disclosure
@@ -981,30 +981,35 @@
           claims needed.
         </p>
         <h5>
-          Unlinkable presentations and revocation
+          Unlinkable presentations
         </h5>
         <p>
           <a data-cite=
           "credential-considerations#unlinkable-presentations">Unlinkability</a>
           is a property that ensures that, if a user presents attributes from a
           credential multiple times, verifiers cannot link these separate
-          presentations to conclude they concern the same user. Protocols are
-          required to limit linkability wherever possible. This can be done by
-          supporting the presentation of claims that are not inherently
-          linkable without introducing additional linkability, and support for
-          primitives such as <a data-cite=
-          "vc-data-model#zero-knowledge-proofs">Zero-Knowledge Proofs</a> and
-          Blind Signatures.
+          presentations to conclude they concern the same user
+          (verifier-verifier linkability), or that verifiers can not collude
+          with issuers to report the exchange of a credential from a wallet to
+          the issuer (verifier-issuer linkability). The former is a property
+          that can be maintained by the wallet and issuer, e.g. through issuing
+          fresh credentials for individual verifiers.
         </p>
         <p>
-          When used for unlinkable presentations, revocation mechanisms
-          supported by the protocol are expected to maintain unlinkablility. In
-          other words, revocation checking for a credential that was presented
-          unlinkably would not introduce an identifier that can be used to
-          track the user.
+          While the latter is achievable, e.g. through <a data-cite=
+          "vc-data-model#zero-knowledge-proofs">Zero-Knowledge Proofs</a>,
+          design choices of the API such as encrypted responses make it
+          impossible for a user agent to prove that verifier-issuer
+          unlinkability was achieved in practice. Nonetheless, protocols are
+          requested to limit linkability wherever possible.
+        </p>
+        <p class="issue" data-number="279">
+          Which level of unlinkability is the goal for this API? Can we
+          normatively enforce support for any particular unlinkability features
+          as part of the protocol registry inclusion criteria?
         </p>
         <h5>
-          No dependence on "phone home" mechanisms
+          "Phone home" mechanisms
         </h5>
         <p>
           <a data-cite="credential-considerations#no-phoning-home">"Phoning
@@ -1014,22 +1019,43 @@
           and profiling of individuals.
         </p>
         <p>
-          Phoning home can happen in a number of ways, some of which are harder
-          to prevent than others. While linkable presentations have an inherent
-          risk of correlation through verifier to issuer communication,
-          protocol authors can not assume that this will happen on every
-          presentation and are encouraged to protect users against phoning home
-          by wallet applications. Unlinkable presentations, on the other hand,
-          are by design at risk of identification through phone home
-          mechanisms.
+          Similar to unlinkability, it is impossible for user agents to ensure
+          that an issuer isn't actively involved in the creation or validation
+          of credential presentations after a user has given permission to
+          proceed with a credential request. From that point on, the wallet
+          application owns this decision. While some wallets can be considered
+          user agents, it is generally recommended that the user agent
+          implementing the Digital Credentials API designs its permission
+          experience to prevent exposure of a request to the wallet application
+          before user confirmation.
         </p>
         <p>
-          As such, the credential formats used in protocols are required to
-          support offline revocation methods such as <a href=
-          "https://eprint.iacr.org/2024/657.pdf">Cryptographic Accumulators</a>
-          or [[vc-bitstring-status-list]]. It is further expected that protocol
-          design and specification discourages the use of phone home wherever
-          possible.
+          Protocols are required to support mechanisms that allow issuers,
+          wallets and verifiers avoid or reduce the dependence on "phone home"
+          mechanisms.
+        </p>
+        <p class="issue" data-number="279">
+          Which level of unlinkability is the goal for this API? To what degree
+          can the spec mandate restrictions to issuer involvement?
+        </p>
+        <h5>
+          Unlinkable revocation
+        </h5>
+        <p class="issue" data-number="280">
+          A common instance of issuer involvement in a credential exchange is
+          for credential revocation checks. This is particularly challenging
+          when presentations are intended to be verifier-issuer unlinkable.
+          When credential presentations are made unlinkable through the use of
+          e.g. Zero-Knowledge Proofs, the credential formats used in protocols
+          are expected to support offline revocation methods such as <a href=
+          "https://eprint.iacr.org/2024/657.pdf">Cryptographic
+          Accumulators</a>. It is further expected that protocol design and
+          specification discourages the involvement of verifiers for the
+          purpose of revocation where possible.
+        </p>
+        <p class="issue" data-number="280">
+          We should discuss whether and unlinkable revocation techniques are
+          practical enough to be required normatively.
         </p>
         <h5>
           Support for user transparency, permission and consent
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+        <h5>
+          Selective disclosure
+        </h5>
+        <p>
+          <a data-cite=
+          "credential-considerations#selective-disclosure">Selective
+          disclosure</a> is a fundamental technique for <a data-cite=
+          "privacy-principles#data-minimization">data minimization</a> that
+          allows holders to share the minimum required information that is
+          needed by a verifier. Protocols can facilitate this selective
```
```suggestion
          requested by a verifier. Protocols can facilitate this selective
```

index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+          disclosure by allowing the verifier to specify the exact claims
+          needed, as well as through support of cryptographic primitives such
+          as <a data-cite="vc-data-model#zero-knowledge-proofs">Zero-Knowledge
+          Proofs</a>.
```
mohamedamir: I think this belongs better to the next section "Unlinkable presentations and revocation".
IIUC, ZKP has little to do with "Selective disclosure"

mohamedamir: which I now see you added already.

johannhof: Yeah, you're right, I can remove it from this section, thanks!

index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+          Encrypting credential responses
+        </h5>
+        <p>
+          To prevent exposure of user information to other parties in transit,
+          protocols are required to support and mandate encrypted responses in
+          a credential exchange.
+        </p>
+        <p class="issue" data-number="255">
+          Actually write these in the protocol requirements
+        </p>
```
simoneonofri, line 1119:
```suggestion
        </p>
        <h5>
           Sign presentation requests
         </h5>
         <p>
          To ensure the integrity of the submission request and prevent it from 
          being modified in transit,  protocols are required to support and 
          mandate encrypted requests in a credential exchange.          
         </p>
```

On the one hand, it's a safety issue, but I think it's fine here for now.

index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+          is a property that ensures that, if a user presents attributes from a
+          credential multiple times, verifiers cannot link these separate
+          presentations to conclude they concern the same user. Protocols are
+          required to limit linkability wherever possible. This can be done by
+          supporting the presentation of claims that are not inherently
+          linkable without introducing additional linkability, and support for
+          primitives such as <a data-cite=
+          "vc-data-model#zero-knowledge-proofs">Zero-Knowledge Proofs</a> and
+          Blind Signatures.
+        </p>
```
simoneonofri, line 997: Shall we consider the various cases of unlinkability? (along the lines of EUDI)
a. same verifier should not recognise the same holder
b. two verifiers should not recognize the same holder
c. issuer should not know when a credential is presented or verified
d. even with the collusion of verifier and issuer
e. or in the event of observation by a third party external to the three main actors?

simoneonofri (changes requested):
Great work @johannhof , thank you.
I added a requirement and a comment on the unlinkability part.

index.html
RByers: You mean "signed requests" not "encrypted requests", right? Requiring it to be supported makes sense to me, but I don't think it can reasonably be mandated as always required.

RByers (approved):
This seems like a great start to me. I'd be supportive of landing this and iterating.

index.html
johannhof: +1 to Rick here

index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+          "credential-considerations#selective-disclosure">Selective
+          disclosure</a> is a fundamental technique for <a data-cite=
+          "privacy-principles#data-minimization">data minimization</a> that
+          allows holders to share the minimum required information that is
+          requested by a verifier. Protocols are expected to facilitate
+          selective disclosure by allowing the verifier to specify the exact
+          claims needed.
+        </p>
+        <h5>
+          Unlinkable presentations and revocation
```
martinthomson: These are two entirely different things (indeed, they are often in tension).

index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+          presentations to conclude they concern the same user. Protocols are
+          required to limit linkability wherever possible. This can be done by
+          supporting the presentation of claims that are not inherently
+          linkable without introducing additional linkability, and support for
+          primitives such as <a data-cite=
+          "vc-data-model#zero-knowledge-proofs">Zero-Knowledge Proofs</a> and
+          Blind Signatures.
+        </p>
+        <p>
+          When used for unlinkable presentations, revocation mechanisms
```
martinthomson:
```suggestion
          When used for unlinkable presentations, any revocation mechanisms
```

index.html
martinthomson: I generally agree, except that this can be simpler.  The IETF documents focus on two things:

1. two presentations from the same credential cannot be linked to each other
2. a presentation cannot be linked to an credential

That addresses most of the cases.  (This text only covers the presentation linkability problem, which is a big problem.)  Especially if you include a blanket assumption of collusion, which is generally consistent with the approach we take elsewhere.

index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+          supported by the protocol are expected to maintain unlinkablility. In
+          other words, revocation checking for a credential that was presented
+          unlinkably would not introduce an identifier that can be used to
+          track the user.
+        </p>
+        <h5>
+          No dependence on "phone home" mechanisms
```
martinthomson: I wouldn't use the "phone home" colloquialism here, but instead focus on eliminating active involvement by issuers in the creation or validation of presentations.

There are cases where active involvement is acceptable, in particular, where there is necessarily linkability (because the information in the presentation is identifying, usually).  That just highlights the conflicted nature of this API.

index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+          presentation and are encouraged to protect users against phoning home
+          by wallet applications. Unlinkable presentations, on the other hand,
+          are by design at risk of identification through phone home
+          mechanisms.
+        </p>
+        <p>
+          As such, the credential formats used in protocols are required to
+          support offline revocation methods such as <a href=
+          "https://eprint.iacr.org/2024/657.pdf">Cryptographic Accumulators</a>
+          or [[vc-bitstring-status-list]]. It is further expected that protocol
```
martinthomson: Not sure if an IACR eprint is acceptable here.  The bitstring status list relies on verifier-issuer linkability.  (I'm not sure if the same is true for accumulators.)

index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+          User understanding and participation are non-negotiable properties of
+          a credential exchange. The protocol is expected to support all
+          involved parties in enabling user participation and provide them with
+          any of the information that is vital for informed permission or
+          consent. In particular, this involves granting verifiers the ability
+          to pass pertinent information along to the user agent or the wallet
+          in an unencrypted, structured text format.
+        </p>
+        <h5>
+          Encrypting credential responses
```
martinthomson, line 1108: I'm not sure that I agree that this is a requirement.  If the recipient is the code running in the webpage, then the "in transit" that applies is entirely within the API, where we generally do not worry about interception.

I know that there has been a bunch of discussion about the reason for this arrangement, but it seems like this needs better motivation than is provided here.

The same goes for @simoneonofri's signature suggestion, which suffers the same basic problem, plus the usual signature problems (who is signing? what are the expectations of the recipient?).  At least here I can see how the recipient might offer encryption keys as part of its request.

index.html
```diff
@@ -647,20 +647,20 @@ <h3>
...
+      <a href="https://www.w3.org/groups/wg/fedid/">Federated Identity Working
+      Group</a>.
         <aside class="note" title="Organizing reviews">
           Once an expression of registration is received via GitHub, the
           registry maintainers will organize the privacy review with the
-          <a href="https://www.w3.org/Privacy/IG/">Privacy Interest Group</a> .
-          Please see the [[[[security-privacy-questionnaire]]] for the kind of
-          questions that will be asked of the protocol you are registering.
+          <a href="https://www.w3.org/groups/wg/privacy/">Privacy Working
+          Group</a> . Please see the [[[[security-privacy-questionnaire]]] for
```
martinthomson: Is it three or four brackets?

index.html
```diff
@@ -946,6 +946,115 @@ <h3>
...
+          Support for user transparency, permission and consent
+        </h5>
+        <p>
+          User understanding and participation are non-negotiable properties of
+          a credential exchange. The protocol is expected to support all
+          involved parties in enabling user participation and provide them with
+          any of the information that is vital for informed permission or
+          consent. In particular, this involves granting verifiers the ability
+          to pass pertinent information along to the user agent or the wallet
+          in an unencrypted, structured text format.
```
martinthomson: I don't think that this is anywhere near sufficient.  Let's trace the requirements:

1. "European Digital Identity Wallets shall, in particular: (a) support common protocols and interfaces:
[...] (vii) for authenticating and identifying relying parties by implementing authentication mechanisms in accordance with Article 5b;" --> "Member States shall make the information referred to in paragraph 2 publicly available online in electronically signed or sealed form suitable for automated processing." -- https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=celex%3A32024R1183
2. The EUDI ARF https://eu-digital-identity-wallet.github.io/eudi-doc-architecture-and-reference-framework/latest/architecture-and-reference-framework-main/#6632-wallet-unit-authenticates-the-relying-party-instance shows more phoning home than ET
3. openid4vp is better, but its connection to the EUDI ARF is pretty tenuous: "The Verifier Info parameter is optional. Wallets MAY use them to make authorization decisions or to enhance the user experience, but they SHOULD ignore any unrecognized or unsupported Verifier Info types." -- https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-verifier-info
4. This document.

The net effect of this is that this is a lack of a clear path for satisfying the requirements, or even for ensuring that the requirements are met.  It is my view that verifier validation and authorization is not optional, even if various upstream documents make it optional.  This system cannot achieve any reasonable set of accountability goals if random actors can ask for credentials without there being some sort of non-repudiable log that records the fact that they request that information.

"Granting verifiers the ability to pass \[..] information \[..] unencrypted, structured text format" doesn't cut it.  Not even close.

mohamedamir (approved):
Happy to merge this and iterate!
Looks like a great start for me!

index.html
npdoty: The view in today's WG call was that the group believed that likely no states, even European member states, would actually require that requesting government-issued credentials be verified in advance or that that information would be available to the user (either via the user agent or the wallet). Although @marcoscaceres in #262 argues that the privacy threats are limited because issuers would enforce pre-registration/verification requirements.

index.html
johannhof: Clarified 1:1 that I meant unlinkable revocation specifically, but also that we should file an issue on whether that is a thing we can require (or even recommend) in practice.

index.html
johannhof: Yeah... I think it's one pair too much (I should point out that I didn't write those brackets 😛)

johannhof: Oh it's just the opening bracket. Anyway, I'll fix it.

index.html
johannhof: Yeah I agree that we should expand this to verifier - issuer linkability, but talking to Martin it sounds like we should have a broader discussion about linkability in the context of the DC API, so I'm leaving this in with a big issue as caveat.

index.html
johannhof: Yeah I think the best we can do for now is, to Martin's point, mandate that verifier validation and authorization must be readable by the user agent, in case they want to restrict (either entirely disallow or add friction) instances where no or invalid authorization is present.

index.html
johannhof: Yeah, the in transit thing is fair, I can take another stab at justifying this.

index.html
johannhof: Hmm, I don't think this can be a requirement for now then, and we should probably discuss this as part of the unlinkability discussion.

One thing that's important here is that we can't actually enforce unlinkability, in the same way that we can't assume that verifiers and issuers will necessarily collude. My goal was to ensure that protocols offer particularly issuers that want to restrict verifier linkability, and reduce their own exposure to information, the means to do so, but I probably should discuss this more with the group.

index.html
johannhof: It's an informative reference which to me means "anything goes" but I honestly have no idea what is actually acceptable here.

As mentioned above, I think I'm going to relativize the language a bit here for now pending further discussions, after which we may be able to have clearer protocol level guidance.

index.html
johannhof: Filed #279

johannhof:
> Thanks for starting here. Looking at this, I'm further convinced of a few high-level points.
> 
> Overall, I have a preference for an API that only claims to provide identifiable information. That is, we do not seek to solve trait-based authorization flows, such as age verification, with this API. That is because the set of requirements that apply to those cases are fundamentally different to those that involve identification.

I think I'm with you on not claiming that this API can provide any particular guarantees in terms of unlinkability, and that we should probably declare that all requests through the API are to be considered identifiable by our threat model. With that said, I do believe there are practical implementation differences (it's theoretically possible to do ZKPs even if we can't verify it), that will be relevant for users. I tried to express these things in the considerations but it's a thin line to walk and I'm happy to keep iterating (but would prefer to do that in a follow-up PR).

> The other thing here is that this does nothing to constrain the set of protocols that might be involved. There's a bigger discussion about the relative importance of verifier accountability that this brushes aside, effectively deferring to protocols.
> 
> I would be happier with this if there was a clear articulation of the requirement to be able to trace the authorization by which a verifier makes the request for a credential. I realize that this might not be achievable under the current suite of protocols and whatnot, but that's no reason to let this slide.

I filed #281 and also added some text specifically for verifier authorization. It's probably a bit more vague than you'd want it to be, but it's early days for verifier auth systems and we can iterate from the discussion in that issue.

index.html
johannhof: Added some text and filed #281

johannhof:
@mohamedamir can you take a quick look and confirm that you're still happy to merge and iterate based on this? Thanks!

index.html
```diff
@@ -945,6 +946,163 @@ <h3>
...
+        <p>
+          Similar to unlinkability, it is impossible for user agents to ensure
+          that an issuer isn't actively involved in the creation or validation
+          of credential presentations after a user has given permission to
+          proceed with a credential request. From that point on, the wallet
+          application owns this decision. While some wallets can be considered
+          user agents, it is generally recommended that the user agent
+          implementing the Digital Credentials API designs its permission
+          experience to prevent exposure of a request to the wallet application
+          before user confirmation.
```
mohamedamir: Should we mention here that this can happen at any layer in the system and not necessarily in the user agent?
or since the phrasing is "it is generally recommended ...." we don't need that?
WDYT?

index.html
```diff
@@ -945,6 +946,163 @@ <h3>
...
+          selective disclosure by allowing the verifier to specify the exact
+          claims needed.
+        </p>
+        <h5>
+          Unlinkable presentations
+        </h5>
+        <p>
+          <a data-cite=
+          "credential-considerations#unlinkable-presentations">Unlinkability</a>
+          is a property that ensures that, if a user presents attributes from a
```
mohamedamir, line 989: This principle is exclusively relevant for attributes that cannot be linked to a specific identity. In cases involving inherently linkable data, like a passport number, this point is moot.
Not sure if this is self-explanatory or worth highlightng!

index.html
```diff
@@ -945,6 +946,163 @@ <h3>
...
+          When credential presentations are made unlinkable through the use of
+          e.g. Zero-Knowledge Proofs, the credential formats used in protocols
+          are expected to support offline revocation methods such as <a href=
+          "https://eprint.iacr.org/2024/657.pdf">Cryptographic
+          Accumulators</a>. It is further expected that protocol design and
+          specification discourages the involvement of verifiers for the
+          purpose of revocation where possible.
+        </p>
+        <p class="issue" data-number="280">
+          We should discuss whether and unlinkable revocation techniques are
```
```suggestion
          We should discuss whether unlinkable revocation techniques are
```

index.html
```diff
@@ -945,6 +946,163 @@ <h3>
...
+        </p>
+        <h5>
+          Support for user transparency, permission and consent
+        </h5>
+        <p>
+          User understanding and participation are non-negotiable properties of
+          a credential exchange. The protocol is expected to support all
+          involved parties in enabling user participation and provide them with
+          any of the information that is vital for informed permission or
+          consent.
```
mohamedamir: Optional: I find this a bit hard to parse
```suggestion
          a credential exchange. The protocol is expected to help all involved parties 
          enable user participation by providing the information vital for informed permission and/or consent.
```

index.html
```diff
@@ -945,6 +946,163 @@ <h3>
...
+        </p>
+        <h5>
+          Support for verifier authorization
+        </h5>
+        <p>
+          Verifier authorization refers to the process by which a verifier
+          proves its identity and demonstrates that it is legitimately entitled
+          to request specific attributes or credentials. This is particularly
+          useful when exchanging sensitive data, such as from government-issued
+          credentials. Authorizing verifiers can limit unnecessary or abusive
```
```suggestion
          credentials. Verifier authorization can limit unnecessary or abusive
```

mohamedamir:
> @mohamedamir can you take a quick look and confirm that you're still happy to merge and iterate based on this? Thanks!

I took another look. (and added some minor comments)
It still looks great especially with more concrete areas for iteration by linking those open issues!
Great work Johann!
Your work in unblocking this is invaluable!

index.html
johannhof: Yeah, I could link out to the previously merged permissions section where I point this out already?

index.html
johannhof: I felt like I had that in there once, it's worth repeating I guess :)

index.html
johannhof: I'm going to punt this on a follow-up issue in the interest of iterating quickly :)

johannhof: Actually I guess we have several issues for signed requests, most recently #281

index.html
johannhof: I opened #279 and #280 for discussion unlinkability and revocation. Feel free to file additional issues or comments if that doesn't capture it all.

PR #271, opened 2025-06-12
Adds additional terminology

### LINKED ISSUE #270: protocol identifier is undefined
For #221, we need a precise definition of "protocol identifier", that can be used for processing the string that comes in.

timcappalli:
Closes #270 

Adds additional missing terminology.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/271.html" title="Last updated on Jun 12, 2025, 3:09 PM UTC (f733d02)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/271/bd989f6...f733d02.html" title="Last updated on Jun 12, 2025, 3:09 PM UTC (f733d02)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -170,6 +170,60 @@
     </p>
     <dl class="definitions" data-sort="" data-cite="vc-data-model-2.0">
       <dt>
+        <dfn>Client</dfn>
+        <br />
+        <dfn>Digital Credentials API Client</dfn>
+      </dt>
+      <dd>
+        Also referred to herein as simply a [=client=]. See also
+        [=Conforming User Agent=]. A [=Digital Credentials API Client=] is an
+        intermediary entity typically implemented in the user agent
+        (in whole, or in part). Conceptually, it underlies the Digital
+        Credentials API and embodies the implementation of the 
+        {{DigitalCredential/[[Create]](origin, options, sameOriginWithAncestors)}}
+        and {{DigitalCredential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}}
+        internal methods. It is responsible for both marshalling the inputs for
+        the underlying credential provider operations, and for returning the
+        results of the latter operations to the Digital Credential API's
+        callers.
+      </dd>
+      <dt>
+        <dfn>Client Device</dfn>
+      </dt>
+      <dd>
+        The hardware device on which the [=client=] runs, for example a
+        smartphone, a laptop computer or a desktop computer, and the operating
+        system running on that hardware.
+
+        The distinctions between a [=client device=] and a [=client=] are:
+        <ul>
+          <li>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
This tries to add too many things at once. 

Please split this up into multiple requests. In particular, the client part needs quite a bit of changes. We also don't need a conforming user agent, IMO.    

The protocol identifier is not precise enough as written, please see #276 instead.

marcoscaceres (changes requested):
Please split this into multiple PRs.

index.html
```diff
@@ -169,6 +169,60 @@ <h2>
...
+        <dfn>Conforming User Agent</dfn>
+      </dt>
+      <dd>
+        A user agent implementing, in cooperation with the underlying
+        [=client device=], the Digital Credentials API and algorithms given in
+        this specification, and handling communication between credential
+        providers, verifiers, and issuers.
+      </dd>
+      <dt>
+        <dfn>Credential provider</dfn>
```
MasterKale, line 220: Teeny tiny nitpick, for sake of consistency:

```suggestion
        <dfn>Credential Provider</dfn>
```

index.html
```diff
@@ -169,6 +169,60 @@ <h2>
...
+      </dt>
+      <dd>
+        The hardware device on which the [=client=] runs, for example a
+        smartphone, a laptop computer or a desktop computer, and the operating
+        system running on that hardware.
+
+        The distinctions between a [=client device=] and a [=client=] are:
+        <ul>
+          <li>
+            a single [=client device=] MAY support running multiple clients,
```
MasterKale, line 201: Should "clients" here get linked like "client device" does?

```suggestion
            a single [=client device=] MAY support running multiple [=client | clients=],
```

index.html
```diff
@@ -169,6 +169,60 @@ <h2>
...
+            i.e., browser implementations, which all have access to the same
+            credential providers available on that [=client device=], and
```
MasterKale, line 203: Would you want to link to the new "credential provider" definition here too?

```suggestion
            [=credential provider | credential providers=] available on that [=client device=], and
```

MasterKale (commented):
Setting aside process that might see this PR get split up, and accounting for the addition of a definition for "protocol identifier" via #276, this seems like a great list of terms to bring into the spec.

"Conforming User Agent" as a glue definition has value for sake of being explicit, though if we're going to end up referring to browsers as "clients" throughout the spec then the value of "confirming user agent" may get diminished somewhat.

Do we need a separate term for a device that has access to a credential provider (i.e. a smartphone) but is separate from the client device running the client invoking the DC API (i.e. a laptop)?

index.html
timcappalli: yeah didn't get a chance to add it

PR #277, opened 2025-06-16
Security Considerations: Writing first round of threats and mitigatio…

simoneonofri:
(Web API level)

A first draft of the identified threats and potential mitigations (some already applied), particularly at the Web API level.

*Threats*
- SOP Violation
- Fingerprinting and Cross-Device Tracking
- Cross Site Scripting (XSS), Cross Site Request Forgery (CSRF)
- Clickjacking & UI redressing
- Reply Attack
- Quishing
-  Phishing/Harvesting

*Mitigations (already implemented or to be considered)*
-  Data Minimization
-  Secure contexts
-  Limit API usage
-  Informing the user
- Transient activation

Things to consider:
- What else could go wrong (if there are other threats)
- What can we do about the threats we have identified
- Do we like the countermeasures we already have in place
- Are there other mitigations to consider or write down
- Overlaps/joint with Privacy 

[cc'ing @Sh-Amir and @ZAnsaroudi]

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/277.html" title="Last updated on Jul 7, 2025, 4:48 AM UTC (3dc5e39)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/277/4624c60...3dc5e39.html" title="Last updated on Jul 7, 2025, 4:48 AM UTC (3dc5e39)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,18 +18,21 @@
           email: "marcosc@apple.com",
           company: "Apple Inc.",
           companyURL: "https://apple.com",
+          w3cid: 39125,
         },
         {
           name: "Tim Cappalli",
           email: "tim.cappalli@okta.com",
           company: "Okta",
           companyURL: "https://okta.com",
+          w3cid: 111190,
         },
         {
           name: "Mohamed Amir Yosef",
           email: "mamir@google.com",
           company: "Google Inc.",
           companyURL: "https://google.com",
+          w3cid: 158857,
         },
         {
           name: "Sam Goto",
@@ -37,11 +40,13 @@
           company: "Google Inc.",
           companyURL: "https://google.com",
           retiredDate: "2025-05-02",
+          w3cid: 50308,
         },
         // Add additional editors here.
         // https://github.com/w3c/respec/wiki/editors
       ],
       latestVersion: null,
+      canonicalURI: "TR",
       shortName: "digital-credentials",
       specStatus: "ED",
       edDraftURI: "https://w3c-fedid.github.io/digital-credentials/",
@@ -60,6 +65,25 @@
           href: "https://github.com/w3c-cg/threat-modeling/blob/main/models/decentralized-identities.md",
           authors: ["Simone Onofri"],
           date: "2025-05-28",
+          publisher: "W3C"
+        },
+        "custom-schemes": {
+          title: "Concerns with custom schemes for identity presentment",
+          href: "https://github.com/w3c-fedid/digital-credentials/blob/main/custom-schemes.md",
+          authors: ["Rick Byers"],
+          date: "2024-05-01",
+          publisher: "W3C"
+        },
+        "presenting-credentials-on-the-web": {
+          title: "Presenting Credentials on the Web",
+          href: "https://docs.google.com/document/d/1Ppaz_EnhzHqPOz5UusRJvbSunh-RXPWgJ3Np_TM2EE0/",
+          authors: ["Simone Onofri"]
+        },
+        "identity-web-impact": {
+          title: "Identity & the Web",
+          href: "https://www.w3.org/reports/identity-web-impact",
+          authors: ["Simone Onofri"],
+          date: "2025-02-25",
           publisher: "W3C"
         },
       },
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -865,6 +865,136 @@ <h3>
           the session management utilized on the website.
         </p>
       </section>
+  <h3 id='threats-and-attacks'>Threats and Attacks</h3>
+  <h4 id='same-origin-policy-violations'>Same-Origin Policy Violations</h4>
+  <p>SOP violation occurs when a script or resource from one origin breaks the browser's foundational Same-Origin
+      Policy—intended to segregate data between origins—by accessing or tampering with another origin's data, cookies,
```
```suggestion
      Policy, which is intended to segregate data between origins, by accessing or tampering with another origin's data, cookies,
```

index.html
```diff
@@ -865,6 +865,136 @@ <h3>
...
+    <p>Permission controls require explicit user permission before any API use is allowed. This ensures that users know when
+        their device’s vibration feature is being accessed. Permissions must be granted per site, ensuring each website
+        obtains consent separately (e.g., not subdomain permission).</p>
+    <p><strong>Threats and Attacks</strong></p>
+    <ul>
+        <li>Fingerprinting and Cross-Device Tracking</li>
+    
+    </ul>
+    <h4 id='transient-activation'>Transient activation</h4>
+    <p>Transient activation ensures that powerful or sensitive web features can only be triggered immediately following a
```
marcoscaceres: This redefines "transient activation", which is not appropriate. This should only say, "The API relies on transient activation, preventing websites from calling the API without user interaction."

index.html
```diff
@@ -865,6 +865,136 @@ <h3>
...
+        and subdomains</p>
+    <h4 id='informing-the-user'>Informing the user</h4>
+    <p>When handling sensitive API calls involving personal data or credentials, including explicit user approval at the
+        point of use is important. Instead of allowing background or automated exchanges, the user agent must present a
+        clear, interactive prompt, such as a chooser UI, that shows the requesting origin and the data being accessed. This
+        ensures users are fully informed and retain control over each operation, significantly reducing risks from silent
+        data sharing or unintended leaks.</p>
+    <p>The API prompts the user to show the request's origin before proceeding.</p>
+    <h4 id='permission-request'>Permission request</h4>
+    <p>Permission controls require explicit user permission before any API use is allowed. This ensures that users know when
```
marcoscaceres: This is not true. This should be removed. There is no permission screen. 

We rely on Permission Policy only. And only to prevent third-party iframes from calling the API.

index.html
```diff
@@ -865,6 +865,136 @@ <h3>
...
+      believe they are talking directly to each other.</p>
+  <p><strong>Mitigations</strong></p>
+  <ul>
+      <li>Secure contexts (Web API level) </li>
+      <li>Encryption (Protocol level) </li>
+      <li>Signature (Protocol level) </li>
+      <li>Nonce (Protocol level)</li>
+  
+  </ul>
+  <h4 id='reply-attack'>Reply Attack</h4>
```
```suggestion
  <h4 id='reply-attack'>Replay Attack</h4>
```

index.html
```diff
@@ -865,6 +865,136 @@ <h3>
...
+      and later replays that same exchange to another party. This tricks systems into accepting old, valid credentials as
+      fresh, leading to unauthorized authentication or credential disclosure.</p>
+  <p><strong>Mitigations</strong></p>
+  <ul>
+      <li>Secure contexts (Web API level) </li>
+      <li>Nonce (Protocol level)</li>
+  
+  </ul>
+  <h4 id='quishing'>Quishing</h4>
+  <p>Quishing occurs when a malicious site tricks the user into replacing a legitimate QR code, tricking it into
```
npdoty, line 1142: I don't understand "replacing" here. It seems like the attack is presenting the user with a maliciously-crafted QR code which when followed, will lead to a credential presentation request that will deliver the results to an unexpected party (either the attacker, or as a confused deputy confirming the user's identity for a request that the attacker made to some other verifier). Maybe the attacker is inserting it in a place where it looks to the user like a legitimate request from a different verifier, and that's a kind of replacement?

index.html
```diff
@@ -865,6 +865,136 @@ <h3>
...
+    <h4 id='informing-the-user'>Informing the user</h4>
+    <p>When handling sensitive API calls involving personal data or credentials, including explicit user approval at the
+        point of use is important. Instead of allowing background or automated exchanges, the user agent must present a
+        clear, interactive prompt, such as a chooser UI, that shows the requesting origin and the data being accessed. This
+        ensures users are fully informed and retain control over each operation, significantly reducing risks from silent
+        data sharing or unintended leaks.</p>
+    <p>The API prompts the user to show the request's origin before proceeding.</p>
+    <h4 id='permission-request'>Permission request</h4>
+    <p>Permission controls require explicit user permission before any API use is allowed. This ensures that users know when
+        their device’s vibration feature is being accessed. Permissions must be granted per site, ensuring each website
```
npdoty: vibration?

marcoscaceres: Yeah, this is kinda what I mean... there's a lot of copy/pasta here and it's not adding a lot of value.

simoneonofri: I am trying to collect the common "defenses" of a typical Web API (since I don't think they are tracked elsewhere, and I am gathering what I have from various reviews)

@marcoscaceres What threats do you think are specific to the DC API and, therefore, need to be mitigated?

As we can see, several countermeasures are being "imported" from a generic model for Web APIs (basically this PR). In contrast, others are more related to the ecosystem (and therefore perhaps managed at the protocol level, credentials profile, etc.).

Thanks :)

simoneonofri: > vibration?

@npdoty fixed! thx

index.html
simoneonofri: I removed the Permission API and wrote about the Permission Policy what you noted. thank you!

index.html
simoneonofri: For the framing I was giving it, the QR code is managed directly by the API and displayed securely and not via an  (controlled by an adversary).

Does that make sense to you, or do you think we should frame it differently?

index.html
simoneonofri: updated! thx

simoneonofri:
@marcoscaceres @timcappalli @npdoty @RByers, thanks for the feedback. We did a full rewrite of the section here: 
https://github.com/w3c-fedid/digital-credentials/wiki/Security-Considerations-Section

If you have any feedback/questions before Friday, they are welcome so we can converge on Friday

TallTed:
@simoneonofri — I suggest moving the draft from the wiki, which is surprisingly difficult to edit (e.g., I have no edit link), to either a distinct fork, or a PR against the existing section, upon which it should be easy for any of us to submit suggested revisions. (One such revision is to change a number — but not all! — of instances of `Digital Credentials` to the singular, especially in `the Digital Credential API` which reads far better than `the Digital Credentials API`).

simoneonofri:
@RByers @mohamedamir @TallTed thank you. 

We have been working on a Google Docs, if that's helpful, or where ever you prefer.

Issue #265, opened 2025-06-11
voluntary usage and protections against coercion


npdoty:
A concern for both privacy and free expression is that if sites learn that digital credentials are broadly available, they may be increasingly willing (or increasingly coerced) into demanding presentation of credentials in order to access services. That may exclude, discriminatorily, people who are unwilling or unable to present a credential for whatever reason, and may increase the privacy and security risks to those who are able to present a credential.

API availability may provide a signal that leads to this sort of coercion, so we should document and mitigate it.

Don't let it be detectable that the user has a credential, or the site might demand it. (Withholding or greasing is often considered as a mitigation here, although it might not easily apply in this case.)

Require in some way that sites provide an alternative, in order to maintain access and decrease coercion for use of digital credentials.

johannhof:
Yes, this is an important issue, which is also mentioned in my PR #262, but coercion from visibility into credential availability is a good perspective that I haven't really considered yet. I'm writing a PR for #245 which I believe could be an appropriate home for this topic?

Two thoughts on your suggestions: 

> Don't let it be detectable that the user has a credential, or the site might demand it. (Withholding or greasing is often considered as a mitigation here, although it might not easily apply in this case.)

Maybe I'm missing something, but how is this detectable in a way that we'd grease this information?

> Require in some way that sites provide an alternative, in order to maintain access and decrease coercion for use of digital credentials.

I like the intent behind this, but have to say this sounds nearly impossible to do, unless I'm not considering something. Even if we, say, were to require websites to offer a PDF document upload besides accepting digital credentials, they could decide not to process it anyway.

hlflanagan:
Discussed on the [10 June 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-06-10-DCAPI-B-notes.md#265-voluntary-usage-and-protections-against-coercion).

PR #262, opened 2025-06-11
Write initial privacy considerations for unnecessary use

johannhof:
@npdoty there are some things here that strongly touch on ethical considerations which I think you'd probably be able to say much more detailed and eloquently than I can, so I'm happy to get your thoughts and potential contributions (maybe as a follow-up). But I hope this is a solid start that captures the main issues and, where possible, mitigations.

Issue #255, opened 2025-06-04
Protocol registry: Reviewing is not sufficient


martinthomson:
There are two requirements for protocols that I think need further elaboration:

> MUST have undergone privacy review [...]

And

> MUST have undergone security review [...]

Technically, a review saying "this protocol is awful in every way" satisfies these criteria.

It would be more useful if there were a set of concrete privacy and security requirements that a protocol needed to satisfy, such a review would be able to say whether a standard was achieved or not.  It might be the case that there are subjective elements to a review, but there should also be a minimum bar that each protocol needs to clear.

This goes beyond the present set of requirements in the current [inclusion criteria](https://w3c-fedid.github.io/digital-credentials/#general-inclusion-criteria).  I don't have a comprehensive list to hand, but one should be possible to develop.  And once developed, that list should be in the spec.  For instance, does the protocol depend on [phoning home](https://nophonehome.com/)?  Does the protocol (or the formats it conveys) guarantee unlinkability of presentations?  Or - given that unlinkability doesn't make sense for some use cases - under what conditions does the API require the protocol provide unlinkability?  What sort of transparency affordances does the protocol include?  What sorts of covert channels are acceptable?

simoneonofri:
Hello, yes, this is similar to listing the mitigations that need to be managed outside the API (e.g., protocol level)

We can evaluate using an "Implementation Requirements" section, or produce a questionnaire.

This list can help speed up the reviews.

On the technical side, as we're talking about protocols, we can start from RFC 3552 (Security), RFC 6973 (Privacy), and RFC 9620 - draft (Harms). On a high level, we can model them with LINDDUN. We had discussed this in a side meeting at IETF 120.

npdoty:
I think this was the point I was raising in https://github.com/w3c-fedid/digital-credentials/issues/226

hlflanagan:
Discussed as part of a discussion on #226 . See [notes](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-02-18-DCAPI-B-notes.md#privacy-review-for-registry-inclusion-226).

hlflanagan:
Discussed as part of issue triage on [13 May 2026 call](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-05-13-DCAPI-B-notes.md#issue-triage-starting-with-issue-198-oldest-to-newest). Still need to document requirements somewhere (WG Note?) as to what privacy review future protocols will need to go through to be included in future versions of the DC API.

PR #283, opened 2025-06-18
Add initial privacy considerations for fingerprinting and data leakage

johannhof:
@npdoty this covers parts of #265, not sure if it fully closes it though

I also added a TODO in anticipation of #219

***
<a href="https://pr-preview.s3.amazonaws.com/johannhof/digital-credentials/pull/283.html" title="Last updated on Jun 30, 2025, 5:26 PM UTC (73ae124)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/283/866310c...johannhof:73ae124.html" title="Last updated on Jun 30, 2025, 5:26 PM UTC (73ae124)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1566,20 +1566,19 @@
           users.
         </p>
         <p>
-          This attack might be harder for third-party attackers
-          (such as scripts embedded on the verifier's pages, but not actively
-          collaborating with them for the purpose of tracking) because
-          response encryption is mandatory and responses should be decrypted
-          on the verifier's server. There may be some decrypted results 
-          not reflected back to client-side JavaScript.
+          This attack might be harder for third-party attackers (such as
+          scripts embedded on the verifier's pages, but not actively
+          collaborating with them for the purpose of tracking) because response
+          encryption is mandatory and responses should be decrypted on the
+          verifier's server. There may be some decrypted results not reflected
+          back to client-side JavaScript.
         </p>
-        <p>  
-          To further help users understand
-          the full consequences of sharing credentials with with verifiers,
-          user agents could add additional context to their permission prompts,
-          for example informing users who recently cleared their cookies, or
-          are using a "private browsing" mode, that credentials can be used to
-          reidentify them on the page.
+        <p>
+          To further help users understand the full consequences of sharing
+          credentials with verifiers, user agents could add additional context
+          to their permission prompts, for example informing users who recently
+          cleared their cookies, or are using a "private browsing" mode, that
+          credentials can be used to reidentify them on the page.
         </p>
         <h4 id="leaking-incidental-data">
           Leaking incidental data with credential presentations
--- a/index.html
+++ b/index.html
@@ -1555,15 +1555,14 @@
         <p>
           While the API ensures that no user data is ever shared without a
           permission prompt, the longevity and uniqueness of real world
-          identifiers that are likely to be returned by the Digital
-          Credentials API make it a potential target for trackers and
-          fingerprinters.
+          identifiers that are likely to be returned by the Digital Credentials
+          API make it a potential target for trackers and fingerprinters.
         </p>
         <p>
           Even with selective disclosure, attackers might combine data from a
-          digital credential (such as the user's age, or the credential issuer, timestamps,
-          see [[[#leaking-incidental-data]]]) to reidentify and/or fingerprint
-          users.
+          digital credential (such as the user's age, or the credential issuer,
+          timestamps, see [[[#leaking-incidental-data]]]) to reidentify and/or
+          fingerprint users.
         </p>
         <p>
           This attack might be harder for third-party attackers (such as
--- a/index.html
+++ b/index.html
@@ -1555,7 +1555,7 @@
         <p>
           While the API ensures that no user data is ever shared without a
           permission prompt, the longevity and uniqueness of real world
-          identifiers that are likely to be presented through the Digital
+          identifiers that are likely to be returned by the Digital
           Credentials API make it a potential target for trackers and
           fingerprinters.
         </p>
--- a/index.html
+++ b/index.html
@@ -1547,6 +1547,47 @@
       </section>
       <section>
         <h3>
+          Fingerprinting and Data Leakage
+        </h3>
+        <h4>
+          Browser fingerprinting
+        </h4>
+        <p class="issue" data-number="219">
+          The Digital Credentials API exposes information about which
+          credential exchange protocols are supported by the user agent, which
+          has the potential to be used for browser fingerprinting (see
+          [[[fingerprinting-guidance]]]). We need to add privacy considerations
+          for the involved mitigations here.
+        </p>
+        <h4>
+          Avoiding leaks of credential availability
+        </h4>
+        <p>
+          The Digital Credentials API can not make it possible for websites to
+          learn whether a certain credential is available or not without going
+          through a <a href="#user-permission-and-transparency">user permission
+          flow</a>. Revealing the presence of credentials is a risk to user
+          privacy, as the presence of a credential is personal information that
+          the user might not have preferred to share with the site, and, in
+          combination with other signals, could be used to identify the user
+          without their permission. It is also a risk to free expression, as
+          websites might increasingly start to demand the presentation of these
+          credentials from the user in order to access services, excluding
+          individuals who are unwilling or unable to present credentials.
+        </p>
+      </section>
+      <section>
+        <h3>
+          Over Collection of Data
+        </h3>
+        <p class="issue" title="Work in progress">
+          Explain how the API could be used to request more data than necessary
+          for a transaction and how the ecosystem can mitigate that over
+          collection.
+        </p>
+      </section>
+      <section>
+        <h3>
           User Permission and Transparency
         </h3>
         <p class="issue" title="Work in progress"></p>
--- a/index.html
+++ b/index.html
@@ -1562,7 +1562,7 @@
         <p>
           Even with selective disclosure, attackers might combine data from a
           digital credential (such as the user's age, or the credential issuer,
-          see [[[leaking-incidental-data]]]) to reidentify and/or fingerprint
+          see [[[#leaking-incidental-data]]]) to reidentify and/or fingerprint
           users.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -1563,8 +1563,8 @@
           Avoiding leaks of credential availability
         </h4>
         <p>
-          The Digital Credentials API can not make it possible for websites to
-          learn whether a certain credential is available or not without going
+          The Digital Credentials API does not enable sites to
+          learn whether a credential is available without first going
           through a <a href="#user-permission-and-transparency">user permission
           flow</a>. Revealing the presence of credentials is a risk to user
           privacy, as the presence of a credential is personal information that
--- a/index.html
+++ b/index.html
@@ -1573,7 +1573,7 @@
           without their permission. It is also a risk to free expression, as
           websites might increasingly start to demand the presentation of these
           credentials from the user in order to access services, excluding
-          individuals who are unwilling or unable to present credentials.
+          individuals who are unwilling to present credentials.
         </p>
       </section>
       <section>
--- a/index.html
+++ b/index.html
@@ -1561,7 +1561,7 @@
         </p>
         <p>
           Even with selective disclosure, attackers might combine data from a
-          digital credential (such as the user's age, or the credential issuer,
+          digital credential (such as the user's age, or the credential issuer, timestamps,
           see [[[#leaking-incidental-data]]]) to reidentify and/or fingerprint
           users.
         </p>
--- a/index.html
+++ b/index.html
@@ -1553,20 +1553,21 @@
           Browser fingerprinting
         </h4>
         <p class="issue" data-number="219">
-          The Digital Credentials API exposes information about which
-          credential exchange protocols are supported by the user agent, which
-          has the potential to be used for browser fingerprinting (see
-          [[[fingerprinting-guidance]]]). We need to add privacy considerations
-          for the involved mitigations here.
+          When the Digital Credentials API exposes information about which
+          credential exchange protocols are supported by the user agent, it
+          will mitigate browser fingerprinting through not customizing its
+          response based on e.g. installed wallets. We need to add privacy
+          considerations for the involved risks and mitigations of that
+          functionality here (see [[[fingerprinting-guidance]]]).
         </p>
         <h4>
           Avoiding leaks of credential availability
         </h4>
         <p>
-          The Digital Credentials API does not enable sites to
-          learn whether a credential is available without first going
-          through a <a href="#user-permission-and-transparency">user permission
-          flow</a>. Revealing the presence of credentials is a risk to user
+          The Digital Credentials API does not enable sites to learn whether a
+          credential is available without first going through a <a href=
+          "#user-permission-and-transparency">user permission flow</a>.
+          Revealing the presence of credentials would be a risk to user
           privacy, as the presence of a credential is personal information that
           the user might not have preferred to share with the site, and, in
           combination with other signals, could be used to identify the user
--- a/index.html
+++ b/index.html
@@ -1556,7 +1556,7 @@
           While the API ensures that no user data is ever shared without a
           permission prompt, the longevity and uniqueness of real world
           identifiers that are likely to be presented through the Digital
-          Credentials API makes it a potential target for trackers and
+          Credentials API make it a potential target for trackers and
           fingerprinters.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -1569,15 +1569,9 @@
           scripts embedded on the verifier's pages, but not actively
           collaborating with them for the purpose of tracking) because response
           encryption is mandatory and responses should be decrypted on the
-          verifier's server. There may be some decrypted results not reflected
-          back to client-side JavaScript.
-        </p>
-        <p>
-          To further help users understand the full consequences of sharing
-          credentials with verifiers, user agents could add additional context
-          to their permission prompts, for example informing users who recently
-          cleared their cookies, or are using a "private browsing" mode, that
-          credentials can be used to reidentify them on the page.
+          verifier's server. The verifier could thus ensure not to reflect back
+          decrypted information back to client-side JavaScript. Not all
+          verifiers will choose to do so, however.
         </p>
         <h4 id="leaking-incidental-data">
           Leaking incidental data with credential presentations
@@ -1602,15 +1596,17 @@
           verifiers.
         </p>
         <h4>
-          Revealing device properties through feature and protocol availability
+          Revealing device properties through protocol availability
         </h4>
-        <p class="issue" data-number="219">
-          When the Digital Credentials API exposes information about which
-          credential exchange protocols are supported by the user agent, it
-          will mitigate browser fingerprinting through not customizing its
-          response based on e.g. installed wallets. We need to add privacy
-          considerations for the involved risks and mitigations of that
-          functionality here (see [[[fingerprinting-guidance]]]).
+        <p>
+          The Digital Credentials API exposes information about which
+          credential exchange protocols are supported by the user agent through
+          {{DigitalCredential/userAgentAllowsProtocol()}}. It mitigates browser
+          fingerprinting and revealing information about the user's device
+          configuration through not customizing its response based on e.g.
+          which wallet applications are installed on a user's device. The
+          returned information is thus, at best, equivalent to a user agent
+          version.
         </p>
         <h4>
           Avoiding leaks of credential availability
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1129,6 +1129,37 @@ <h3>
...
+          for the involved mitigations here.
+        </p>
+        <h4>
+          Avoiding leaks of credential availability
+        </h4>
+        <p>
+          The Digital Credentials API can not make it possible for websites to
+          learn whether a certain credential is available or not without going
+          through a <a href="#user-permission-and-transparency">user permission
+          flow</a>. Revealing the presence of credentials is a risk to user
```
```suggestion
          The Digital Credentials API should not enable sites to
          learn whether a credential is available without first going
          through a <a href="#user-permission-and-transparency">user permission
          flow</a>. Revealing the presence of credentials is a risk to user
```

index.html
```diff
@@ -1129,6 +1129,37 @@ <h3>
...
+          privacy, as the presence of a credential is personal information that
+          the user might not have preferred to share with the site, and, in
+          combination with other signals, could be used to identify the user
+          without their permission. It is also a risk to free expression, as
+          websites might increasingly start to demand the presentation of these
+          credentials from the user in order to access services, excluding
+          individuals who are unwilling or unable to present credentials.
```
```suggestion
          individuals who are unwilling to present credentials.
```

While excluding those unable to present credentials is indeed a concern, the risk that comes from sites knowing that a credential is available is that the site knows a credential is available and will exclude those who are unwilling to share it.

npdoty:
There are additional fingerprinting and tracking risks when credentials are affirmatively presented: 

- characteristics of credentials (even ones not inherently linkable) can become part of a user fingerprint
- presented credentials may include unique cross-origin identifiers otherwise unrelated to the meaningful contents of the credential (for example, a device key used to sign presentations)

Browsers that wish to detect and protect against fingerprinting would need to know what kind of presentations are being requested in order to inform users or to help them understand the implications of sharing or to help them clear the state mechanisms being used to recognize them. (Wallets are poorly placed to address this issue, since it affects the privacy protections that browsers can provide against tracking users in browsing to multiple sites.)

npdoty (changes requested):
Small edits on the existing wording.

Fingerprinting considerations should also consider the implications of issuing and presenting credentials.

index.html
```diff
@@ -1129,6 +1129,37 @@ <h3>
...
       </section>
+      <section>
+        <h3>
+          Fingerprinting and Data Leakage
+        </h3>
+        <h4>
+          Browser fingerprinting
+        </h4>
+        <p class="issue" data-number="219">
+          The Digital Credentials API exposes information about which
```
marcoscaceres, line 1602: I disagree that this is a fingerprinting vector. The response to this is just if the User Agent knows the string, and not if any wallet supports it. As such, the mitigation has already been taken care of: it's not linkable to any wallet and can't be used to check if a wallet is actually supported or on the user's system... just that the API might forward the request to the possible wallets, but it knows and says nothing about the available wallets.

index.html
```diff
@@ -1129,6 +1129,37 @@ <h3>
...
+          credential exchange protocols are supported by the user agent, which
+          has the potential to be used for browser fingerprinting (see
+          [[[fingerprinting-guidance]]]). We need to add privacy considerations
+          for the involved mitigations here.
+        </p>
+        <h4>
+          Avoiding leaks of credential availability
+        </h4>
+        <p>
+          The Digital Credentials API can not make it possible for websites to
```
```suggestion
          The Digital Credentials API does not make it possible for websites to
```

index.html
marcoscaceres: These need to be written as statements of fact, which is codified in the design of the API (in the algorithms). 

As written, this should be a bug report. 

We can keep this text as rationale of *why the API doesn't allow this*, but not as guidance for "the API should not do this" (which would be a bug if it did). 

Make sense?

marcoscaceres (changes requested):
This should be written as statements of fact, that the API doesn't allow these things by design.

index.html
npdoty:
```suggestion
          The Digital Credentials API does not enable sites to
          learn whether a credential is available without first going
          through a <a href="#user-permission-and-transparency">user permission
          flow</a>. Revealing the presence of credentials is a risk to user
```

Reworded suggestion, per @marcoscaceres, to describe the intended state and reasoning, rather than using language that could be misread as introducing a normative requirement (which will be a normative requirement elsewhere, I believe).

johannhof: Thanks, I agree with Marcos and those seem like the right words now :)

index.html
johannhof: Hah, good catch, that wouldn't make sense.

index.html
johannhof: Right, that is the mitigation. I agree with you it's not a fingerprinting vector, but that's because we're smart and hopefully doing it correctly. I think it's fair to call out that there was a fingerprinting vector here that is being mitigated, this helps set context for why the API works that way vs. revealing which protocols are actually supported by wallets.

I can reword the inline issue text to make that more clear :)

npdoty:
> > There are additional fingerprinting and tracking risks when credentials are affirmatively presented:
> > 
> > * characteristics of credentials (even ones not inherently linkable) can become part of a user fingerprint
> 
> I'm not sure about this one - yes, what you're saying is technically correct, but it does not match the definition of "fingerprinting" we commonly use. The location returned by the Geolocation API, or the pixels returned by the Camera API, could also be turned into a fingerprint if the user accepts these prompts everywhere.

I see what you mean, but we don't have a bright line distinction between which characteristics are part of fingerprinting and which are just directly revealed information that can be used to identify someone elsewhere. Our [definition of fingerprinting surface](https://w3c.github.io/fingerprinting-guidance/#dfn-fingerprinting-surface) explicitly includes environmental characteristics like sensor readings including location information. The relevant part to me seemed to be that you might reveal some small observable characteristic (like your age, or issuer of your age credential) that isn't inherently identifying, but could be used in combination with a mix of other characteristics (IP address, UA string, etc.) to identify or re-identify the user in a way that's opaque and hard to protect against without systematic responses, like browser fingerprinting.

> > * presented credentials may include unique cross-origin identifiers otherwise unrelated to the meaningful contents of the credential (for example, a device key used to sign presentations)
> 
> This seems good to add, thanks!
> 
> > Browsers that wish to detect and protect against fingerprinting would need to know what kind of presentations are being requested in order to inform users or to help them understand the implications of sharing or to help them clear the state mechanisms being used to recognize them. (Wallets are poorly placed to address this issue, since it affects the privacy protections that browsers can provide against tracking users in browsing to multiple sites.)
> 
> > Issuance of credentials also implicates fingerprinting in a cookie-like way. Issuing a credential so that it can be requested later is a state mechanism that can be used to track users, even after they clear their local state, even if there's nothing personally identifying in the credential content itself.
> 
> I'm not fully understanding this - are you saying that credentials could be requested without a prompt after local state has been cleared?

Credentials that are issued by a site in a way that would typically indicate less risk (not government-issued or high-assurance, not including personally-identifying or sensitive information, maybe even limited to the same origin), could, without the user realizing, include cookie-like identifiers. User permission prompts would be a helpful mitigation, especially if paired with some notice that this could be used to re-identify you if you've previously cleared cookies/state. Users who log back in have a clearer model of re-identification. Users who log out and clear all state and then still show a discount/fan card might not realize they're re-spawning an identifier.

johannhof:
@npdoty I think that makes sense to me, I tried to capture it in the text... LMK what you think :)

index.html
```diff
@@ -1128,6 +1128,87 @@ <h3>
...
+          This attack is made significantly harder for third-party attackers
+          (such as scripts embedded on the verifier's pages, but not actively
+          collaborating with them for the purpose of tracking) by the fact that
+          response encryption is mandatory. To further help users understand
+          the full consequences of sharing credentials with with verifiers,
+          user agents could add additional context to their permission prompts,
+          for example informing users who recently cleared their cookies, or
+          are using a "private browsing" mode, that credentials can be used to
+          reidentify them on the page.
+        </p>
```
npdoty: This paragraph seems like two very distinct points.

I would be more cautious/less defensive about how protective it is to encrypt results to the verifier.

```suggestion
        <p>
          This attack might be harder for third-party attackers
          (such as scripts embedded on the verifier's pages, but not actively
          collaborating with them for the purpose of tracking) because
          response encryption is mandatory and responses should be decrypted
          on the verifier's server. There may be some decrypted results 
          not reflected back to client-side JavaScript.
        </p>
        <p>  
          To further help users understand
          the full consequences of sharing credentials with with verifiers,
          user agents could add additional context to their permission prompts,
          for example informing users who recently cleared their cookies, or
          are using a "private browsing" mode, that credentials can be used to
          reidentify them on the page.
        </p>
```

index.html
```diff
@@ -1128,6 +1128,87 @@ <h3>
...
+          Fingerprinting and Data Leakage
+        </h3>
+        <h4>
+          Browser fingerprinting
+        </h4>
+        <p>
+          While the API ensures that no user data is ever shared without a
+          permission prompt, the longevity and uniqueness of real world
+          identifiers that are likely to be presented through the Digital
+          Credentials API makes it a potential target for trackers and
```
```suggestion
          Credentials API make it a potential target for trackers and
```

(Longevity and uniqueness make the API a potential target....)

npdoty:
thanks @johannhof I think that captures a lot more of it. we can walk through the fingerprinting-guidance in more detail in the future, but this should get it started. suggested a couple changes.

index.html
johannhof: Thanks, seems both fair

johannhof:
Thanks for the help Nick!

@marcoscaceres this is ready for your review then :)

index.html
```diff
@@ -1139,6 +1139,91 @@ <h3>
...
+        <h3>
+          Fingerprinting and Data Leakage
+        </h3>
+        <h4>
+          Browser fingerprinting
+        </h4>
+        <p>
+          While the API ensures that no user data is ever shared without a
+          permission prompt, the longevity and uniqueness of real world
+          identifiers that are likely to be presented through the Digital
```
```suggestion
          identifiers that are likely to be returned by the Digital
```

index.html
```diff
@@ -1139,6 +1139,91 @@ <h3>
...
+          Credentials API make it a potential target for trackers and
+          fingerprinters.
+        </p>
+        <p>
+          Even with selective disclosure, attackers might combine data from a
+          digital credential (such as the user's age, or the credential issuer,
```
marcoscaceres, line 1563:
```suggestion
          digital credential (such as the user's age, or the credential issuer, timestamps,
```

index.html
```diff
@@ -1139,6 +1139,91 @@ <h3>
...
+          see [[[#leaking-incidental-data]]]) to reidentify and/or fingerprint
+          users.
+        </p>
+        <p>
+          This attack might be harder for third-party attackers (such as
+          scripts embedded on the verifier's pages, but not actively
+          collaborating with them for the purpose of tracking) because response
+          encryption is mandatory and responses should be decrypted on the
+          verifier's server. There may be some decrypted results not reflected
```
marcoscaceres: If this is guidance, then verifiers should be told to be on the lookup for things they put into the page that would allow for traceability of users if they got XSS'ed. The other problem is that they can share the decrypted data with third parties once decrypted... and users might not understand that can happen.

index.html
```diff
@@ -1139,6 +1139,91 @@ <h3>
...
+          back to client-side JavaScript.
+        </p>
+        <p>
+          To further help users understand the full consequences of sharing
+          credentials with verifiers, user agents could add additional context
+          to their permission prompts, for example informing users who recently
+          cleared their cookies, or are using a "private browsing" mode, that
+          credentials can be used to reidentify them on the page.
```
marcoscaceres: I don't think anyone is doing this, but sure 🤷. My suggestion would be to fix the protocols that are leaking the identifiers in the first place, and that each returned property be evaluated. 

The problem is at the protocol/format level, in that it allows the tracking vectors in the first place. A lot of the time, certain IDs might be needed (e.g., the actual drivers license number)... but in other cases, it might be sufficient to just say "credential was created on date" without an insane amount of precision of including the microseconds.

index.html
```diff
@@ -1139,6 +1139,91 @@ <h3>
...
+        <p>
+          While the Digital Credentials API does not control the content of a
+          credential response, user agents can help protect users against this
+          type of tracking through clearly highlighting which information
+          likely gets shared with the verifier beyond what was requested, and,
+          more broadly, by identifying and blocking abuse of the API by
+          verifiers.
+        </p>
+        <h4>
+          Revealing device properties through feature and protocol availability
```
marcoscaceres: I still don't think this should be here... not as written: "protocol availability" is not a fingerprinting concern as currently being specified.

index.html
```diff
@@ -1139,6 +1139,91 @@ <h3>
...
+          "#user-permission-and-transparency">user permission flow</a>.
+          Revealing the presence of credentials would be a risk to user
+          privacy, as the presence of a credential is personal information that
+          the user might not have preferred to share with the site, and, in
+          combination with other signals, could be used to identify the user
+          without their permission. It is also a risk to free expression, as
+          websites might increasingly start to demand the presentation of these
+          credentials from the user in order to access services, excluding
+          individuals who are unwilling to present credentials.
+        </p>
```
marcoscaceres, line 1626:
```suggestion
        <p>
          Revealing the presence of credentials would be a risk to user
          privacy, as the presence of a credential is personal information that
          the user might not have preferred to share with the site, and, in
          combination with other signals, could be used to identify the user
          without their permission. Because of this, the Digital Credentials API
          does not enable sites to learn whether a
          credential is available without first going through a <a href=
          "#user-permission-and-transparency">user permission flow</a>.
          </p>
          <h4>Risk to free expression and anonymity</h4>
          <p>
          The Digital Credentials API poses a risk to free expression and anonymity online. 
          This is because websites, or regulatory bodies, could increasingly start to demand the presentation of these
          credentials from the user in order to access services, excluding
          individuals who are unwilling to present credentials.
        </p>
```

marcoscaceres (changes requested):
Some suggestions... I think the free expression issue should be called out separately.

index.html
npdoty: I think #262 on unnecessary use of credentials is intended to be the place for longer discussion of coercive use of credentials to limit free expression or anonymous access. It came up here because protecting against disclosure of presence of a credential is one partial way to mitigate that kind of coercion. 

I have no strong feelings on whether we describe that mitigation here, or link to the other section, or both.

index.html
johannhof: Hence the "not actively collaborating" bit

johannhof: But I don't feel strongly about this block, if you'd prefer to remove this it's fine with me

index.html
johannhof: Yeah, this particular bit about clearing cookies seems a little far-fetched to me, too. It's based on a suggestion by @npdoty, so I'll let him speak to it. I will say that theoretically even moderately coarse values such as "date credential was created on" can be combined with other signals such as the issuing authority to reidentify users.

index.html
johannhof: I agree, it's not a fingerprinting concern, because it's mitigated by the spec requirements. That seems worth calling out. Would you suggest we remove other sections from the privacy considerations once they get mitigated?

index.html
johannhof: Hmm yeah I'm not sure I'd add a separate sub-section for that here. It's under credential availability leakage because that specifically could be abused to demand a credential from the user if it's theoretically available.

So, I'd prefer to remove this paragraph than move it into its own subsection which doesn't make sense when we have #262

index.html
npdoty: I don't know why this threat would be dismissed as far-fetched; I think our experience on the Web has been that re-identification by combining lots of attributes is an extremely widespread practice and that attackers are happy to use all available details and add new details as they become more widely available. Date created would be a huge amount of entropy! Just issuing authority is a lot, especially when combined with other fingerprinting surface.

I think it's more of a risk if not-inherently-identifying attributes (like age) become commonly requested on many sites (as has been proposed by many legislators).

johannhof: To explain what I believe is far-fetched: I think we should absolutely call out tracking / reidentification risks, and I agree (as said above) that date issued / issuing authority might be sufficient for reidentification, especially combined with other signals. I'm not sure the specific scenario of "a tracker on a website (or the website itself?) relies on the user's RWI to restore a tracking cookie after it was deleted" is particularly realistic, it just doesn't scale, even with massive adoption of DCs.

index.html
```diff
@@ -1205,6 +1205,90 @@ <h3>
...
+          credential response, user agents can help protect users against this
+          type of tracking through clearly highlighting which information
+          likely gets shared with the verifier beyond what was requested, and,
+          more broadly, by identifying and blocking abuse of the API by
+          verifiers.
+        </p>
+        <h4>
+          Revealing device properties through feature and protocol availability
+        </h4>
+        <p class="issue" data-number="219">
```
johannhof: Note from @marcoscaceres: Let's get this written up in this PR as #221 will merge soon.

index.html
johannhof: We'll resolve this by adding a disclaimer along the lines of what Marcos mentions above

index.html
marcoscaceres: Ok, discussed with @johannhof... let's leave it as it.

index.html
johannhof: Discussed this with Marcos: We aligned that this specific suggestion is unlikely to be considered by user agents in practice, but as mentioned the concern is valid, so I'll remove this last part but leave the description of the problem in place.

index.html
```diff
@@ -1540,6 +1540,86 @@ <h5>
...
+          fingerprint users.
+        </p>
+        <p>
+          This attack might be harder for third-party attackers (such as
+          scripts embedded on the verifier's pages, but not actively
+          collaborating with them for the purpose of tracking) because response
+          encryption is mandatory and responses should be decrypted on the
+          verifier's server. The verifier could thus ensure not to reflect back
+          decrypted information back to client-side JavaScript. Not all
+          verifiers will choose to do so, however.
```
npdoty, line 1574: Asking websites to protect against fingerprinting would be a new approach, but not one likely to be effective. I expect that almost every verifier would reflect at least some of the data back to the site, in order to show the information collected, confirm it with the user, present some action or conclusion based on that data, etc. So it's not just that this protection might not work, but more that I don't think we expect that it will almost ever work.

index.html
npdoty: Unfortunately I do agree that browsers who implement this API (at least initially) are unlikely to adopt optional transparency mechanisms to help their users understand the tracking implications of overuse of credentials. I don't think we're going to figure it out immediately, but I've opened #302 to track potential normative mitigations.

And I agree that we should fix the protocols that introduce this unnecessary additional privacy-harmful information, or come up with API designs that provide only the necessary information. It's harmful that these protocols are being more widely deployed today despite these problems. We can raise that during privacy reviews of exchange protocols for the registry (if that is in fact a meaningful requirement #255 #226 #288) and with those groups separately.

index.html
simoneonofri: @npdoty

It may seem like a strange request :) 

But, for example, if encryption and signatures are required on the protocol side, we can prevent a malicious extension from seeing the data, which is then reflected by the site and can be viewed.

It's similar to when we say in an input type=password not to reflect the password, for example, when you enter it incorrectly (something I used to encounter quite often years ago).

It certainly needs to be better understood at the UX level, but for example, not reflecting the data immediately but in a summary screen increases the complexity of exploitation.

I think it's worth pointing out (perhaps not as a MUST).

johannhof: Honestly I'm happy to remove this entire block if it's not convincing enough. Maybe @marcoscaceres wants to make the deciding vote on this?

wseltzer (approved):

johannhof:
@marcoscaceres friendly ping on another quick look at this  :)

index.html
mohamedamir: My 5 cents are: yes, 3rd party attackers can always read the data reflected in the website, but this data is as good as any data entered in a form.

Encryption protects against reading the cryptographically verified the data. It is the website's responsibility to keep this safe and not reflect it in plain text (not without obfuscation for example)

index.html
```diff
@@ -1545,6 +1545,86 @@ <h5>
...
+          This additional information could be used to reidentify and
+          fingerprint users, which is especially relevant when an otherwise
+          unlinkable presentation is made.
+        </p>
+        <p>
+          While the Digital Credentials API does not control the content of a
+          credential response, user agents can help protect users against this
+          type of tracking through clearly highlighting which information
+          likely gets shared with the verifier beyond what was requested, and,
+          more broadly, by identifying and blocking abuse of the API by
```
mohamedamir: We debated about this before, but how to define "abuse"?

mohamedamir (approved):
Thank Johann!
This LGTM with non-blocking comments!

index.html
johannhof: Ok, let's keep the text for now, I don't think it's factually wrong even if it might be unrealistic to hope that verifiers broadly adhere to best practices here.

index.html
johannhof: I guess this is why I qualified it with "more broadly" here - but I can replace "abuse of the API" with "fingerprinting through the API". The point I'm trying to make is that most browsers at this point have more general efforts to detect fingerprinting and that this API could be considered in those.

PR #287, opened 2025-06-19
Add additional issues to privacy considerations

johannhof:
This adds a few additional inline issues corresponding to open questions to the privacy considerations.


index.html
```diff
@@ -967,6 +967,9 @@ <h3>
         <h4>
           Exchange Protocol Considerations for User Privacy
         </h4>
+        <p class="issue" data-number="255">
+          Actually write these in the protocol requirements
+        </p>
```
marcoscaceres: If you do the following, the aside gets filled with the text from the issue on GitHub IIRC. 

```suggestion
        <aside class="issue" data-number="255"></aside>
```

index.html
```diff
@@ -1005,16 +1008,16 @@ <h5>
...
         <p>
-          Through the Digital Credentials API, the user agent can help verifiers
-          and wallets exchange unlinkable attributes, but it can not guarantee
-          that no linkable information is passed between verifiers and wallets.
-          It is recommended that user agents account for this fact in their
-          user permission experience.
+          Through the Digital Credentials API, the user agent can help
+          verifiers and wallets exchange unlinkable attributes, but it can not
+          guarantee that no linkable information is passed between verifiers
+          and wallets. It is recommended that user agents account for this fact
```
marcoscaceres: It's not the user agent that presents this information: it's the clients/wallets (at least on Apple's platforms).

index.html
johannhof: Yeah, I think it's okay to do it in this instance but looking at how it gets rendered it displays the full text of the issue which can get quite long, so we probably shouldn't do it everywhere 😄

index.html
johannhof: So, this is just reformatting noise from tidy (I'll rebase on main to fix it)

I'm also not 100% sure I understand your feedback on this particular point, is there anything you'd like to change? Happy to follow up in a separate PR if needed.

npdoty (approved):
Looks good to me, moving around issue blocks to more appropriate places (not losing any) and adding one.

timcappalli (approved):

PR #354, opened 2025-08-21
chore(explainer): fix link to risk.md from ping

marcoscaceres:
@npdoty, the original link to this is now a 404, but I'm sure what to link this too? Can you help me find the right place to link to. 

(Asking for the TAG)

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -46,7 +46,7 @@ Please see the  [requesting a digital credential across origins](https://www.w3.
 
 There are many alternatives that were considered, most notably:
 
-- [Do nothing](https://github.com/w3cping/credential-considerations/blob/main/risks.md#the-consequences-of-a-failure-to-act-are-as-valid-as-those-of-acting) (the intentional status quo in browsers for the past several years)
+- [Do nothing](https://github.com/w3c/credential-considerations/blob/main/credentials-considerations.md#risks) (the intentional status quo in browsers for the past several years)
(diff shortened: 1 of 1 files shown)


npdoty:
I don't think "Do nothing" is a descriptive link title for the credential considerations document, which doesn't advocate for doing nothing and doesn't describe the considerations of doing nothing.

I think maybe the original interest was in linking to a point that @RByers had made about the potential costs of not implementing an API. In the history, that's available here:
https://github.com/w3c/credential-considerations/blob/a335e21ebea9c09be2f09562811b42cc1f9a6966/risks.md#the-consequences-of-a-failure-to-act-are-as-valid-as-those-of-acting
I don't think we have that in the wider merged document at the moment. Rick, maybe you have a proposal for that, or another place to make that point that these editors can link to?

marcoscaceres:
@simoneonofri, would you mind doing what you suggest as a followup?

npdoty:
I'm confused why this PR was merged. I explained both that the link title was misleading, and that the link did not go to the content that was the intended reference.

npdoty:
It seemed like I was specifically asked to review and provide feedback on whether this was the right link. I checked and answered and included others who could help correct the link to the content that the authors were originally referring to and then the PR was merged anyway. This felt pretty confusing to me, for what it's worth.

Follow-up issue tracked here: https://github.com/w3c-fedid/digital-credentials/issues/363

PR #306, opened 2025-06-30
Define Coordinator and Client

marcoscaceres:
Closes #305

PR #381, opened 2025-09-24
User Agent Automation and WebDriver test support

### LINKED ISSUE #357: Add means to test Digital Credentials API
Currently, the web-platform-tests for the Digital Credentials (DC) API are limited to testing various error conditions, such as malformed requests or multiple concurrent requests. This leaves significant gaps in our testing capabilities.

For example,, we are unable to test:

1. The end-to-end **successful flow** (the "happy path") of a user being issued and presenting a credential.
2. User interactions with the selector UI, such as aborting the operation while the dialog is active.

To address this, we need a testing mechanism that can simulate a complete wallet interaction.

I would like to propose two potential alternatives for discussion:

**Alternative 1: Extend WebDriver with Virtual Wallets**
We could follow a similar approach to the WebAuthn specification, which introduced the `virtual_authenticator` interface to `WebDriver`. This approach would involve creating a **virtual wallet** testing API.

The virtual wallet would be configurable through WebDriver to operate under a specific issuance and presentation protocol (e.g. openid4vp). It would need to fully support `get()` and `create()` methods, allowing tests to simulate the entire credential lifecycle.

**Pros:**

- Provides a powerful and reusable test API that web developers could also leverage for their own application testing.
- Aligns with existing testing patterns on the web platform. And hence we won't have to implement things from scratch

**Cons:**
- The implementation could be complex, particularly regarding the correct handling of cryptographic protocols required to encrypt and format responses.

**Alternative 2: Introduce an Action-Based Testing Protocol**
A lighter-weight alternative would be to introduce a simple, action-based protocol specifically for testing the API's interaction with the wallet selector UI. This could be similar to geolocation setting testing APIs.

This protocol would allow tests to issue commands to the wallet UI to simulate user behavior. Example commands could include:

- `accept_next_request(credential_data)`: Instructs the wallet to accept the next API call and respond with the specified credential.
- `reject_next_request(error_reason)`: Rejects the next API call.
- `hold_next_request()`: Keeps the UI open without responding, allowing for testing of timeout logic.
- `abort_after(timeout)`: Simulates the user closing the UI after a specified duration.

**Pros:**
- Simpler to implement as it focuses on behavior rather than complex cryptographic protocols.
- Directly addresses the need to test UI interactions and different user choices.

**Cons:**
- May be less realistic than a full virtual wallet implementation.
- Might not be as useful for web developers' own testing needs.

mohamedamir:
Closes #357

- [X] Modified Web platform tests https://github.com/web-platform-tests/wpt/pull/58587

Implementation commitment:

- [x] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=306292)
- [X] [Chromium](https://crbug.com/447093830) 
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/381.html" title="Last updated on May 11, 2026, 5:09 PM UTC (5c51691)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/381/9e4579c...5c51691.html" title="Last updated on May 11, 2026, 5:09 PM UTC (5c51691)">Diff</a>


index.html
```diff
@@ -856,6 +856,22 @@ <h3>
...
+          `data`.
+          </li>
+          <li>If the mode is "wait", the promise never resolves.
+          </li>
+          <li>If the mode is "decline", reject the promise with a
+          "NotAllowedError" `DOMException`.
+          </li>
+        </ol>
+      </li>
+      <li>Otherwise (if [=next wallet behavior=] is not set), proceed with the
```
```suggestion
      <li>Otherwise (i.e., if [=next wallet behavior=] is not set), proceed with the
```

TallTed: or
```suggestion
      <li>Otherwise (that is, if [=next wallet behavior=] is not set), proceed with the
```

index.html
```diff
@@ -856,6 +856,22 @@ <h3>
           </li>
         </ol>
       </li>
+      <li>If the session's [=next wallet behavior=] is set:
+        <ol>
+          <li>If the mode is "respond", resolve the promise with a new
```
marcoscaceres: Nit: let's link these up properly to their definitions (e.g. {{DigitalCredential}})... "respond" should be formally defined, etc.

index.html
```diff
@@ -856,6 +856,22 @@ <h3>
...
+          `DigitalCredential` object whose `data` member is set to the stored
+          `data`.
+          </li>
+          <li>If the mode is "wait", the promise never resolves.
+          </li>
+          <li>If the mode is "decline", reject the promise with a
+          "NotAllowedError" `DOMException`.
```
```suggestion
          {{"NotAllowedError"}} {{DOMException}}.
```

index.html
```diff
@@ -917,6 +933,22 @@ <h3>
           </li>
         </ol>
       </li>
+      <li>If the session's [=next wallet behavior=] is set:
```
marcoscaceres: As this is duplicative, maybe move the step to its own algorithm? then we only need to maintain one set of steps.

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+              </td>
+              <td>
+                Indicates whether the endpoint node supports all Virtual Wallet
+                commands.
+              </td>
+            </tr>
+          </tbody>
+        </table>
+        <p>
+          When <a data-cite="webdriver#dfn-validate-capabilities">validating
```
marcoscaceres: nit: avoid using data-cite... if something hasn't been exported from Web Driver, we should send a PR to web driver to export the definition we need.

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+          capabilities</a>, the extension-specific substeps to validate
+          <code>digital-credentials:virtualWallet</code> with
+          <code>value</code> are the following:
```
marcoscaceres: Nit.. use shorthand: 

```suggestion
          `value` are the following:
```

Also in other places

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+      <section id="virtual-wallet-commands">
+        <h3>
+          Virtual Wallet Commands
+        </h3>
+        <p>
+          The following commands allow controlling the behavior of a virtual
+          wallet for testing purposes. These commands affect the <dfn>next
+          wallet behavior</dfn> of the next
+          <code>navigator.credentials.get({digital: ...})</code> or
+          <code>navigator.credentials.create({digital: ...})</code> call.
```
marcoscaceres: These should probably link to the algorithms, not to the methods themselves.

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+                <td>
+                  <code>/session/{session
+                  id}/digital-credentials/response</code>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+          <p>
+            The parameters object is a JSON <a data-cite=
+            "webdriver#dfn-object">Object</a> with the following key:
```
marcoscaceres: as above... we should get "object" exported from Web Driver.

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+                  The data to be returned by the wallet.
+                </td>
+              </tr>
+            </tbody>
+          </table>
+          <p>
+            The [=remote end steps=] are:
+          </p>
+          <ol class="algorithm">
+            <li>If <code>parameters</code> is not a JSON <a data-cite=
```
marcoscaceres: Are JSON object and Web Driver objects different things? 

Also, I wonder if this should really be framed in Web Driver 2 terminology, which I think is CBOR. 

We should definitely have the Web Driver folks check all this before we merge it too.

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+            <li>Set the session's [=next wallet behavior=] to <code>{ "mode":
+            "wait" }</code>.
+            </li>
+            <li>Return [=success=].
+            </li>
+          </ol>
+        </section>
+        <section id="set-wallet-to-decline">
+          <h4>
+            <dfn>Set Wallet to Decline</dfn>
```
marcoscaceres: Maybe we should say "abort" here to keep terminology consistent? (i.e., affectively this become an AbortError, right)?

index.html
```diff
@@ -856,6 +856,22 @@ <h3>
           </li>
         </ol>
       </li>
+      <li>If the session's [=next wallet behavior=] is set:
```
marcoscaceres: I wonder if we need this these steps at all?... this should really be covered by #306 with the behavior defined in the automated section.

marcoscaceres (changes requested):
I like where this is going, but I'm not loving the way the model integrates with the algorithms. 

What we should have is basically: 

 * There is (or are) virtual wallet(s).
 * Each virtual wallets is in some state.
 * Virtual wallet's state can be changed. 

Virtual wallets take precedence over platform [=holders=] when in automation.

index.html
timcappalli:
```suggestion
      <li>If the session's [=next credential manager behavior=] is set:
```

index.html
timcappalli:
```suggestion
      <li>Otherwise (if [=next credential manager behavior=] is not set), proceed with the
```

index.html
timcappalli:
```suggestion
      <li>If the session's [=next credential manager behavior=] is set:
```

index.html
```diff
@@ -917,6 +933,22 @@ <h3>
...
+          `data`.
+          </li>
+          <li>If the mode is "wait", the promise never resolves.
+          </li>
+          <li>If the mode is "decline", reject the promise with a
+          "NotAllowedError" `DOMException`.
+          </li>
+        </ol>
+      </li>
+      <li>Otherwise (if [=next wallet behavior=] is not set), proceed with the
```
timcappalli:
```suggestion
      <li>Otherwise (if [=next credential manager behavior=] is not set), proceed with the
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+              </th>
+              <th>
+                Description
+              </th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td>
+                Virtual Wallet Support
```
timcappalli:
```suggestion
                Virtual Credential Manager Support
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+              </td>
+              <td>
+                <code>digital-credentials:virtualWallet</code>
```
timcappalli:
```suggestion
                <code>digital-credentials:virtualCredentialManager</code>
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+              </td>
+              <td>
+                [=boolean=]
+              </td>
+              <td>
+                Indicates whether the endpoint node supports all Virtual Wallet
```
timcappalli:
```suggestion
                Indicates whether the endpoint node supports all Virtual Credential Manager
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+                commands.
+              </td>
+            </tr>
+          </tbody>
+        </table>
+        <p>
+          When <a data-cite="webdriver#dfn-validate-capabilities">validating
+          capabilities</a>, the extension-specific substeps to validate
+          <code>digital-credentials:virtualWallet</code> with
```
timcappalli:
```suggestion
          <code>digital-credentials:virtualCredentialManager</code> with
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+          [=invalid argument=].
+          </li>
+          <li>Otherwise, let <code>deserialized</code> be set to
+          <code>value</code>.
+          </li>
+        </ol>
+        <p>
+          When <a data-cite="webdriver#dfn-matching-capabilities">matching
+          capabilities</a>, the extension-specific steps to match
+          <code>digital-credentials:virtualWallet</code> with
```
timcappalli:
```suggestion
          <code>digital-credentials:virtualCredentialManager</code> with
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+          <code>value</code> are the following:
+        </p>
+        <ol class="algorithm">
+          <li>If <code>value</code> is <code>true</code> and the
+            <a data-cite="webdriver#dfn-endpoint-node">endpoint node</a> does
+            not support any of the Virtual Wallet commands, the match is
```
timcappalli:
```suggestion
            not support any of the Virtual Credential Manager commands, the match is
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+        </ol>
+      </section>
+      <section id="virtual-wallet-commands">
+        <h3>
+          Virtual Wallet Commands
+        </h3>
+        <p>
+          The following commands allow controlling the behavior of a virtual
+          wallet for testing purposes. These commands affect the <dfn>next
+          wallet behavior</dfn> of the next
```
timcappalli:
```suggestion
          The following commands allow controlling the behavior of a virtual
          credential manager for testing purposes. These commands affect the <dfn>next
          credential manager behavior</dfn> of the next
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+          <code>navigator.credentials.get({digital: ...})</code> or
+          <code>navigator.credentials.create({digital: ...})</code> call.
+        </p>
+        <section id="set-wallet-response">
+          <h4>
+            <dfn>Set Wallet Response</dfn>
```
timcappalli:
```suggestion
        <section id="set-cm-response">
          <h4>
            <dfn>Set Credential Manager Response</dfn>
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+          </h4>
+          <p>
+            The [=Set Wallet Response=] [[WebDriver]] [=extension command=]
```
timcappalli:
```suggestion
            The [=Set Credential Manager Response=] [[WebDriver]] [=extension command=]
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+            configures the virtual wallet to respond with a specific data
```
timcappalli:
```suggestion
            configures the virtual credential manager to respond with a specific data
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+            <tbody>
+              <tr>
+                <td>
+                  <code>data</code>
+                </td>
+                <td>
+                  string
+                </td>
+                <td>
+                  The data to be returned by the wallet.
```
timcappalli:
```suggestion
                  The data to be returned by the credential manager.
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+            <li>Set the session's [=next wallet behavior=] to <code>{ "mode":
+            "respond", "data": parameters.data }</code>.
+            </li>
+            <li>Return [=success=].
+            </li>
+          </ol>
+        </section>
+        <section id="set-wallet-to-wait">
+          <h4>
+            <dfn>Set Wallet to Wait</dfn>
```
timcappalli:
```suggestion
            <dfn>Set Credential Manager to Wait</dfn>
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+            [=invalid argument=].
+            </li>
+            <li>Set the session's [=next wallet behavior=] to <code>{ "mode":
+            "respond", "data": parameters.data }</code>.
+            </li>
+            <li>Return [=success=].
+            </li>
+          </ol>
+        </section>
+        <section id="set-wallet-to-wait">
```
timcappalli:
```suggestion
        <section id="set-cm-to-wait">
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+          <h4>
+            <dfn>Set Wallet to Wait</dfn>
+          </h4>
+          <p>
+            The [=Set Wallet to Wait=] [[WebDriver]] [=extension command=]
+            configures the virtual wallet to simulate a user waiting
+            indefinitely in the wallet UI.
```
timcappalli:
```suggestion
            The [=Set Credential Manager to Wait=] [[WebDriver]] [=extension command=]
            configures the virtual credential manager to simulate a user waiting
            indefinitely in the credential manager UI.
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+                  id}/digital-credentials/response_wait</code>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+          <p>
+            The [=remote end steps=] are:
+          </p>
+          <ol class="algorithm">
+            <li>Set the session's [=next wallet behavior=] to <code>{ "mode":
```
timcappalli:
```suggestion
            <li>Set the session's [=next credential manager behavior=] to <code>{ "mode":
```

index.html
timcappalli:
```suggestion
        <section id="set-cm-to-decline">
          <h4>
            <dfn>Set Credential Manager to Decline</dfn>
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+          </ol>
+        </section>
+        <section id="set-wallet-to-decline">
+          <h4>
+            <dfn>Set Wallet to Decline</dfn>
+          </h4>
+          <p>
+            The [=Set Wallet to Decline=] [[WebDriver]] [=extension command=]
+            configures the virtual wallet to simulate a user declining the
+            request.
```
timcappalli:
```suggestion
            The [=Set Credential Manager to Decline=] [[WebDriver]] [=extension command=]
            configures the virtual credential manager to simulate a user declining the
            request.
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+                  id}/digital-credentials/response_decline</code>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+          <p>
+            The [=remote end steps=] are:
+          </p>
+          <ol class="algorithm">
+            <li>Set the session's [=next wallet behavior=] to <code>{ "mode":
```
timcappalli:
```suggestion
            <li>Set the session's [=next credential manager behavior=] to <code>{ "mode":
```

index.html
```diff
@@ -2141,6 +2173,268 @@ <h2>
...
+          <li>If <code>value</code> is <code>true</code> and the
+            <a data-cite="webdriver#dfn-endpoint-node">endpoint node</a> does
+            not support any of the Virtual Wallet commands, the match is
+            unsuccessful.
+          </li>
+          <li>Otherwise, the match is successful.
+          </li>
+        </ol>
+      </section>
+      <section id="virtual-wallet-commands">
```
timcappalli:
```suggestion
      <section id="virtual-cm-commands">
```

index.html
marcoscaceres: We haven't really used "wallet" formally... we've stuck to holder... so I wonder if this should be "virtual holder"

index.html
marcoscaceres: Best to avoid acronyms 🙏

TallTed: I don't know how much acronym avoidance matters in fragment identifiers, but it still might be better to abbreviate somewhat less, perhaps as —
```suggestion
      <section id="virtual-cred-mgr-commands">
```

index.html
mohamedamir: They aren't necessarily required no.
I was reluctant to add them to avoid complicating the algorithm unnecessarily.
I will revert all changes to the algorithm.

mohamedamir: I will revert all changes to the algorithm as discussed in another comment thread.

index.html
mohamedamir: I will revert all changes to the algorithm as discussed in another comment thread.

index.html
mohamedamir: I will revert all changes to the algorithm as discussed in another comment thread.

index.html
mohamedamir: I will revert all changes to the algorithm as discussed in another comment thread.

mohamedamir: I will revert all changes to the algorithm as discussed in another comment thread.

index.html
mohamedamir: I will revert all changes to the algorithm as discussed in another comment thread.

index.html
mohamedamir: I will revert all changes to the algorithm as discussed in another comment thread.

mohamedamir: I will revert all changes to the algorithm as discussed in another comment thread.

index.html
mohamedamir: @timcappalli I am aware of your preferences to use "Credential Manager" instead of holder or wallet, but I think we should do this holistically in the whole spec at one time if we decide to. I won't start with the testing.

@marcoscaceres Yes, I can see we haven't used wallet formally, but given this API is exposed to web developers as well as browser engineers, I believe "wallet" is the more widely-used term in the ecosystem.
"holder" might be even more confusing IMHO.
We can also add a definition to wallet in this section. WDYT?

index.html
mohamedamir: Done!

index.html
mohamedamir: YEs, but it's to simulate a user "declining" the flow.
I feel "decline" is more natural.
Let me know if you feel strongly.

index.html
marcoscaceres: We can leave ReSpec to generate the ids. It knows best.

index.html
marcoscaceres: Discussed with @mohamedamir and drafted:
https://github.com/w3c-fedid/digital-credentials/pull/386/

And hopefully that covers the "manager" aspect that @timcappalli would like included in the definition, while giving us a well understood concept by developer and implementers.

index.html
```diff
@@ -2141,6 +2141,265 @@ <h2>
         This section is a work in progress as this document evolves.
       </p>
     </section>
+    <section id="user-agent-automation" data-cite="webdriver">
```
marcoscaceres: Nit: let ReSpec handle the id generation. 
```suggestion
    <section data-cite="webdriver">
```

index.html
marcoscaceres: Pro-tip: you can also omit the section tag (ReSpec adds it automatically)

index.html
marcoscaceres: No, "declining" is fine... was mostly an editorial suggestion.

index.html
```diff
@@ -2141,6 +2141,265 @@ <h2>
...
+              </th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td>
+                Virtual Wallet Support
+              </td>
+              <td>
+                `digital-credentials:virtualWallet`
```
```suggestion
                `digital-credentials:virtualHolder`
```

index.html
TallTed: It would be helpful to link to this comment thread from all the "I will revert all changes to the algorithm as discussed in another comment thread." comments, and to add a bit more about what's happening. My suggested language changes remain my suggestions, unless there are other changes being made that obviate my suggestions, to which I would appreciate direct link, whether it's to another comment on this PR or to a comment on another PR. (Note that the timestamp of each comment conceals an `href` that goes directly to that comment, making it easy to link from here to there.)

index.html
TallTed: I think "virtual (whatever)" is not accurate and will be misunderstood.

Better might be "software (whatever)" (even if it's implemented on hardware, it's still software doing the lifting).

Rather than `Virtual Credential Manager Support`, I would suggest `Support for Software Credential Managers` or perhaps `Support for Agentic Credential Managers`.

index.html
TallTed: ... but if you keep the `<section>` tags straight, you don't have to keep the `<h1>`–`<h5>` straight (because ReSpec adjusts their depth automatically)

index.html
marcoscaceres: @TallTed, right, but these are literal "virtual" things tho (i.e., fake and temporary and created temporally for purpose). This follows what web authn does too (Virtual authenticators). See https://www.w3.org/TR/webauthn-2/#sctn-automation-virtual-authenticators

mohamedamir: I discussed this separately with both @timcappalli and @marcoscaceres.
I am keeping it as "Virtual Wallet" for better web developer ergonomics.
I will resolve all related comments accordingly!

index.html
```diff
@@ -2267,7 +2141,265 @@ <h2>
...
-    <section class="prefer-full-spec-title" id="index"></section>
+    <section id="user-agent-automation" data-cite="webdriver">
+      <h2>
+        User Agent Automation
+      </h2>
+      <p>
+        For the purposes of user agent automation and web application testing,
+        this document defines a number of [[WebDriver]] [=extension commands=].
+      </p>
+      <section id="webdriver-extension-capability">
```
marcoscaceres: Nit: no need for section here.

index.html
```diff
@@ -2267,7 +2141,265 @@ <h2>
...
+                Indicates whether the endpoint node supports all Virtual Wallet
+                commands.
+              </td>
+            </tr>
+          </tbody>
+        </table>
+        <p>
+          When <a data-cite="webdriver#dfn-validate-capabilities">validating
+          capabilities</a>, the extension-specific substeps to validate
+          `digital-credentials:virtualWallet` with `value` are the following:
```
marcoscaceres: The "value" is an argument, right? If so use |value|. 

```suggestion
          `digital-credentials:virtualWallet` with |value| are the following:
```

Same with the other variables that are set.

index.html
```diff
@@ -2267,7 +2141,265 @@ <h2>
...
+          of the Virtual Wallet commands, the match is unsuccessful.</code>
+          </li>
+          <li>
+            <code>Otherwise, the match is successful.</code>
+          </li>
+        </ol>
+      </section>
+      <section id="virtual-wallet-commands">
+        <h3>
+          <code>Virtual Wallet Commands</code>
```
marcoscaceres: Nit: why wrap in code?

index.html
```diff
@@ -2267,7 +2141,265 @@ <h2>
...
+              <tr>
+                <th>
+                  HTTP Method
+                </th>
+                <th>
+                  URI Template
+                </th>
+              </tr>
+            </thead>
+            <tbody>
```
marcoscaceres: Nit: tbody is redundant here... gets added "for free".

index.html
marcoscaceres: oh, here... JSON Object might not bet the same as a JSON "Web Driver" Object.

mohamedamir: I checked and web driver references ECMAScript's object.
https://www.w3.org/TR/webdriver2/#dfn-object

Should I then just use [JSON], or probably no need to reference anything?
WDYT?

index.html
mohamedamir: Done!

index.html
mohamedamir: Done!

index.html
mohamedamir: Removed the tag

index.html
mohamedamir: Done!

index.html
mohamedamir: Resolving as per our discussion!

index.html
mohamedamir: Resolving as per our discussion

index.html
mohamedamir: Resolving as per our discussion

index.html
mohamedamir: Resolving as per our discussion

index.html
mohamedamir: Resolving as per our discussion

index.html
mohamedamir: Resolving as per our discussion

index.html
mohamedamir: Resolving as per our discussion

index.html
mohamedamir: Resolving as per our discussion

index.html
mohamedamir: Yup, sorry!
Removed!

index.html
mohamedamir: Removed!

index.html
mohamedamir: @marcoscaceres regarding your comment in this thread, what should actually be done here?
Should we add a section to the coordinator?
I am wondering if I misunderstood your comments here regarding the automated section.
Or you think it's sufficiently clear that no need to update the algorithms?

index.html
mohamedamir: @sadym-chromium 
Hi Maksim,

Could you chime in here?
Also, a review would be really appreciated!

Thank you!

index.html
```diff
@@ -2136,6 +2136,248 @@ <h2>
         MUST be operable in a device-independent manner.
       </p>
     </section>
+    <section data-cite="webdriver">
+      <h2>
+        User Agent Automation
```
whimboo: If you consider to add support for WebDriver BiDi at a later point it might be good to already put the classic support into its subsection. Otherwise it's unclear which specification this PR is related (unless you check the links in detail).

index.html
mohamedamir: Done!

index.html
mohamedamir: Thank you for the feedback!
Done!

index.html
mohamedamir: Done!

index.html
```diff
@@ -2136,6 +2136,166 @@ <h2>
...
+      </h2>
+      <p>
+        This section describes how to automate User Agent interactions for
+        testing.
+      </p>
+      <p>
+        For the purposes of user agent automation and web application testing,
+        this document defines a number of [[WebDriver-BiDi]] [=extension
+        commands=].
```
sadym-chromium: NIT: even though it is not strictly prohibited, having the break lines in the formal definitions in the spec makes search through the source code and editing harder. Consider not breaking line inside `<dfn>...</dfn>` and `[=...=]`

index.html
```diff
@@ -2136,6 +2136,166 @@ <h2>
...
+          <thead>
+            <tr>
+              <th>
+                Method
+              </th>
+            </tr>
+          </thead>
+          <tr>
+            <td>
+              `digitalCredentials.setResponse`
```
sadym-chromium: Is the scope of the command global for the whole browser? Is it aligned with implementation? Other option would be to make it more flexible and set per top-level browsing context (page), user context (browser context in CDP terms) or global.

index.html
```diff
@@ -2136,6 +2136,166 @@ <h2>
...
+        The following commands allow controlling the behavior of a virtual
+        wallet for testing purposes. These commands affect the <dfn>next wallet
+        behavior</dfn> of the next
+        {{DigitalCredential/[[DiscoverFromExternalSource]](origin, options,
+        sameOriginWithAncestors)}} or {{DigitalCredential/[[Create]](origin,
+        options,sameOriginWithAncestors)}} call.
+      </p>
+      <section id="set-wallet-response">
+        <h4>
+          <dfn>Set Wallet Response</dfn>
```
sadym-chromium: Is mismatch between the command ("digitalCredentials.setResponse") and description ("Set Wallet Response") intentional? I find it a bit confusing.

index.html
```diff
@@ -2136,6 +2136,166 @@ <h2>
...
+        this document defines a number of [[WebDriver-BiDi]] [=extension
+        commands=].
+      </p>
+      <h3>
+        Virtual Wallet Commands
+      </h3>
+      <p>
+        The following commands allow controlling the behavior of a virtual
+        wallet for testing purposes. These commands affect the <dfn>next wallet
+        behavior</dfn> of the next
```
sadym-chromium: I would shape it more like
> A [=remote end=] has a \<dfn>next wallet behavior\</dfn> which is a string (or define all possible striong literals here) or null, initially null."

After that, in the definition of `DigitalCredential/[[DiscoverFromExternalSource]]` and `DigitalCredential/[[Create]]` check the if `[=next wallet behavior=]` is not null, and if so return it and set to null.

This would make the behavior well-defined and aligns with the other specifications.

Examples:
* https://w3c.github.io/webdriver-bidi/#download-behavior
* https://www.w3.org/TR/geolocation/#dfn-set-emulated-position-data referenced from https://www.w3.org/TR/geolocation/#dfn-set-emulated-position-data

index.html
marcoscaceres: Ah, then let's use {{Object}}. 

I'll see if I can get Web Driver folks to not redefine things.

mohamedamir: This comment threat is no obsolete, resolving!

index.html
mohamedamir: Resolving this since webdriver experts are already on the PR!

index.html
mohamedamir: Done!
Thank you!

index.html
mohamedamir: I have re-written the section.
Resolving this!

index.html
mohamedamir: Done!
Thank you!

index.html
```diff
@@ -2453,6 +2507,107 @@ <h2>
...
+              The `digitalCredentials.setVirtualWalletBehavior` Command
+            </h5>
+            <pre class="cddl">
+digitalCredentials.SetVirtualWalletBehavior = (
+  method: "digitalCredentials.setVirtualWalletBehavior",
+  params: digitalCredentials.SetVirtualWalletBehaviorParameters
+)
+        </pre>
+            <p>
+              <strong>Return Type:</strong> `EmptyResult`
```
whimboo: Please make sure to also tag the return type as CDDL Also each CDDL definition needs the `local-cddl` or `remote-cddl` attribute set. You can check the BiDi or another specification (which implements BiDi commands) in how to get this done.

index.html
```diff
@@ -2453,6 +2507,107 @@ <h2>
...
+        <section>
+          <h4>
+            Types
+          </h4>
+          <pre class="cddl">
+digitalCredentials.VirtualWalletAction = "respond" / "decline" / "wait"
+
+digitalCredentials.SetVirtualWalletBehaviorParameters = {
+  action: digitalCredentials.VirtualWalletAction,
+  ? response: object,
```
whimboo: How do those objects look like? I'm asking because I wonder if it can be defined in more details? There is also no `object` type definition in CDDL as far as I can see.

index.html
```diff
@@ -2453,6 +2507,107 @@ <h2>
...
+            <ol class="algorithm">
+              <li>Let |action| be the value of the `action` field of |command
+              parameters|.
+              </li>
+              <li>Let |response| be the value of the `response` field of
+              |command parameters|, if present, and `null` otherwise.
+              </li>
+              <li>If |action| is `"respond"`:
+                <ol>
+                  <li>If |response| is `null`, return an [=error=] with [=error
```
whimboo: Given that `object` as used above is not a valid CDDL type, it's not clear if `null` is covered. If not we need to explicitly specify `null` as alternative type.

index.html
```diff
@@ -1464,6 +1491,33 @@ <h3>
       </li>
       <li>Let |promise| be [=a new promise=].
       </li>
+      <li>If the [=user agent=]'s [=active virtual wallet behavior=] is not
```
whimboo: This is all the same content as above if I didn't miss something.  Couldn't this be set as a new algorithm that will just be called from those two locations?

mohamedamir: Great idea!
I've extracted the virtual wallet logic into a new Handle Virtual Wallet Behavior algorithm.

index.html
mohamedamir: The definition of those object is beyond this spec.
As far as the DC API is concerned, a valid JSON object is good enough.

I have used `text` instead of `object`

Could you please take another logic also on the parsing logic and let me know if it makes sense?

index.html
mohamedamir: Done!
Regarding`VirtualWalletAction`, I tagged it as `remote-cddl`
Does this makes sense?

index.html
mohamedamir: I used text.
What do you think?

whimboo: If the object is not recursive something like `{ * text => any }` should probably work as well and you could hook into [WebDriver BiDi's serialization algorithm](https://w3c.github.io/webdriver-bidi/#serialize-as-a-remote-value) to have true type-attributed values for the responses which not only have to be atomic types.

index.html
whimboo: That should be fine, yes.

index.html
```diff
@@ -2453,6 +2463,160 @@ <h2>
         MUST be operable in a device-independent manner.
       </p>
     </section>
+    <section data-cite="webdriver webdriver-bidi">
```
whimboo, line 2754: Please make sure to add a link to this section to https://w3c.github.io/webdriver-bidi/#external-specifications.

index.html
mohamedamir: Done!
Thanks a lot for the pointer!
Could you please take another look?

index.html
mohamedamir: This has been addressed!

index.html
```diff
@@ -2607,6 +2569,56 @@ <h5>
...
+              </li>
+            </ol>
+          </li>
+          <li>If |action| is `"respond"`:
+            <ol>
+              <li>Let |json string| be the result of [=serialize an Infra value
+              to a JSON string|serializing=] |response|.
+              </li>
+              <li>Let |js object| be the result of parsing JSON into a
+              JavaScript value given |json string| in |global|'s [=relevant
```
```suggestion
              <li>Let |JSON string| be the result of [=serialize an Infra value
              to a JSON string|serializing=] |response|.
              </li>
              <li>Let |JS object| be the result of parsing JSON into a
              JavaScript value given |JSON string| in |global|'s [=relevant
```

index.html
```diff
@@ -2607,6 +2569,56 @@ <h5>
...
+              realm=].
+              </li>
+              <li>Let |credential| be a new {{DigitalCredential}} instance.
+              </li>
+              <li>Set |credential|'s {{DigitalCredential/data}} attribute to
+              |js object|.
```
```suggestion
              |JS object|.
```

TallTed (changes requested):
Small stuff

index.html
```diff
@@ -2453,6 +2463,163 @@ <h2>
...
+          options, sameOriginWithAncestors)}} and
+          {{Credential/[[Create]](origin,options, sameOriginWithAncestors)}}
+          calls.
+        </p>
+        <section>
+          <h4>
+            Types
+          </h4>
+          <pre class="cddl" data-cddl-module="remote-cddl">
+digitalCredentials.VirtualWalletAction = "respond" / "decline" / "wait"
```
whimboo: nit: lets have it alphabetically sorted.

```suggestion
digitalCredentials.VirtualWalletAction = "decline" / "respond" / "wait"
```

index.html
```diff
@@ -2453,6 +2463,163 @@ <h2>
...
+
+digitalCredentials.SetVirtualWalletBehaviorParameters = {
+  action: digitalCredentials.VirtualWalletAction,
+  ? response: { * text =&gt; any },
```
whimboo: These lines are exported as CDDL and as such should not contain HTML entity references.

```suggestion
  ? response: { * text => any },
```

index.html
```diff
@@ -2453,6 +2463,163 @@ <h2>
...
+}
+    </pre>
```
whimboo: Is it just this line for which the intentation is off?

index.html
```diff
@@ -2453,6 +2463,163 @@ <h2>
...
+            </h5>
+            <p>
+              <strong>Command Type:</strong>
+            </p>
+            <pre class="cddl" data-cddl-module="remote-cddl">
+digitalCredentials.SetVirtualWalletBehavior = (
+  method: "digitalCredentials.setVirtualWalletBehavior",
+  params: digitalCredentials.SetVirtualWalletBehaviorParameters
+)
+      </pre>
```
whimboo: Same here for the indentation.

index.html
```diff
@@ -2453,6 +2463,163 @@ <h2>
...
+            <p>
+              <strong>Return Type:</strong>
+            </p>
+            <pre class="cddl" data-cddl-module="local-cddl">
+digitalCredentials.SetVirtualWalletBehaviorResult = EmptyResult
+      </pre>
```
whimboo, line 2785: and here.

index.html
```diff
@@ -2453,6 +2463,163 @@ <h2>
...
+            <p>
+              The [=remote end steps=] for the
+              `digitalCredentials.setVirtualWalletBehavior` command, given
+              |session| and |command parameters|, are:
+            </p>
+            <ol class="algorithm">
+              <li>Let |action| be the value of the `action` field of |command
```
whimboo: In the BiDi spec we use the following nowadays. It simplifies related lines. Feel free to adapt here and where it's helpful:

```suggestion
              <li>Let |action| be |command parameters|["<code>action</code>"]. 
```

Also make sure to not add line breaks in references.

index.html
```diff
@@ -2453,6 +2463,163 @@ <h2>
...
+          </h4>
+          <pre class="cddl" data-cddl-module="remote-cddl">
+digitalCredentials.VirtualWalletAction = "respond" / "decline" / "wait"
+
+digitalCredentials.SetVirtualWalletBehaviorParameters = {
+  action: digitalCredentials.VirtualWalletAction,
+  ? response: { * text =&gt; any },
+}
+    </pre>
+          <ul>
```
whimboo: Lets make sure those entries are all related to `action` within `SetVirtualWalletBehaviorParameters`. Maybe add as well a sentence first which starts with: `The <code>SetVirtualWalletBehaviorParameters</code> type represents ...`.

index.html
```diff
@@ -2453,6 +2463,163 @@ <h2>
...
+              |session| and |command parameters|, are:
+            </p>
+            <ol class="algorithm">
+              <li>Let |action| be the value of the `action` field of |command
+              parameters|.
+              </li>
+              <li>Let |response| be the value of the `response` field of
+              |command parameters|, if present, and `null` otherwise.
+              </li>
+              <li>If |action| is `"respond"`:
```
whimboo: Shall we maybe check `response` to be undefined for the other action types and raise an error if it's not?

whimboo (changes requested):
Thanks. I've a couple more questions and notes regarding the WebDriver specific section.

index.html
mohamedamir: @marcoscaceres I struggled a lot with, `tidy` is auto-converting it, and I couldn't get it to keep it by updating the config.

The only solution I have found is to to keep it inside a `<script>` that has some css to make it visible.
But even if with that, tidy format the code weirdly and then look like this screenshot.

What do you think?
Any ideas how to fix it?
FWIW: it seems bikeshed has nicer treatment for cddl while respec doesn't.

image

index.html
mohamedamir: Done! (was a weird effect from tidy as well)

index.html
mohamedamir: Done!

index.html
mohamedamir: Done!

index.html
mohamedamir: We are using ReSpec IIUC, ReSpec [doesn't have support](https://github.com/w3c/webref/issues/1353#issuecomment-2392949396) for CDDL (it doesn't index the members ...etc.)
@marcoscaceres am I overlooking something?

index.html
mohamedamir: Done!
FWIW: those line breaks were a `tidy` issue.

index.html
mohamedamir: Good call!
Done!

index.html
mohamedamir: Resolving this since the algorithms have been updated.

index.html
whimboo: @mohamedamir, mind creating a WebDriver BiDi PR to get this external specification added?

index.html
whimboo: How complicated would it be to move to Bikeshed? Only that way we can collect all the CDDL definitions from external specs via webref. 

CC @tidoust who helped us with https://github.com/w3c/webdriver-bidi/pull/927.

index.html
mohamedamir: I have sent 
https://github.com/w3c/webdriver-bidi/pull/1107
for review.

index.html
marcoscaceres: I’ve never had issues with Tidy? Happy to check what’s happening. 

Converting to bikeshed has too many tradeoffs (like, we can’t tidy!).

tidoust: > How complicated would it be to move to Bikeshed? Only that way we can collect all the CDDL definitions from external specs via webref.

As far as I can tell, the CDDL block in this PR is correctly flagged for the crawler to pick it up and for CDDL definitions to start appearing in webref once the PR gets merged. There should be no need to move to Bikeshed for that!

whimboo: Oh, that's wonderful to hear! Thanks for this feedback.

In this case it's all fine from my side. Thank you @mohamedamir!

whimboo (approved):

index.html
mohamedamir: Thank you both very much.!
@whimboo Thanks a lot for all the guidance and support through out this PR!
Much appreciated!

index.html
```diff
@@ -1395,6 +1395,11 @@ <h3>
       <li>[=credential request coordinator/Prepare credential requests=] with
       |document|, |requests|, |promise|, and |signal|.
       </li>
+      <li>Let |handled| be the result of running [=handle virtual wallet
+      behavior=] given |promise| and |global|.
+      </li>
+      <li>If |handled| is `true`, return |promise|.
+      </li>
       <li>Return |promise|.
```
TallTed: What's the point of this `If` test, given that both `|handled| is true` and *all other values of `|handled|`* (thus, in aggregate, *all values of `|handled|`*) result in `return |promise|`?

index.html
```diff
@@ -1452,6 +1457,11 @@ <h3>
...
```
TallTed: As in [lines +1401 to 1403](https://github.com/w3c-fedid/digital-credentials/pull/381/files#r3002197491)

timcappalli (approved):

marcoscaceres:
I did a systematic review of the testing framework against all the algorithms (including the new "present the credential request" algorithm from #419). Here's what I found:

## What the framework CAN test (great coverage)

The `"wait"` action is especially useful — it enables testing:
- **Concurrent requests** — `"wait"` + second `get()` call to verify the error
- **Abort signal during presentation** — `"wait"` + abort the signal
- **Document becoming inactive** — `"wait"` + navigate away

And `"respond"` covers the happy path well.

## Issues I'd like to discuss

### 1. `"decline"` rejects with `NotAllowedError` — should be `AbortError`

The `"decline"` action currently rejects with `NotAllowedError`. But we agreed in the [#419 discussion](https://github.com/w3c-fedid/digital-credentials/pull/419#discussion_r2561523770) that user cancellation should reject with `AbortError` — distinct from platform permission denial (`NotAllowedError`). The "present the credential request" algorithm (step 5) reflects this:

> If the user cancels the operation or no credential was selected: Let |error| be a newly created `"AbortError"` DOMException.

Since `"decline"` simulates the user dismissing the chooser, it should reject with `AbortError` to match the spec. `NotAllowedError` is reserved for when the *platform* denies permission (step 6a).

### 2. `"respond"` doesn't set `protocol`

The `"respond"` action creates a `DigitalCredential` with `.data` but doesn't set `.protocol`. The spec requires both attributes (present step 7c-e). Tests can't verify that the correct protocol enum value is returned.

Should the `"respond"` action accept an optional `protocol` parameter? Or should it default to the first valid protocol from the request?

### 3. No way to test platform error mapping

The "present" algorithm maps 4 types of platform errors (step 6):
- Permission denied → `NotAllowedError`
- Malformed request data → `TypeError`  
- Already in progress → `InvalidStateError`
- Catch-all → `OperationError`

With only respond/decline/wait, none of these error conditions are testable. Would it be worth adding an `"error"` action that accepts an error type parameter to simulate platform-level failures?

### 4. No way to test response parsing failures

Because `"respond"` pre-validates the response as a JSON object (`{ * text => any }`), there's no way to test:
- Holder returning invalid JSON (present step 7a → parse error)
- Holder returning valid JSON that isn't an object, e.g. `"hello"` or `42` (present step 7b → `TypeError`)

A `"respond-raw"` action accepting an arbitrary string could cover these paths.

## Summary

| Framework action | What it tests | Gap |
|-----------------|---------------|-----|
| `"respond"` | Happy path | Missing `protocol`; can't test bad responses |
| `"decline"` | User cancellation | Wrong error type (`NotAllowedError` vs `AbortError`) |
| `"wait"` | Concurrent requests, abort, timeouts | None — works well |
| (missing) | Platform errors | No error injection capability |
| (missing) | Response parsing failures | No raw response injection |

Happy to help iterate on this. The framework is a great foundation — these are refinements to get full algorithm coverage.

marcoscaceres (changes requested):
See https://github.com/w3c-fedid/digital-credentials/pull/381#issuecomment-4221822734 ... there's still some gaps. 

We can split those into issues or address them here. I don't mind.

index.html
mohamedamir: The algorithm isn't complete yet

It will be something like 
```
      <li>[=credential request coordinator/Prepare credential requests=] with
      |document|, |requests|, |promise|, and |signal|.
      </li>
      <li>Let |handled| be the result of running [=handle virtual wallet
      behavior=] given |promise| and |global|.
      </li>
      <li>If |handled| is `true`, return |promise|.
      </li>
      <li>[=credential request coordinator/Present credential requests=] with
      |document|, |requests|, |promise|, and |signal|.
      </li>
      <li>Return |promise|.
```

once https://github.com/w3c-fedid/digital-credentials/pull/419 is merged.!

index.html
mohamedamir: Similar to above!

mohamedamir:
Thanks @marcoscaceres for the feedback

> ### 1. `"decline"` rejects with `NotAllowedError` — should be `AbortError`

That's not correct, we agreed to return `NotAllowedError` for user cancellation, and keep `AbortError` for programmatic Abort from the developer.

> ### 2. `"respond"` doesn't set `protocol`
Great find, fixed!

> ### 3. No way to test platform error mapping
I am not sure if this is something we would like to add given that platform error aren't standardized.
WDYT?

> ### 4. No way to test response parsing failures
I filed https://github.com/w3c-fedid/digital-credentials/issues/498 to track this issue, since I am not sure it is worth the complexity at the moment.

marcoscaceres:
CDDL suport inbound! https://github.com/speced/respec/pull/5201 😎

marcoscaceres:
> I am not sure if this is something we would like to add given that platform error aren't standardized.
> WDYT?

See the "present" branch... I've put them into classes of errors, so maybe we could? We could even formalize the classes of platform errors for this purpose and link to them from here. That would be kinda nice.

marcoscaceres:
> That's not correct, we agreed to return NotAllowedError for user cancellation, and keep AbortError for programmatic Abort from the developer.

Yeah, sorry... I think I was looking at the wrong version of the spec.

mohamedamir:
> > I am not sure if this is something we would like to add given that platform error aren't standardized.
> > WDYT?
> 
> See the "present" branch... I've put them into classes of errors, so maybe we could? We could even formalize the classes of platform errors for this purpose and link to them from here. That would be kinda nice.

Fair enough, I have submitted https://github.com/w3c-fedid/digital-credentials/issues/500 to track this.
We can revisit once the present part is merged!

marcoscaceres:
Thanks for working on this. WebDriver automation is critical for testability, and the overall approach (virtual wallet behavior rather than full virtual wallets) is the right one. We've been evaluating this from a WebKit implementation perspective, and I want to flag several issues we've found. Grouped by category:

## Algorithm ordering

The "handle virtual wallet behavior" hook is inserted *after* the "Prepare credential requests" step. In WebKit, "Prepare" validates requests, transitions the coordinator to a "Presenting" state, and calls `showDigitalCredentialsPicker()`, which sends IPC to the UIProcess to present a system sheet. By the time the virtual wallet check runs, the picker is already being shown.

This means:

- The virtual wallet would resolve the promise while the coordinator's state machine is still in "Presenting" state, bypassing the normal resolution flow.
- The coordinator's `dismissDigitalCredentialsPicker` is never called, potentially leaving the UIProcess picker alive.

The virtual wallet check needs to run before the presentation flow begins.

## Scoping and lifecycle

The spec says "Set the **user agent's** active virtual wallet behavior." This is a UA-global singleton. Other WebDriver automation APIs scope their state differently: FedCM commands use `/session/{session id}/fedcm/...` (session-scoped), and WebAuthn virtual authenticator commands take a `browsingContextHandle` parameter (context-scoped).

UA-global scoping has two consequences:

1. **Test isolation is broken.** A behavior set in one test leaks into subsequent tests unless explicitly overwritten. There is no `clearVirtualWalletBehavior` command and no specified cleanup when the WebDriver session ends. Compare with WebAuthn, which provides `removeVirtualAuthenticator` and `removeAllVirtualAuthenticatorCredentials`.

2. **Cross-context interference during automation.** If multiple tabs or sessions are active, a page in one context could observe the virtual wallet behavior configured for a different context.

## Automation security boundary

The `handle virtual wallet behavior` algorithm runs on every credential request, including in non-automated browsing. It checks whether the "active virtual wallet behavior" is null, but does not verify that the user agent is under automation. While `setVirtualWalletBehavior` is gated to WebDriver BiDi (content JS cannot invoke it), adding an explicit "if not under automation, return false" guard as step 0 would provide defense in depth.

## BiDi/Classic protocol mismatch

We also reviewed the linked WPT PR ([web-platform-tests/wpt#58587](https://github.com/web-platform-tests/wpt/pull/58587)) to understand how the spec's commands would be exercised in practice. We realize the WPT implementation may not have been updated to match the current spec text yet, but flagging the current gaps so they're tracked:

This PR defines only BiDi commands (CDDL types like `digitalCredentials.SetVirtualWalletBehaviorParameters`), but the WPT implementation currently uses WebDriver Classic HTTP endpoints: `POST digital-credentials/response`, `POST digital-credentials/response_wait`, and `POST digital-credentials/response_decline`. These Classic endpoints are not defined in this PR or the WebDriver specification. The WPT tests will need to align with whichever wire protocol the spec ultimately defines.

The WPT testdriver API also currently diverges from the spec's command shape in a couple of ways that will need reconciling:

- `set_virtual_wallet_response(data)` passes a plain string as `data`, but the spec's CDDL requires `response` to be `{ * text => any }` (a map).
- The testdriver omits the `protocol` parameter, which the spec requires to be non-null when action is "respond" (step 4.1 would return an error).

## Interoperability

The module is OPTIONAL, but the WPT tests have no feature detection or metadata gating. They will permanently fail in browsers that don't implement the module. Gecko has no implementation commitment listed in this PR.

The tests also hard-code `openid4vp-v1-unsigned` and `openid4vci` as protocol values. Since the virtual wallet bypasses all protocol validation and returns whatever was configured, the tests pass without exercising any browser's actual protocol-handling code paths.

## Adversarial testing capability

For security testing, we'd want this API to support two attack directions:

**Wallet → RP/content process:** Can the virtual wallet return data that tests the content process's resilience? The current API allows arbitrary response shapes, which is good. But there is no timing control (the "wait" action is binary, with no configurable delay for testing race conditions), no mechanism for sequenced responses (e.g., "decline the first request, respond to the second"), and no way to configure different behaviors for `create()` vs `get()`.

**RP → wallet/UA:** The web page already controls request data, but the virtual wallet ignores it and returns canned responses. There is no way for a test to observe what the UA forwarded to the virtual wallet (e.g., whether `topOrigin` and `documentOrigin` were correctly included, whether request validation occurred). Compare with WebAuthn's `getVirtualAuthenticatorCredentials`, which provides this kind of introspection.

mohamedamir:
Hi @marcoscaceres,

Thanks for the detailed review and for flagging these issues! I have addressed your feedback in the latest commits. Here is a summary of how I handled each point:

1. **Algorithm ordering**: I moved the `handle virtual wallet behavior` hook inside the `Prepare credential requests` algorithm, specifically **after** the validation steps but **before** `Present the credential request`. This ensures that request validation still occurs while avoiding triggering the UI presentation flow in automated tests.
2. **Scoping and lifecycle**: I have updated the spec to associate the `active virtual wallet behavior` with the **`WebDriver session`** instead of the global `user agent`. This should prevent cross-session interference. I also added a `"clear"` action to `VirtualWalletAction` to allow tests to explicitly reset the behavior, addressing the lack of a cleanup mechanism.
3. **Automation security boundary**: I added an explicit check as the first step in `Handle Virtual Wallet Behavior`: *"If the user agent is not under automation, return false."*
4. **BiDi/Classic protocol mismatch**:  I am working on updating the WPT tests to use WebDriver BiDi to align with this spec.

Please let me know if this addresses your concerns or if you have further feedback!

TallTed:
There appear to have been a large number of changes since [my last review](https://github.com/w3c-fedid/digital-credentials/pull/381#pullrequestreview-4022541675). It seems likely that I don't need to [re-review every line of this PR](https://github.com/w3c-fedid/digital-credentials/pull/381#event-24803368395), but rather only those recent changes. However, GitHub is refusing to provide a link to "all changes since I last viewed". Perhaps you can quickly build such a link?

marcoscaceres:
Hi @TallTed,

Mohamed's most recent commit (`ba531c5`, April 23) addresses feedback I gave earlier that day. Here's what changed in the spec language:

**1. Algorithm ordering (moved the virtual wallet hook earlier)**
The "handle virtual wallet behavior" steps were removed from both `get()` and `create()` (where they ran *after* "Prepare credential requests") and moved *into* the "Prepare credential requests" algorithm itself, before the presentation flow begins. This prevents the virtual wallet from resolving a promise while the UI picker is already being shown.

**2. Scoping: session instead of global**
All references to "the user agent's active virtual wallet behavior" were changed to "the WebDriver session's active virtual wallet behavior." This scopes the state to individual sessions rather than being UA-global, preventing cross-session test interference.

**3. New "clear" action**
A `"clear"` value was added to `VirtualWalletAction`. The `setVirtualWalletBehavior` algorithm now checks for `"clear"` and sets the session's active behavior to `null`, giving tests an explicit cleanup mechanism.

**4. Automation security guard**
A new first step was added to the "handle virtual wallet behavior" algorithm: "If the user agent is not under automation, return false." This is a defense-in-depth check so the virtual wallet path can never run outside of WebDriver automation.

Those are the substantive language changes. Some minor reformatting of the `VirtualWalletAction` description list also happened (line wrapping), but no wording changes there.

**Suggested focus areas for your editorial eye:**

- **The `VirtualWalletAction` descriptions** ("decline", "respond", "wait", "clear" bullet list). These were rewrapped and reformatted. Worth checking that they're parallel in structure, consistent in voice, and clear to a first-time reader.
- **The "handle virtual wallet behavior" algorithm steps.** These were relocated and reworded ("user agent's" became "WebDriver session's"). Worth checking that the new phrasing reads naturally and that subject/object references are unambiguous.
- **The `"clear"` action description** is noticeably shorter than the other three. It might benefit from a bit more explanatory text to match the level of detail in its siblings.
- **Consistency of "return false" / "return true"** language across the handle algorithm. The new automation guard and the existing steps should use a consistent style.

Hope that helps!

index.html
```diff
@@ -1680,8 +1685,8 @@ <h3>
       </li>
       <li>[=Consume user activation=] of |global|.
       </li>
-      <li>Let |promise| be [=a new promise=] in the [=relevant realm=] of
-      [=this=].
+      <li>Let |promise| be [=a new promise=].
```
```suggestion
      <li>Let |promise| be [=a new promise=] in the [=relevant realm=] of
      [=this=].
```

index.html
```diff
@@ -1682,6 +1687,7 @@ <h3>
       </li>
       <li>Let |promise| be [=a new promise=] in the [=relevant realm=] of
       [=this=].
+
```
```suggestion
```

marcoscaceres:
## Spec Clarifications for DC PR #381 — Draft Comment

**Target:** https://github.com/w3c-fedid/digital-credentials/pull/381

---

During the WebKit implementation of this PR (https://bugs.webkit.org/show_bug.cgi?id=306292), I identified a few issues that need clarification:

### 1. Type mismatch: `protocol` field in CDDL vs `DigitalCredential.protocol` attribute

The CDDL defines the `protocol` parameter as free text:

```
? protocol: text,
```

But the `handle virtual wallet behavior` algorithm (step 9.4) sets:

> Set |credential|'s `protocol` attribute to |protocol|.

`DigitalCredential.protocol` is of type `DigitalCredentialProtocol` (a union of `DigitalCredentialPresentationProtocol` and `DigitalCredentialIssuanceProtocol`), which is an IDL enum. Setting an arbitrary text value into an enum attribute doesn't work — the value must be one of the defined enum values.

**Suggestion:** Either:
- Constrain the CDDL `protocol` field to only accept valid enum values (and validate at `setVirtualWalletBehavior` time with "invalid argument" for unknown values), or
- Change the CDDL type from `text` to an enum that mirrors `DigitalCredentialProtocol`.

### 2. Session end behavior for "wait" action

The spec says the "wait" action returns `true` (leaving the promise unsettled), but doesn't specify what happens when the automation session ends while a credential request is in "wait" state.

In WebKit, IPC CompletionHandlers must be resolved. We reject pending handlers with `AbortError` on session destruction, but this should be specified.

### 3. Concurrent "wait" requests for the same context

If two credential requests arrive for the same browsing context while behavior is "wait", the spec doesn't say whether both should be held or if the second supersedes the first.

In our implementation, the second request causes the first to be aborted (rejected with `AbortError`), but this should be clarified.

---

*Found during WebKit implementation: https://bugs.webkit.org/show_bug.cgi?id=306292*

index.html
```diff
@@ -2705,8 +2710,235 @@ <h2>
...
+          virtual wallet behavior=] for |context ID|.
+          </li>
+          <li>If |behavior| is `null`, set |behavior| to the current
+          [=WebDriver session=]'s default [=active virtual wallet behavior=].
+          </li>
+          <li>If |behavior| is `null`, return `false`.
+          </li>
+          <li>Let (|action|, |protocol|, |response|) be |behavior|.
+          </li>
+          <li>If |action| is `"wait"`, return `true`.
```
marcoscaceres: **Context destruction during "wait"**

When the `"wait"` action leaves a promise unsettled, the spec says the session-end MUST reject it (covered in the note below). But what about browsing context destruction? If the tab/frame that initiated the credential request is closed while the promise is pending, the spec doesn't explicitly require the UA to reject it.

Suggestion: add a normative requirement like: "If the browsing context associated with a pending `"wait"` promise is destroyed, the user agent MUST reject the promise with an `"AbortError"` DOMException."

index.html
```diff
@@ -2705,8 +2710,235 @@ <h2>
...
+              <li>Return `true`.
+              </li>
+            </ol>
+          </li>
+        </ol>
+        <p class="note">
+          When the [=WebDriver session=] ends, the [=user agent=] MUST
+          [=reject=] any unsettled {{Promise}} objects that were left pending
+          by the `"wait"` action with an {{"AbortError"}} {{DOMException}}.
+        </p>
```
marcoscaceres: **Individual request validation failures (existing spec text, but relevant to this section)**

The "validate credential requests" algorithm (step 8 in the existing spec) says "If validation fails, throw an appropriate exception." This makes a single bad request kill the entire operation. Some UAs (WebKit) skip individual invalid requests and log a console warning instead, which is arguably better UX (one bad request shouldn't block valid ones).

The spec might want to allow this: "The user agent MAY skip individual requests that fail protocol-specific validation, logging a developer-visible warning, rather than throwing." The empty-result case is already handled by step 7 of "prepare credential requests" (TypeError if validatedRequests is empty).

marcoscaceres (approved):
A few comments after implementing, but looks good. Excellent work @mohamedamir!

index.html
mohamedamir: This doesn't belong to this PR, does it?
This is related to the "validate credential requests" algorithm not the user agent automation.

Am I overlooking something?

index.html
mohamedamir: Done!

index.html
```diff
@@ -2746,8 +2751,243 @@ <h2>
...
```
marcoscaceres: Sorry, this was supposed to be advisory:  

```suggestion
        <p class="note">
          When the [=WebDriver session=] ends, the [=user agent=]
          [=rejects=] unsettled {{Promise}} objects that were left pending
          by the `"wait"` action.
        </p>
```

It's the same as destroying the the frame ... so it just behaves the same way as the document is "no longer fully active".

index.html
marcoscaceres: no, oops.. fixing now...

index.html
```diff
@@ -2746,8 +2751,243 @@ <h2>
...
+          If a new credential request arrives for a [=browsing context=] while
+          a previous request is pending due to the `"wait"` action, the [=user
+          agent=] MUST [=reject=] the previous request's {{Promise}} with an
+          {{"AbortError"}} {{DOMException}} before handling the new request.
+        </p>
+        <p class="note">
+          If the [=browsing context=] associated with a pending `"wait"`
+          promise is destroyed, the [=user agent=] MUST [=reject=] the promise
+          with an {{"AbortError"}} {{DOMException}}.
+        </p>
```
```suggestion
```

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+        Automated Testing
+      </h2>
+      <p>
+        For the purposes of user-agent automation and application testing, this
+        document defines extension modules for the [[[webdriver-bidi]]]
+        specification.
+      </p>
+      <p>
+        It is OPTIONAL for a [=user agent=] to support them.
+      </p>
```
marcoscaceres, line 2789:
```suggestion
      <p>
        For the purposes of user-agent automation and application testing, this
        document defines extension modules for the [[[webdriver-bidi]]]
        specification. It is OPTIONAL for a [=user agent=] to support them.
      </p>
```

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+      <section>
```
marcoscaceres: Nit: we can delete this `<section>` (ReSpec adds section elements automagically)

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+        </h3>
+        <p>
+          The `digitalCredentials` module contains commands for managing and
+          simulating the remote end behavior of digital credential wallets
+          during {{DigitalCredential/[[DiscoverFromExternalSource]](origin,
+          options, sameOriginWithAncestors)}} and
+          {{Credential/[[Create]](origin,options, sameOriginWithAncestors)}}
+          calls.
+        </p>
+        <section>
```
marcoscaceres: As above.

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+
+              digitalCredentials.SetVirtualWalletBehaviorParameters = {
+                action: digitalCredentials.VirtualWalletAction,
+                ? context: text,
+                ? protocol: text,
+                ? response: { * text =&gt; any },
+              }
+          </pre>
+          <p>
+            The {^ digitalCredentials.VirtualWalletAction ^} type represents
```
```suggestion
            The {^digitalCredentials.VirtualWalletAction^} type represents
```

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+          <pre class="cddl" data-cddl-module="remote-cddl">
+              digitalCredentials.VirtualWalletAction = "decline" / "respond" / "wait" / "clear"
+
+              digitalCredentials.SetVirtualWalletBehaviorParameters = {
+                action: digitalCredentials.VirtualWalletAction,
+                ? context: text,
+                ? protocol: text,
+                ? response: { * text =&gt; any },
+              }
+          </pre>
```
marcoscaceres: Nit.. whitespace. 

```suggestion
          <pre class="cddl" data-cddl-module="remote-cddl">
            digitalCredentials.VirtualWalletAction = "decline" / "respond" / "wait" / "clear"
  
            digitalCredentials.SetVirtualWalletBehaviorParameters = {
              action: digitalCredentials.VirtualWalletAction,
              ? context: text,
              ? protocol: text,
              ? response: { * text =&gt; any },
            }
          </pre>
```

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+          <p>
+            The {^ digitalCredentials.VirtualWalletAction ^} type represents
+            the different types of virtual wallet actions.
+          </p>
+          <ul>
+            <li>{^ digitalCredentials.VirtualWalletAction/"decline" ^}: The
```
```suggestion
            <li>{^digitalCredentials.VirtualWalletAction/"decline"^}: The
```

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+            virtual wallet simulates a user cancellation or rejection, aborting
+            the request.
+            </li>
+            <li>{^ digitalCredentials.VirtualWalletAction/"respond" ^}: The
```
```suggestion
            <li>{^digitalCredentials.VirtualWalletAction/"respond"^}: The
```

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+            virtual wallet simulates a successful user interaction and returns
+            the predefined credential response.
+            </li>
+            <li>{^ digitalCredentials.VirtualWalletAction/"wait" ^}: The
```
```suggestion
            <li>{^digitalCredentials.VirtualWalletAction/"wait"^}: The
```

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+            virtual wallet simulates an active, pending prompt, effectively
+            leaving the active promise unsettled to test timeouts or concurrent
+            request handling.
+            </li>
+            <li>{^ digitalCredentials.VirtualWalletAction/"clear" ^}: Clears
```
```suggestion
            <li>{^digitalCredentials.VirtualWalletAction/"clear"^}: Clears
```

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+                ? context: text,
+                ? protocol: text,
+                ? response: { * text =&gt; any },
+              }
+          </pre>
+          <p>
+            The {^ digitalCredentials.VirtualWalletAction ^} type represents
+            the different types of virtual wallet actions.
+          </p>
+          <ul>
```
marcoscaceres: This looks like it should be a definition list that defines each thing, instead of a ul. 

```suggestion
          <dl>
```

index.html
```diff
@@ -2746,8 +2751,227 @@ <h2>
...
+        </section>
+        <section>
+          <h4>
+            Commands
+          </h4>
+          <section>
+            <h5>
+              The `digitalCredentials.setVirtualWalletBehavior` Command
+            </h5>
+            <p>
```
marcoscaceres: Same with these `<p>` strong combos ... these should be in a `dl`

marcoscaceres (commented):
Some nits... getting copilot on them.

marcoscaceres:
@copilot see my latest review... can you fix just the nits and markup (redundant sections) and tidy it?

index.html
mohamedamir: Done!

index.html
mohamedamir: Done!

index.html
```diff
@@ -2767,9 +2767,8 @@ <h3>
         The `digitalCredentials` module contains commands for managing and
         simulating the remote end behavior of digital credential wallets during
         {{DigitalCredential/[[DiscoverFromExternalSource]](origin, options,
-        sameOriginWithAncestors)}} and
-        {{Credential/[[Create]](origin,options, sameOriginWithAncestors)}}
-        calls.
+        sameOriginWithAncestors)}} and {{Credential/[[Create]](origin,options,
```
TallTed: for consistency
```suggestion
        sameOriginWithAncestors)}}, and {{Credential/[[Create]](origin, options,
```

index.html
```diff
@@ -2746,8 +2751,230 @@ <h2>
         MUST be operable in a device-independent manner.
       </p>
     </section>
+    <section data-cite="webdriver webdriver-bidi">
+      <h2>
+        Automated Testing
+      </h2>
+      <p>
+        For the purposes of user-agent automation and application testing, this
```
```suggestion
        For purposes of user-agent automation and application testing, this
```

index.html
```diff
@@ -1680,8 +1685,8 @@ <h3>
       </li>
       <li>[=Consume user activation=] of |global|.
       </li>
-      <li>Let |promise| be [=a new promise=] in the [=relevant realm=] of
-      [=this=].
+      <li>Let |promise : Promise| be [=a new promise=] in the [=relevant
```
TallTed, line 1688: This appears to be the only instance of ` : ` (space-colon-space) in this document. I haven't checked the other documents in the repo so I don't know whether it's the only instance of ` : ` in the repo. It appears to be a typo. Please confirm or deny... :-)

marcoscaceres: ReSpec’s inline-variable parser doesn’t mind, but we can quickly fit it. @copilot can you send a PR for consistency fixing this?

PR #386, opened 2025-10-09
Define Digital Wallet

marcoscaceres:
Closes #???

Introduces a formal definition for "Digital Wallet", clarifying its role and relationship to the "holder" concept. These changes improve clarity, precision, and alignment with the specification's terminology. Consistently use the defined term "[=digital wallet=]" in place of previous informal references like "digital wallet" or "wallet".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/386.html" title="Last updated on Dec 2, 2025, 4:49 AM UTC (c83322c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/386/61739d8...c83322c.html" title="Last updated on Dec 2, 2025, 4:49 AM UTC (c83322c)">Diff</a>

PR #387, opened 2025-10-09
Add terminology section, merge in Model

### LINKED ISSUE #352: Add a terminology section
This is a complicated ecosystem. We need a dedicated terminology section at the beginning of the spec (probably between scope and model).

We have a bunch of definitions scattered throughout different sections. We're also missing definitions for things like `credential manager` (holder is technically a person, not where the credential is stored, and that is an important entity).

Let's use this issue to aggregate terms which should be defined (the definition can be a reference to another spec).

Some to start:
- Issuer
- Verifier
- Holder
- Credential manager
- Wallet
- Client platform

timcappalli:
Closes #352 

* Adds a new Terminology section covering #352 
* Merges in the existing Model section

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/387.html" title="Last updated on Oct 10, 2025, 1:37 PM UTC (fc4ded2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/387/67bf63d...fc4ded2.html" title="Last updated on Oct 10, 2025, 1:37 PM UTC (fc4ded2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -468,12 +468,12 @@
         </p>
       </dd>
       <dt>
-        <dfn>Wallet</dfn>
         <dfn>Digital wallet</dfn>
       </dt>
       <dd>
         A user friendly term for a [=credential manager=] for verifiable digital
-        credentials and other objects like payment cards and tickets.
+        credentials and other objects like payment cards and tickets. Also known
+        as a "wallet" or "identity wallet".
       </dd>
       <dt>
         <dfn data-dfn-for="digital credential" data-local-lt=


index.html
```diff
@@ -435,7 +435,70 @@ <h2 class="informative">
...
+      </dt>
+      <dd>
+        An application, hardware device, or service which securely stores,
+        organizes, manages, and enables presentation of credentials. Digital
+        wallets, password managers, and passkey managers are all examples of
+        credential managers.
+      </dd>
+      <dt>
+        <dfn>Wallet</dfn>
+        <dfn>Digital wallet</dfn>
```
timcappalli: @marcoscaceres any idea why doesn't this work in ReSpec? It works fine with Bikeshed.

image

image

image

marcoscaceres: try:

```suggestion
        <dfn data-local-lt=“wallet”>Digital wallet</dfn>
```

to create an alias.

marcoscaceres: See https://respec.org/docs/#data-local-lt

timcappalli: Yeah I know about that one, but I was hoping for it to still be visible to the reader (see WebAuthn screenshot in the other [comment](https://github.com/w3c-fedid/digital-credentials/pull/387#discussion_r2417377536))

marcoscaceres: Ah! No, please don’t do that. That’s considered bad practice. We have *-lt and -local-lt in both ReSpec and Bikeshed specifically for that reason.

marcoscaceres: Note the case in WebAuth defines entirely different terms, not the alias. 

But they should fix the definition term and definition description markup as what they are doing there is not good. It breaks the semantic relationship expected by HTML, where a defined term should be followed by its definition (bad for accessibility, not just readability).

marcoscaceres: @timcappalli, see https://github.com/w3c-fedid/digital-credentials/pull/389 for the VC definitions. That gets around having to have them here and includes the complete list (plus for all the specs).

index.html
```diff
@@ -435,7 +435,70 @@ <h2 class="informative">
...
+    <h2>
+      Terminology
+    </h2>
+    <h3>Defined by this specification</h3>
+    <!--
+      // NOTE: some of these will eventually reference the VDC Architecture spec from IETF
+    -->
+    <dl class="definitions" data-sort="ascending" data-cite="">
+      <dt>
+        <dfn>Credential manager</dfn>
```
marcoscaceres, line 451: This definition is good, but should move to the Credential Management spec.

index.html
```diff
@@ -435,7 +435,70 @@ <h2 class="informative">
...
+        [=verifier=] are the same entity or component.
+      </dd>
+    </dl>
+
+    <h3>Defined by other specifications</h3>
+    <ul>
+      <li>Holder: <a href="https://www.w3.org/TR/vc-data-model-2.0/#holder">W3C Verifiable Credentials Data Model v2.0</a></li>
+      <li>Issuer: <a href="https://www.w3.org/TR/vc-data-model-2.0/#issuer">W3C Verifiable Credentials Data Model v2.0</a></li>
+      <li>Verifier: <a href="https://www.w3.org/TR/vc-data-model-2.0/#verifier">W3C Verifiable Credentials Data Model v2.0</a></li>
+    </ul>
```
marcoscaceres: Handled automatically by #389. 

```suggestion
```

index.html
```diff
@@ -435,7 +435,70 @@ <h2 class="informative">
       request data=] and [=digital credential/request data|credential request
       data=] to and from such software.
       </li>
-    </ul><!--
+    </ul>
+    <!--
+    // MARK: Terminology
+    -->
+    <h2>
+      Terminology
```
marcoscaceres, line 443: so, I'd be open to moving all these to the Model section, and we could rename the Model section "Terminology". 

However, we should only include terms we are using in the spec, and also negotiate with the VC 2.0 folks on where each makes most sense... my gut feeling is that a lot of these are generic enough to be pushed to VC 2.0 unless there's a strong reason to have them in our spec (and we explicitly make use of them).

index.html
```diff
@@ -435,7 +435,70 @@ <h2 class="informative">
...
+        <dfn>Verification service</dfn>
+      </dt>
+      <dd>
+        An entity that performs the cryptographic validation of a 
+        [=verifiable presentation=] on behalf of the [=verifier=]. In many
+        cases, the [=verifier=] and verification service are the same entity or
+        component.
+      </dd>
+      <dt>
+        <dfn>Relying party</dfn>
```
marcoscaceres: This seems to contradict/redefine https://www.w3.org/TR/vc-data-model-2.0/#dfn-relying-parties

index.html
```diff
@@ -435,7 +435,70 @@ <h2 class="informative">
...
+      <dt>
+        <dfn>Issuance service</dfn>
+      </dt>
+      <dd>
+        An entity that performs the act of creating and delivering a verifiable
+        digital credential on behalf of the issuer. In many cases, the
+        [=issuer=] and issuance service are the same entity or component.
+      </dd>
+      <dt>
+        <dfn>Verification service</dfn>
```
marcoscaceres: This seems to redefine VC 2.0 [=verifier=]... I wonder if we should update [=verifier=] instead in VC 2.0 instead.

index.html
```diff
@@ -435,7 +435,70 @@ <h2 class="informative">
...
+      <dt>
+        <dfn>Wallet</dfn>
+        <dfn>Digital wallet</dfn>
+      </dt>
+      <dd>
+        A user friendly term for a [=credential manager=] for verifiable digital
+        credentials and other objects like payment cards and tickets.
+      </dd>
+      <dt>
+        <dfn>Issuance service</dfn>
```
marcoscaceres: As with the others, we should update [=issuer=] if needed in VC 2.0.

timcappalli: No. Issuance service is a distinct term, hence why it is defined separately.

index.html
timcappalli: No. Verifier service is a distinct term, hence why it is defined separately.

index.html
timcappalli: Not really, as they're not displayed in the terminology section.

index.html
timcappalli: No, I don't think it does.

index.html
msporny: FWIW, "issuer service" is defined here: https://w3c-ccg.github.io/vcalm/#dfn-issuer-service in a spec that's been in incubation for 5+ years.

index.html
msporny: FWIW, "verifier service" is defined here: https://w3c-ccg.github.io/vcalm/#dfn-verifier-service in a spec that's been in incubation for 5+ years.

We also make the distinction between the verifier coordinator (which runs business rules) a workflow service (which performs delivery, among other things), and a verifier service.

It looks like this group might be getting ready to rehash some hard won architectural and terminology debates that we've had over the past 5+ years in the VC groups.

index.html
timcappalli: Great! Thanks for pointing that out. Happy to update the reference when that document becomes a CR.

index.html
timcappalli: Great! Thanks for pointing that out. Happy to update the reference when that document becomes a CR.

Not trying to rehash anything :)

index.html
msporny: While broadly used, we've found the term "relying party" to be problematic because it's vague and uses language that's not immediately obvious to English speakers... "relying on what, specifically?"... the terminology also tends to conflate roles with entities (as the definitions in this section do). That was another hard won learning in the VCWG -- an entity (a person, an organization, etc.) can play many roles (issuer, holder, verifier). Many of the definitions of relying party conflate entities with roles, which leads to confusion.

We use the term verifier (a role) instead and stuck with that terminology which has led to better understanding on what the role does in the ecosystem and that an entity can take on a variety of roles based on the operation that they are performing.

index.html
msporny: @marcoscaceres What @timcappalli is noting is the same point I was trying to make a year or two ago (sorry, can't remember the issue where we were discussing that). People expect to see terms used in the spec, and their definitions, in the terminology section. An "import from external spec" could do this for us so people don't have to jump from spec to spec, but I also understand that requires work that many of us don't have the time for right now.

So, in the meantime, linking to the definition in another spec for terms and then summarizing which terms are defined in other specs somewhere (PR #389) is a reasonable compromise.

index.html
timcappalli: Model is now merged in.

index.html
```diff
@@ -456,34 +464,36 @@ <h2>
         [=claims=] made by an [=issuer=] about one or more [=subjects=].
         <p class="note" title="Focus on digital credentials about people">
           This specification is currently focused on digital credentials
-          pertaining to people.
+          pertaining to people. Also known as a verifiable digital credential.
```
MasterKale, line 467: NIST coined "verifiable digital credential", no? Would it make sense to link to e.g. https://www.nist.gov/blogs/cybersecurity-insights/digital-identities-getting-know-verifiable-digital-credential-ecosystem to help connect some dots?

index.html
```diff
@@ -456,34 +464,36 @@ <h2>
...
       <dt>
-        <dfn data-dfn-for="digital credential" data-local-lt=
-        "presentation">Presentation request</dfn>
+        <dfn>Digital wallet</dfn>
       </dt>
       <dd>
-        A presentation request is a request for a [=digital credential=]
-        composed of [=digital credential/request data=] and a [=digital
-        credential/exchange protocol=].
+        A user friendly term for a [=credential manager=] for verifiable digital
```
MasterKale, line 474: Just a nitpick, take it or leave it:
```suggestion
        A user-friendly term for a [=credential manager=] for verifiable digital
```

index.html
```diff
@@ -456,34 +464,36 @@ <h2>
...
+        credentials and other objects like payment cards and tickets. Also known
+        as a "wallet" or "identity wallet".
       </dd>
       <dt>
-        <dfn data-dfn-for="digital credential">Request data</dfn>
+        <dfn data-dfn-for="digital credential" data-local-lt=
+        "exchange protocol">Exchange protocol</dfn>
```
MasterKale, line 480: Has the ship sailed on our chance to call this "Presentation Protocol"? It's unfortunate we have

**Issuance Protocol > Issuance Request > Issuance Response**

but

**Exchange Protocol > Presentation Request > Presentation Response**

index.html
```diff
@@ -510,15 +520,39 @@ <h2>
...
-        A standardized protocol used for exchanging a [=digital credential=]
-        between a [=holder=] and a [=verifier=]. A protocol is identified by a
-        [=digital credential/protocol identifier=]. See section also
-        [[[#protocol-registry]]].
+        A format that a [=holder's=] software, such as a digital wallet, uses,
+        via an [=digital credential/exchange protocol=], to respond to a
+        [=digital credential/presentation request=] by a [=verifier=].
+      </dd>
+      <dt>
+        <dfn>Relying party</dfn>
```
MasterKale, line 541: What if we also defined "Verification service" as an alias for Relying Party? "Issuers" have an "Issuance service", so why not define something for "Verifiers" as well?

I can't stop proposing parity between paired terms for **presentation** and **issuance** concepts because, in this context, they are arguably two sides of the same coin. I think it'd be helpful to simultaneously define analogous terms, assuming analogous concepts exist on both sides. In this case it's the concept of "the thing that performs the Issuer or Verifier action, that may also be the Issuer or Verifier respectively but allows for third-parties to take over that responsibility."

timcappalli: > "Issuers" have an "Issuance service", so why not define something for "Verifiers" as well?

There is a verifier service in this PR. It is not always the same as relying party.

MasterKale: Oh, huh, you're right, it's at the bottom. I was jumping around the preview a bit too much perhaps. Please disregard.

index.html
marcoscaceres: We should decide who owns it. Should the DC API own it or can we move to VC 2 instead of the incubation? 

(We shouldn’t move definitions around once we decide where they live, as that would break/change dependencies for other specs… we already depend heavily on VC 2, so my preference is we push as much as possible to VC 2 as that community owns most of these terms already and is best poised to maintain these definitions)

index.html
marcoscaceres: Agree with @msporny. We shouldn’t define this and just rely on verifier (and maybe in a note say that this is sometimes called a “relying party” if not already stated).

index.html
```diff
@@ -435,19 +435,27 @@ <h2 class="informative">
...
     </h2>
-    <p class="note" title="Definitions under discussion">
-      The goal of the definitions in this section is to reuse or establish
-      terminology that is common across a variety of digital credential formats
-      and protocols. Discussions surrounding these definitions are active and
-      the definitions are likely to change over the next several months.
-    </p>
-    <dl class="definitions" data-sort="" data-cite="vc-data-model-2.0">
+    <h3>Defined by this specification</h3>
+    <!--
```
marcoscaceres, line 446: We rely on VC 2 tho for terminology? Is the VC WG in agreement with this? We should coordinate with them on who owns the canonical definitions. 

Ideally, IMO, these should be owned by the W3C, as not everyone participates in the IETF.

index.html
marcoscaceres: The W3C requires that we cite normative documents, so citing a blog post is not appropriate.

index.html
marcoscaceres: We can change it. It doesn’t affect the API surface.

marcoscaceres: @MasterKale, would you mind filling a new issue and sending a separate PR?

index.html
marcoscaceres: > It looks like this group might be getting ready to rehash some hard won architectural and terminology debates that we've had over the past 5+ years in the VC groups.

Absolutely not! Why I asked for your input. The VC groups are in the best position to guide us here. Why I reached out… precisely so we don’t do that 🙏

index.html
marcoscaceres: That’s not how other browser API specs work though (see any browser spec, apart from the exceptional few that didn’t get the memo)… what’s beautiful about the this “web” thing is the you can seamlessly jump around documents through “hyper links” 😝 

ReSpec, Bikeshed, and the ton of infrastructure/tools we’ve at the W3C is designed around this - unapologetically opinionated so. Redefining things is redundant and highly discouraged by the tooling by design and unnecessary given hyperlinks. 

Best practice is to define things once and just link. Terminology sections that just link to other specs are superfluous and require the user to click twice to get to the definition they want. Basically, “I saved you a click”.

index.html
msporny: Referring the WG to more hard won discussions around the terminology of workflows, exchanges, and protocols: https://w3c-ccg.github.io/vcalm/#workflows-and-exchanges

index.html
msporny: Yes, agreed @marcoscaceres -- I know what you're going for. My comment was more about what @timcappalli was suggesting, though he too seems to note that references will be updated when VCALM goes standards track (which is expected for the next VCWG recharter to happen this year).

At the very least, please do take a look at the definitions and try to copy as much as possible so we don't diverge.

index.html
msporny: > That’s not how other browser API specs work though

Yah, they're all broken. :P (only half-joking)

> Redefining things is redundant 

Yes, completely agree. 

> Terminology sections that just link to other specs are superfluous

Agree.

> Best practice is to define things once and just link.

Jumping around from document to document is the annoying thing. As the person writing a spec, can't you just show me a definition inline, or at least in the same document? Don't redefine it, just import the definition or (better) just show the definition in an overlay in the spec. I'm rehashing the discussion we had a year or so ago at this point.

When I'm reading a sentence with 5 new terms that I'm not quite sure what they mean, I don't want to have to divert to 5 different specifications to learn about those words.

In any case, this is non-blocking, wish list feedback. If we have to jump to another spec, that's fine for now, though, as a reader, it's annoying.

index.html
marcoscaceres: @msporny anything we can bubble up to a published doc?... or plan to move that other doc out of incubation?

index.html
```diff
@@ -530,15 +564,31 @@ <h2>
...
+        cases, the [=verifier=] and verification service are the same entity or
+        component.
+      </dd>
+    </dl>
+
+    <h3>Defined by other specifications</h3>
+    <ul>
+      <li>Holder: <a href="https://www.w3.org/TR/vc-data-model-2.0/#holder">W3C Verifiable Credentials Data Model v2.0</a></li>
+      <li>Issuer: <a href="https://www.w3.org/TR/vc-data-model-2.0/#issuer">W3C Verifiable Credentials Data Model v2.0</a></li>
+      <li>Verifier: <a href="https://www.w3.org/TR/vc-data-model-2.0/#verifier">W3C Verifiable Credentials Data Model v2.0</a></li>
```
marcoscaceres, line 589: This is now covered by the index. 

```suggestion
```

index.html
msporny: We could move some of those definitions over to VC Data Model v2.1, which is actively being published. If we're in a hurry, we could push VCDM v2.1 into CR if that's the requirement for DC API to have "stable definitions".

The current plan is to move that other doc (VCALM - Verifiable Credential API for Lifecycle Management) out of incubation in the next W3C VCWG recharter expected in early 2026, which is being discussed at W3C TPAC in less than a month. The concepts in the doc are fairly mature (having been incubated for 5+ years at this point, [21 basic implementations](https://github.com/w3c/vc-test-suite-implementations/tree/main/implementations) and about 4 almost complete implementations). It outlines an architecture for verifiable credentials that incorporates all the various digital credential formats (mDL, mDoc, VCs, etc.) and protocols (OID4VCI, OID4VP, VC API workflows/exchanges, proprietary protocols, etc.,). Not complete by any stretch of the imagination, but again, some hard won consensus among the implementers of that specification.

index.html
msporny: Not everyone in the ecosystem believes that "verifiable digital credentials", as defined by NIST, is a good definition. I certainly don't. If asked, some would argue against it vehemently because it adds yet another term that doesn't add much value and whose definition is a bit off. For example: You can add a digital signature to a digital credential to make it verifiable, but that's not the only way... however the NIST definition notes that cryptography seems to be the only way (which is not correct).

The DC API can pass "digital credentials" that might not be verifiable. Verifiability is a quality of some digital credentials, but not others. I know that we're focused on digital credentials that are verifiable here, but ones that are not have utility too (e.g., form auto-fill and delivery of legacy PDFs).

The "Digital Credential API" spec got it right by just calling these things "digital credentials" -- that's the format agnostic word for these blobs of data. Sure, you can call that thing a "verifiable digital credential", but if you're going to insist calling that thing something special, you should probably rename "DC API" to the "VDC API". Whatever is done, be consistent with the naming because it's going to confuse people if you aren't... and please, avoid adding new terms that add little to understanding the ecosystem.

Requiring a reader to look through a glossary before understanding a plain English sentence is tiresome... yes, we did pass that point a while ago, but let's not make the problem worse. We should endeavor to use plain English words if possible so that a casual reader will usually jump to the right conclusion when they read a particular term.

index.html
marcoscaceres: > If we're in a hurry, we could push VCDM v2.1 into CR if that's the requirement for DC API to have "stable definitions".

That sounds great, @msporny. Even in CR, if these are just definitions that don't affect implementations, it should be possible to update the live CR document (depending on the VC working group's processes). 

OK, let's work with @timcappalli and get consensus on who owns what and what should be defined where. 

@MasterKale can you align with what @msporny suggested above if needed so we have the same terminology across vcalm? 

(It’s taking all my energy refraining from making any "be calm" jokes. 🤭)

index.html
marcoscaceres: > Don't redefine it, just import the definition or (better) just show the definition in an overlay in the spec. I'm rehashing the discussion we had a year or so ago at this point.

You are spot on! And yes!!!! @dontcallmedom, @tidoust, and I thought about adding this to the tooling a while back, so we do the same thing as Wikipedia does:

using a pop over

There's a few challenges, like where specs where spec authors were not using the dt + dd pattern correctly (looking at you Web Authn), but I think we can overcome those issues now using AI to do the extraction without bothering Editors.

index.html
```diff
@@ -435,19 +435,27 @@ <h2 class="informative">
...
+      Terminology
     </h2>
-    <p class="note" title="Definitions under discussion">
-      The goal of the definitions in this section is to reuse or establish
-      terminology that is common across a variety of digital credential formats
-      and protocols. Discussions surrounding these definitions are active and
-      the definitions are likely to change over the next several months.
-    </p>
-    <dl class="definitions" data-sort="" data-cite="vc-data-model-2.0">
+    <h3>Defined by this specification</h3>
```
marcoscaceres, line 445:
```suggestion
```

index.html
```diff
@@ -456,34 +464,36 @@ <h2>
...
         <p class="note" title="Focus on digital credentials about people">
           This specification is currently focused on digital credentials
-          pertaining to people.
+          pertaining to people. Also known as a verifiable digital credential.
         </p>
       </dd>
       <dt>
-        <dfn data-dfn-for="digital credential" data-local-lt=
-        "presentation">Presentation request</dfn>
+        <dfn>Digital wallet</dfn>
```
marcoscaceres, line 471: Let's work on this separately in #386. The definition there more closely aligns with how implementations currently work, per discussion with @mohamedamir. We can expand that one a bit if we need to.

index.html
tidoust: Linking to the underlying Respec issue and current status: https://github.com/speced/respec/issues/4522#issuecomment-1933574282 (summary: the logic already exists at the data level, but isn't used for now)

index.html
msporny: > It’s taking all my energy refraining from making any "be calm" jokes.

Please, make those jokes -- sounding like "BE CALM!" is one of the unwritten reasons some of us voted for the name... 'cause everyone's being a bit high strung about all this digital credentials stuff... and maybe we just all need to be calm about it... and as we know, telling someone else how to behave in shouty-caps is always a winning strategy. :P

index.html
msporny: >  thought about adding this to the tooling a while back

Excellent, then we're aligned on the goal, now it's just a simple matter of some angel finding the time to implement it, and based on @tidoust's feedback, we're closer than we were a year ago. In the meantime, we can link to other specs as you noted and folks will have to click through to the other spec to get to the definition.

> I think we can overcome those issues now using AI to do the extraction without bothering Editors.

Bonus points for working AI into the conversation, that is, if you are indeed the real @marcoscaceres.

index.html
```diff
@@ -435,19 +435,27 @@ <h2 class="informative">
...
+    <dl class="definitions" data-sort="ascending" data-cite="">
+      <dt>
+        <dfn>Credential manager</dfn>
+      </dt>
+      <dd>
+        An application, hardware device, or service which securely stores,
+        organizes, manages, and enables presentation of credentials. Digital
+        wallets, password managers, and passkey managers are all examples of
+        credential managers.
+      </dd>
```
marcoscaceres, line 458: Moved here: https://github.com/w3c/webappsec-credential-management/pull/275

```suggestion
```

marcoscaceres: Merged:
https://github.com/w3c/webappsec-credential-management/pull/275

Thanks for the suggested text.

marcoscaceres: Merged.

index.html
MasterKale: Sorry for the delay, Authenticate and PTO ate up my last two weeks. I'm ramping back up on this stuff and will prioritize addressing this thread this week 🙇

marcoscaceres: Amazing! thanks @MasterKale. Please send a separate PR. 🙏

MasterKale: Digging through [VCALM v0.9](https://w3c-ccg.github.io/vcalm/), I'm not seeing anything in there that'd help us make progress here to **renaming "Exchange Protocol" to "Presentation Protocol"** to align with "Presentation Request" and "Presentation Response." VCALM also calls it "Exchange Protocol" which I continue to suggest is a missed opportunity to have more consistent terminology across related words:

Screenshot 2025-10-29 at 2 14 50 PM

And nothing in [VC Data Model 2.1](https://w3c.github.io/vc-data-model/) right now seems to define protocol-specific terminology either; I understand this is was why it was suggested that things from VCALM could be cherry-picked into VCDM 2.1.

Right now given what I mentioned above I don't think any effort needs to be made to move stuff from VCALM into VCDM unless those working on VCALM agree that "Exchange Protocol" makes more sense to be called "Presentation Protocol."

MasterKale: Can we simply alias "Presentation Protocol" to "Exchange Protocol" in our spec, then use "presentation protocol" throughout DC API?

marcoscaceres: That seems confusing… we usually only alias short hands (e.g., “wallet” to be “digital wallet”). 

VCALM is still an incubation, so maybe not too late to unify everything? @msporny?

msporny: Yes, let's please try to do some unification -- aliasing "presentation protocol" to "exchange protocol" would most probably be a mistake. For example, in the VCALM spec, a presentation is something that can happen during an exchange (these terms are not interchangeable). Another protocol that can happen during an exchange is issuance, with the most typical flow being a presentation protocol followed by an issuance protocol all within the same exchange.

hlflanagan:
Discussed during [25 November 2025](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-11-26-DCAPI-B-notes.md) call. Will bring back to the Series A call to make sure we're ready to break this into one PR per term being defined.

hlflanagan:
Discussed briefly https://github.com/w3c-fedid/meetings/blob/main/2025/2025-12-02-DCAPI-A-notes.md#add-terminology-section-merge-in-model-387. @timcappalli holds the pen

PR #204, opened 2025-02-19
Include Digital Credential Issuance in the Spec

mohamedamir:
Closes #167 

The following tasks have been completed:

- [ ] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/51492/)

Implementation commitment:

- [ ] WebKit (link to issue)
- [x] Chromium ([link to issue](https://crbug.com/378330032))
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer
- [ ] Updated digitalcredentials.dev

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/204.html" title="Last updated on May 1, 2025, 8:31 AM UTC (112e275)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/204/5592be0...112e275.html" title="Last updated on May 1, 2025, 8:31 AM UTC (112e275)">Diff</a>

PR #401, opened 2025-12-01
Change from Registry to Protocols

### LINKED ISSUE #69: Add examples of usage in registry
Once we have the something in the registry #58, we should add examples for each. We could make providing an example part of the registry inclusion criteria.

### LINKED ISSUE #136: Describe trust signals in protocol registry

### LINKED ISSUE #211: Add OpenID for Verifiable Presentations (OpenID4VP) to registry
Include OpenID for Verifiable Presentations in the presentation protocol registry.

Current draft: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html

### LINKED ISSUE #212: Add OpenID for Verifiable Credential Issuance (OpenID4VCI) to registry
Include OpenID for Verifiable Credential Issuance in the issuance protocol registry.

Current draft: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html

### LINKED ISSUE #213: Add ISO/IEC 18013-7 Annex C to protocol registry
Include ISO/IEC 18013-7 Annex C in the presentation protocol registry.

Current draft:   https://github.com/ISOWG10/ISO-18013/blob/main/Working%20Documents/Working%20Draft%20WG%2010_N2544_18013-7%20second%20edition%20DTS%20candidate%20-%20NTC.pdf

### LINKED ISSUE #214: [Registry] Re-evaluate "freely available" requirement
As per the 2025-04-11 hybrid meeting discussion, this issue is to track future re-evaluation of the "freely available" requirement in the registry inclusion criteria as discussions between W3C and ISO continue and when ISO 18013-7 Annex C meets the other requirements to be added to the registry (protocol identifier, etc).

### LINKED ISSUE #215: Document WG security review requirements for protocols
>> "MUST have undergone security review by the Federated Identity Working Group."

> @bvandersloot-mozilla: Do we want to define the criteria by which this group would review those?

One comment from the registry criteria PR is that the spec needs to define the criteria for the security review. Where in the spec this should go is TBD and we should discuss here...

### LINKED ISSUE #240: Registry requirements should not use RFC2119 language
Sections that are informative shouldn't be using normative language, unless we make the language forceful, but is can't be MUST/SHOULD/MAY etc.

### LINKED ISSUE #256: Protocol registry: royalty free requirement
The current inclusion criteria for protocols says "MUST be standardized at a [consortium the W3C liaises with](https://www.w3.org/liaisons/)".  This might be a useful criterion, though I can't imagine why it is phrased that way.  The thing I care about here is less about who standardized a protocol, but the characteristics of the protocol itself.

One of those characteristics is already listed: "MUST be defined in a specification which is freely and publicly available".  This is good, with a small caveat (issue forthcoming).  The other is IPR encumbrance.

The W3C has a requirement that implementations of its specifications be free from intellectual property encumbrance that would incur royalties: https://www.w3.org/policies/patent-policy/#sec-Requirements

This API should include the same requirement.  I would prefer that include all components (browser, wallet, website) as that would be most consistent with the intent of the W3C patent policy.

That doesn't necessarily negate any requirement about who the W3C liaises with, but I think that the liaison term is probably a proxy for a more important requirement.

### LINKED ISSUE #257: Protocol registry: specification availability
The inclusion criteria lists "MUST be defined in a specification which is freely and publicly available" as a requirement.  This is good, but it is not sufficient as stated.

I request that this be extended to transitively include all normative dependencies of the referenced specification.  Having chased other specifications into a maze of paywalls, this is a genuine problem.

### LINKED ISSUE #345: Is a registry needed?
In #288, I raised an issue arguing against registries in general. @npdoty suggested I create a new issue for that discussion. I've combined my two comments over there to frame the fundamental question about a registry.

Registries create overhead, security risks, and barriers to entry.

Without a compelling argument why these problems are worth it, I'd argue the downsides outweigh the benefits.

Please discuss.

----

As I brought up at the GDC session on barriers to global adoption, I would prefer to see the registries removed entirely.

A few points:

# Registries

1. Act as a barrier to adoption 
2. Squelch innovation.
3. Propagate Western technocolonialism
4. Don't actually improve the security of the system
5. Put the W3C in an inappropriately elevated position
6. Are primarily a first-mover advantage that controls and limits adoption, rather than advance it
7. Create a novel and unnecessary attack surface

# Registries act as a barrier to adoption
It is inevitable that any registry we make will not be available in all the languages in the world. This puts anyone who is not a native English speaker at a disadvantage. This is a known problem for W3C specifications: it's hard to maintain localized versions. Further, if the information to be placed in the registry MUST be in English, that is also a barrier to those developers who work in a different language context. IMO, developers should be free to innovate in their own native tongue and use that technology regardless of whether or not they can convince the W3C to add it to a registry (for whatever reason they might be denied, including using a name that someone else also uses).

# Registries squelch innovation
There's a reason there isn't a required registry of websites. As TBL stated in his first publication describing the web as a whole, https://cds.cern.ch/record/369245/files/dd-89-001.pdf he explicitly called out the importance of NOT restricting the nature of the content on the web:
> the method of storage must not place its own restraints on the
information.

and

> The system must allow any sort of information to be entered. Another
person must be able to find the information, sometimes without
knowing what he is looking for.

and

> **Non-Centralisation**
Information systems start small and grow. They also start
isolated and then merge. A new system must allow existing
systems to be linked together without requiring any central
control or coordination.

and 

> The only way in which sufficient flexibility can be incorporated is to
separate the information storage software from the information
display software, with a well defined interface between them. Given
the requirement for network access, it is natural to let this clean
interface coincide with the physical division between the user and the
remote database machine3.
This division also is important in order to allow the heterogeneity
which is required at CERN (and would be a boon for the world in
general).

In short, a registry creates restrictions, which goes against the very intent of the World Wide Web, which was specifically designed to allow massive innovation without centralized authorities vetting each new feature or content. This will limit adoption.

# Registries propagate western Technocolonialism
This is a well-known point of historical contention. Registries for the web, specifically DNS and IANA IP blocks, have been widely criticized for enshrining a particularly western, and even American position of privilege. For example, the US got to decide that all .gov domains are for the US government, and initial DNS specification couldn't handle non-western text. Any system that defers to a single registry run by a western-biased organization *will* propagate the kind of technocolonial perspective that places non-Western, non-English speaking, non first-moving parties at a specific, structural disadvantage that is not easily remediated. This will limit adoption.

# Registries don't actually improve the security of the system.
While a registry *could* attempt to vet any particular proposal, there is no guarantee that (a) the vetting is thorough, correct, and still valid, nor (b) that any given implementation of that protocol is implemented in a way that realizes the intended security guarantees. At best, it blesses "well-known" and popular technologies at the expense of perhaps more secure, more trustworthy alternatives. The fact that a protocol is listed in a registry is likely to create a false believe that it is of high quality and that any implementations claiming to use that protocol are also, therefore, somehow trustworthy. In reality, implementers considering which protocols to support should investigate the protocols themselves with their own concerns to see which best fit their needs. For any given use case, any particular protocol might be grossly inappropriate; imo, there simply is no way for a central organization to effectively vet all the possible technologies that might be useful. This will create a false sense of security rather than invite a healthy skepticism.

# Registries put the W3C in an inappropriately elevated position
Although all of us here are, necessarily, supportive of the W3C as a standards development organization (or we wouldn't be here having this debate), there is no reason to believe that the W3C has the capability and moral authority to actually stay on top of real-time vetting of protocols as imagined by this specification. The W3C is not a cybersecurity response agency. We simply don't have the operational processes in place to maintain an appropriate posture with regard to the security of any given protocol. Even if a protocol were reasonably well-vetted by security minded professionals, there is no infrastructure or processes to maintain that assessment over time. The fact is, the W3C is just one of many organizations that have opinions about the quality of different protocols. Placing our own assessments over those of other, often more qualified, organizations would do a substantial disservice to the actual security of DCAPI implementations, by inducing trust because a protocol is listed in the registry despite the inherent inability for the W3C to stay on top of emerging cyber threats (at least with its current operational funding an processes). Best case, the W3C can advocate for those protocols its members approve of, but asserting that such a registry is any indicator of deployed security quality would be a gross misrepresentation.

# Registries are primarily a first-mover advantage that controls and limits adoption, rather than advance it
Registries fundamentally keep things not in the registry in a relegated position. Whatever rules one must satisfy to get into the registry comprise real and unavoidable barriers to anyone not already in the privileged club. If you take the misguided DID Method registry, you'll see what I mean. In two separate incidents, the registry maintainers relied on flawed analysis that forced real innovators to adopt lesser alternatives than simply calling their DID method what they wanted. In one case, it was an ineffective policy about who actually controls an entry. In another, it was a flawed legal argument that allowed a community member to shout down a legitimate entry without effective due process that actually engaged the rule of law (it was an alleged trademark matter). While both of these were eventually resolved (both in a flawed manner, imo), they illustrate the fundamental function of the registry was NOT to embrace innovation, but rather to contain it, to limit it, not advance it. This will limit adoption and usefulness of the DCAPI.

In fact, any registry that is not a listing of requirements, e.g., protocols that are mandatory to implement, will only be able to *exclude* unlisted protocols, since inclusion in the registry won't ensure that any implementation actually supports any given protocol. With that lens, the only function of the registry is to limit the set of supported protocols.

# Registries create a novel and unnecessary attack surface
Once you rely on registries, those registries become their own attack surface for security considerations. It will be possible

a. to get a spec listed in the registry inappropriately 
b. for entries in the registry to be inappropriately removed
c. for entries in the registry to be inappropriately updated

A strong set of institutional processes can minimize the impact of these attacks, but as long as there is a registry, it, by itself, creates threats that can be entirely avoided simply by designing without a registry.

IMO, all we actually need are mechanisms for which components can self-identify the protocols they support and the browser deals which matching both ends of the credential exchange as a dumb pipe, supporting whatever *can* be supported in the architecture. Which is to say, I fully realize that a deployed piece of software is not going to randomly suddenly be able to support a new protocol, but the decision to support any protocol is far less the purview of the W3C as much as the *responsibility* of implementers. In fact, I expect there *will* eventually emerge meta-protocols that allow the equivalent of media codecs to  be downloaded on the fly to support additional functionality. Any kind of restrictive registry would make that sort of dynamic flexibility functionally impossible.

There is sometimes an argument raised that without a registry, you *can't* know what protocol someone is actually referring to. This problem as solved rather elegantly with GUIDs, used by Microsoft to securely identity COM interfaces *created by anyone* without a centralized registry. Yes, there were directories of a sort which could give you the GUIDs for commonly used interfaces, but the technology **worked**: it enabled *any* programmer to define their own COM interface that immediately had interoperable semantics (for identification of the COM interface) so it could be used by any COM-enabled software that knows to look for that particular interface.

I would suggest considering ways that we can remove the registries entirely, all of them.

----

In my second comment

----

> Do you have any suggestions of replacements for the existing registries of which you are aware, at W3C or elsewhere?

Yes,. First, I'd separate this into two functions.

First, for run-time matching, use GUIDs or DIDs or even URIs (RDF style) to allow components to self-describe using globally unique identifiers (which painlessly refactors to URIs without needing to distinguish nuance between URIs, URNs, and URLs). This includes formats that self describe, as well as protocols. The developers of components simply publish the globally unique identifier for their component in their specification and developers decide at dev time which components they want to support, then implementations communicate which components they support at run time.

Second, for discovery, use directories, search engines, and LLMs.  Developers, at dev time, are going to use the web and associated mechanisms to learn about interesting components, just as we do today for nearly all technology we incorporate in our applications. The thing about these kinds of services is that there are no gatekeepers for the meta platform: each directory, search engine, or LLM gets to decide which components it indexes. Reciprocally, developers can use any discovery tool to find the technology they are interested in, without deference to a systemic limitation on the platform. Decentralized discovery is how the Web works. Let's keep using that rather than impose the W3C as a gatekeeper for systems built on its technology.

FWIW, I don't consider automated discovery to be a problem worth solving as streamlining automated downloads of untrusted algorithms without human evaluation is a known security problem. Manual discovery through the web helps ensure that developers are in the loop when incorporating components. Perhaps more to the point: I don't think the convenience of automated discovery is worth the centralizing costs, especially in light of the increased security risks of adding that capability.

> I wonder whether any (probably not all) of your objections might be solvable by adjustments to those registries?

First, turning them into directories for discovery would be a big help. There's nothing wrong with a spec listing related technologies as an aid to developers. It's when you make it a "registry" that leads to conversations like this, where people expect registration to be required and hence, it seems reasonable to make listing in the registry have normative requirements, which is what is proposed in this issue. 

Second, remove any requirements other than editorial review for inclusion, making it clear that the list is a *curated* list maintained by specific editors, and not an official registry that has normative impact. I don't mind if the editors have opinions, but I do find it challenging when rules are imposed on the registry to exclude "bad" listings because they don't satisfy some obscure, political, or ideological reason and hence establish a false notion of approval by the W3C as a whole, rather than the independent opinion of respected editors.

### LINKED ISSUE #396: Paths towards resolution on the registry [TPAC]
Based on the discussion on the 2025-10-29 call (which was based on all of the previous discussions), this is an attempt to map out potential paths forward, to discuss at TPAC 2025.

This list is deliberately exhaustive.

## Current State

- The spec has a protocol registry, and each protocol must go through W3C privacy and security review to be added to the registry, as defined in the spec
- The security and privacy requirements in the spec are closely tied to the registry.

## Option 1

- Drop registry from the spec completely
- Change current security and privacy requirements that map to the registry, to be protocol implementation _recommendations_

## Option 2

- Drop registry from the spec completely
- Change current security and privacy requirements that map to the registry to be protocol implementation **requirements** (which can be used by other organizations as part of certification)

## Option 3

- Change the registry to a "list of known protocols" (protocol identifier and spec link) with no normative requirements outside of avoiding identifier collisions
- Change current security and privacy requirements that map to the registry to be protocol implementation _recommendations_

## Option 4

- Change the registry to a "list of known protocols" (protocol identifier and spec link) with no normative requirements outside of avoiding identifier collisions
- Change current security and privacy requirements that map to the registry to be protocol implementation **requirements** (which can be used by other organizations as part of certification)

## Option 5

- Change the registry to a "list of known protocols" (protocol identifier and spec link) with no normative requirements outside of avoiding identifier collisions
- Change the current security and privacy requirements to only talk about considerations for the API surface itself
- Move ecosystem / protocol security and privacy requirements into its own document/spec

## Option 6

- Drop registry from the spec completely
- Change the current security and privacy requirements to only talk about considerations for the API surface itself
- Move ecosystem / protocol security and privacy requirements into its own document/spec

## Option 7

- Drop registry from the spec
- Normatively reference OpenID4VP 1.0 (signed, unsigned, multisigned), OpenID4VCI, and ISO 18013-7 Annex C in the spec
- Requests with unsupported protocols are ignored

## Related issues

- #345 
- #288 
- #280 
- #266 
- #257 
- #256 
- #255 
- #240 
- #226 
- #215 
- #214 
- #213 
- #212 
- #211 
- #136 
- #69

timcappalli:
This PR makes the following changes:

- Removes all references to a registry
- Adds a new Supported Protocols section, with a table for details about each protocol, and enums  containing the protocol identifiers for presentation and issuance
- Updates intro text to match new pattern
- Updates userAgentAllowsProtocol() method to use enum
- Updates the GetRequest dictionary to use enum
- Updates all references to registry to the new section

Closes #396 
Closes #345
Closes #257
Closes #256
Closes #240
Closes #215
Closes #214
Closes #213 
Closes #212
Closes #211
Closes #136
Closes #69

- [x] Modified Web platform tests: https://github.com/web-platform-tests/wpt/pull/56627/changes

Implementation commitment:

- [x] [WebKit](https://github.com/WebKit/WebKit/pull/55603/) + https://github.com/WebKit/WebKit/pull/57027
- [x] Chromium ([link to issue](https://issues.chromium.org/issues/465006289))
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/401.html" title="Last updated on Jan 26, 2026, 5:22 PM UTC (98ed1e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/401/7ad0be8...98ed1e1.html" title="Last updated on Jan 26, 2026, 5:22 PM UTC (98ed1e1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -114,7 +114,8 @@
     </p>
     <p>
       The API design is agnostic to credential formats and includes support for
-      multiple [=digital credential/exchange protocols=]. See [[[#protocols]]].
+      multiple [=digital credential/exchange protocols=] and [=digital
+      credential/issuance protocols=]. See [[[#protocols]]].
     </p>
     <p>
       The API is designed to support the following goals:
@@ -563,15 +564,15 @@
       Protocols
     </h2>
     <p>
-      Use of the following presentation and issuance protocols are defined by this specification.
-      <!-- TODO: replace with new dfn from 434" -->
+      Use of the following presentation protocols is defined by this specification.
+      <!-- TODO: replace with new dfn from 434 -->
     </p>
     <aside class="issue" data-number="439">
     </aside>
     <table class="data">
       <caption>
         <dfn class="export">Table of supported presentation and [=digital credential/issuance protocol|issuance=] protocols</dfn>
-        <!-- TODO: replace with new dfn from 434" -->
+        <!-- TODO: replace with new dfn from 434 -->
       </caption>
       <thead>
         <tr>
@@ -595,7 +596,7 @@
             [[[OPENID4VP]]] - Unsigned Requests
           </td>
           <td>
-            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+            Presentation Protocol <!-- TODO: replace with new dfn from 434 -->
           </td>
           <td>
             <dfn data-dfn-for="DigitalCredentialPresentationProtocol">`openid4vp-v1-unsigned`</dfn>
@@ -609,7 +610,7 @@
             [[[OPENID4VP]]] - Signed Requests
           </td>
           <td>
-            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+            Presentation Protocol <!-- TODO: replace with new dfn from 434 -->
           </td>
           <td>
             <dfn data-dfn-for="DigitalCredentialPresentationProtocol">`openid4vp-v1-signed`</dfn>
@@ -623,7 +624,7 @@
             [[[OPENID4VP]]] - Multi-signed Requests
           </td>
           <td>
-            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+            Presentation Protocol <!-- TODO: replace with new dfn from 434 -->
           </td>
           <td>
             <dfn data-dfn-for="DigitalCredentialPresentationProtocol">`openid4vp-v1-multisigned`</dfn>
@@ -637,7 +638,7 @@
             [[[ISO18013-7]]] - Annex C
           </td>
           <td>
-            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+            Presentation Protocol <!-- TODO: replace with new dfn from 434 -->
           </td>
           <td>
             <dfn data-dfn-for="DigitalCredentialPresentationProtocol">`org-iso-mdoc`</dfn>
@@ -966,7 +967,7 @@
     </p>
     <p>
       The {{DigitalCredentialCreateRequest/protocol}} member's value is one
-      of the protocol identifiers defined in [[#protocols]].
+      of the protocol identifiers defined in [[[#protocols]]].
     </p>
     <h4>
       The `data` member
@@ -1061,6 +1062,9 @@
       algorithm:
     </p>
     <ol class="algorithm">
+      <li>If |protocol| is not a member of {{DigitalCredentialPresentationProtocol}},
+      return `false`.
+      </li>
       <li>Return `true` if the user agent allows |protocol|, otherwise return
       `false`.
       </li>
@@ -1076,7 +1080,7 @@
     </h4>
       <p>
         This enumeration's values correspond to the supported
-        presentation protocols listed in [[[#protocols]]]. <!-- TODO: replace with new dfn from 434" -->
+        presentation protocols listed in [[[#protocols]]]. <!-- TODO: replace with new dfn from 434 -->
       </p>
       <pre class="idl">
       enum DigitalCredentialPresentationProtocol {


index.html
```diff
@@ -84,6 +84,26 @@
...
+          authors: ["Torsten Lodderstedt", "Kristina Yasuda", "Tobias Looker", "Paul Bastian"],
+          date: "2025-09-16",
+          publisher: "OpenID Foundation"
+        },
+        "iso-18013-7-2025": {
+          title: "ISO/IEC TS 18013-7:2025 Personal identification — ISO-compliant driving licence Part 7: Mobile driving licence (mDL) add-on functions",
+          href: "https://www.iso.org/standard/91154.html",
+          date: "2025-05-01",
+          publisher: "International Organization for Standardization"
+        },
```
marcoscaceres: Let's add this to [specref](https://github.com/tobie/specref), as they are published standards:

```suggestion
```

@copilot, are you able to send a pull request to specref for these, or do we need to do it directly in the other repository?

index.html
```diff
@@ -544,15 +559,105 @@ <h2>
...
       <dd>
         See [=credential request coordinator=].
       </dd>
-    </dl><!--
+    </dl>
+    <!--
+    // MARK: Supported Protocols
+    -->
+    <h2 id="supported-protocols">
+      Supported protocols
```
marcoscaceres: I think we should avoid the word "Supported" as passive voice begs the question "supported by who?". Also, it's not clear what "supported by this specification" means. 

This should really be about, "protocols that are known to be allowed by user agents" as verified by `userAgentAllowsProtocol()` for instance. 

We could even have a column(s) for user agent or engine (WebKit | Gecko | Chromium).

index.html
```diff
@@ -544,15 +559,105 @@ <h2>
...
+            [[[openid4vp-1_0]]] (Unsigned Requests)
+          </td>
+          <td>
+            [=digital credential/Exchange Protocol=]
+          </td>
+          <td>
+            `openid4vp-v1-unsigned`
+          </td>
+          <td>
+            Appendix A
```
marcoscaceres: This would be great to link directly if possible.

index.html
```diff
@@ -767,9 +872,8 @@ <h4>
       denotes the [=digital credential/exchange protocol=].
     </p>
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value can be one
-      of the well-defined protocol identifiers defined in
-      [[[#protocol-registry]]] or a custom protocol identifier.
+      The {{DigitalCredentialCreateRequest/protocol}} member's value MUST be one
```
marcoscaceres: This basically makes it an enum... if we put MUST here, then we need to also specify what happens when the value is not one of these (i.e., be mindful when this is user agent requirement or an authoring requirement).

index.html
```diff
@@ -841,9 +945,8 @@ <h4>
       member denotes the [=digital credential/issuance protocol=].
     </p>
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value is be one
-      of the well-defined keys defined in [[[#protocol-registry]]] or any other
-      custom one.
+      The {{DigitalCredentialCreateRequest/protocol}} member's value MUST be one
```
marcoscaceres: As above...

index.html
```diff
@@ -1125,185 +1228,6 @@ <h2>
         </dd>
       </dl>
     </section>
-    <h2 id="protocol-registry">
-      Registry of protocols
```
marcoscaceres, line 1178: I agree we should delete this, but I'm wondering if any of this is salvageable for how new protocols are added to the spec. Should we have a non-normative note about "file a bug if you think your protocol belongs here?" or something?

marcoscaceres (commented):
Like where this is going ... some initial feedback.

index.html
timcappalli: There is already a request for this. When they get added, I will update. For now, these should stay.

index.html
timcappalli: yes, this PR is not done (it is marked draft and not ready for review)...

index.html
marcoscaceres: Awesome! Merged https://github.com/tobie/specref/pull/898

index.html
marcoscaceres: Noted... ping when ready or want me to have another early look.

index.html
timcappalli: The text says: "supported by this specification."

.cspell/misc.txt
```diff
@@ -8,4 +8,6 @@ linkable
 linkability
 fingerprinters
 birthdates
-reidentify
\ No newline at end of file
+reidentify
+multisigned
+licence
```
TallTed: Not US?
```suggestion
license
```

timcappalli: no.

TallTed: OK. Whole document needs a UK English spellcheck.

index.html
TallTed: Perhaps...?
```suggestion
      Protocols
```

index.html
timcappalli: ReSpec does not provide a way to deep link into a referenced spec (as far as I'm aware).

index.html
```diff
@@ -83,7 +83,7 @@
           authors: ["Simone Onofri"],
           date: "2025-02-25",
           publisher: "W3C"
-        },
+        }
```
```suggestion
        },
```

index.html
marcoscaceres: My original comment stands, as I think "supported" will lead people to think that user agents support both protocols.  

I'd be ok with just "Protocols", as @TallTed suggests.

index.html
```diff
@@ -544,15 +539,120 @@ <h2>
...
+    <h2 id="supported-protocols">
+      Supported protocols
+    </h2>
+    <p>
+      The following [=digital credential/exchange protocols=] are supported
+      by this specification.
+    </p>
+    <table class="data">
+      <caption>
+        Table of supported [=digital credential/exchange protocols=]
```
marcoscaceres: We will probably want to "export" this definition. This table will probably be widely cited.

```suggestion
        <dfn class="export">Table of digital credential exchange protocols</dfn>
```

index.html
```diff
@@ -544,15 +539,120 @@ <h2>
...
+    </dl>
+    <!--
+    // MARK: Supported Protocols
+    -->
+    <h2 id="supported-protocols">
+      Supported protocols
+    </h2>
+    <p>
+      The following [=digital credential/exchange protocols=] are supported
+      by this specification.
```
marcoscaceres: My original comment stands, as I think "supported" will lead people to think that user agents support the listed protocols.  

It be good to just go with "Protocols", as @TallTed suggests for the heading. 

Instead of "supports", could we use either the word "permitted" or "recognized"?

**Permitted** implies the specification allows for these protocols to be used within the API's framework, but does not mandate their implementation in all user agents.

> "The following [=digital credential/exchange protocols=] are permitted by this specification."

**Recognized** suggests the specification formally acknowledges these protocols as valid types for the API's parameters, without promising implementation in every user agents.

> "The following [=digital credential/exchange protocols=] are recognized by this specification."

The "permitted" aspect seems more closely match out intent, specially with respect to the inclusion of the enum.

index.html
```diff
@@ -544,15 +539,120 @@ <h2>
...
+    </p>
```
marcoscaceres, line 569: Here we assure developers that at least one of the things listed will be supported, and how to check for support. We also reflect the reality of what is actually implemented and we mandate that at least _one_ must be implemented. 


```suggestion
    </p>
    <p>
       A [=user agent=] MUST implement at least one of the [=digital credential/exchange protocol=] listed in  the [=table of exchange protocols=]. However, it is OPTIONAL for a user agent to implement all the [=digital credential/exchange protocols=] listed in the [=table of exchange protocols=]. 
    </p>
    <aside class="note" title="Checking which exchange protocols a user agent allows">
      <p>Developers can check which [=digital credential/exchange protocols=]
      a user agent allows by calling the `DigitalCredential.`{{DigitalCredential.userAgentAllowsProtocol()}} static method.
      See [[[#checking-if-protocol-is-allowed]]] for a usage example.
      </p>
    </aside>
```

index.html
```diff
@@ -544,15 +539,120 @@ <h2>
...
+    </table>
+    <h3>
+      Protocols enum
+    </h3>
+      <p>
+        This enumeration's values correspond to the supported
+        [=digital credential/exchange protocols=].
+      </p>
+      <pre class="idl">
+      enum Protocol {
```
marcoscaceres: enums are global to the browser at the C++ level, so the name needs to be globally unique with no chance of collision (this is because other specs could use the enum). 

```suggestion
      enum DigitalCredentialExchangeProtocol {
```

We should also carefully consider if we want split the enum into:

`enum DigitalCredentialIssuanceProtocol` and an `enum DigitalCredentialPresentmentProtocol`.  

My recommendation is that we split them into two enums.

index.html
```diff
@@ -755,7 +855,7 @@ <h3>
     </p>
     <pre class="idl">
       dictionary DigitalCredentialGetRequest {
-        required DOMString protocol;
+        required Protocol protocol;
```
marcoscaceres: I think we should keep this as a DOMString for future compatibly, otherwise (future) older browsers will `TypeError` on new strings we might add to the spec (rather than just ignore them). 

What we should do instead is, in the `[[discoverFromExternalSource]]()` method and in `[[create]]()`, we do an IDL conversation to the appropriate enum value, while ignoring any we don't know.

This is both backwards compatible and future proof (this is what I've implemented in WebKit, btw).

index.html
marcoscaceres: This requirement should be dropped from here and we deal with it in the appropriate algorithm. 

As above, for backwards and future compatibility, we should just drop unknown requests protocols on the floor (as both Chromium and WebKit do today).

index.html
```diff
@@ -544,15 +539,120 @@ <h2>
...
+            OpenID for Verifiable Presentations 1.0 - Signed Requests
+          </td>
+          <td>
+            [=digital credential/Exchange Protocol=]
+          </td>
+          <td>
+            `openid4vp-v1-signed`
+          </td>
+          <td>
+            [[OPENID4VP]] Appendix A
```
```suggestion
            [[OPENID4VP]] <a data-cite="OPENID4VP#appendix-A">Appendix A</a>
```

index.html
```diff
@@ -544,15 +539,120 @@ <h2>
...
+            OpenID for Verifiable Presentations 1.0 - Multi-signed Requests
+          </td>
+          <td>
+            [=digital credential/Exchange Protocol=]
+          </td>
+          <td>
+            `openid4vp-v1-multisigned`
+          </td>
+          <td>
+            [[OPENID4VP]] Appendix A
```
```suggestion
            [[OPENID4VP]] <a data-cite="OPENID4VP#appendix-A">Appendix A</a>
```

index.html
```diff
@@ -544,15 +539,120 @@ <h2>
...
+            OpenID for Verifiable Presentations 1.0 - Unsigned Requests
+          </td>
+          <td>
+            [=digital credential/Exchange Protocol=]
+          </td>
+          <td>
+            `openid4vp-v1-unsigned`
+          </td>
+          <td>
+            [[OPENID4VP]] Appendix A
```
```suggestion
            [[OPENID4VP]] <a data-cite="OPENID4VP#appendix-A">Appendix A</a>
```

marcoscaceres: Should these point more specifically to the, for example "#name-unsigned-request"? 
```suggestion
            [[OPENID4VP]] <a data-cite="OPENID4VP#name-unsigned-request">§ Unsigned Requests</a>
```

index.html
```diff
@@ -544,15 +539,120 @@ <h2>
...
+          <td>
+            `openid4vp-v1-multisigned`
+          </td>
+          <td>
+            [[OPENID4VP]] Appendix A
+          </td>
+        </tr>
+        <tr>
+          <td>
+            ISO/IEC 18013-7:2025 - Annex C
```
marcoscaceres: Might be nicer (let ReSpec write out the full name of the spec): 

```suggestion
            [[[ISO18013-7]]] - Annex C
```

marcoscaceres:
Coincidently, @mohamedamir and I recently updated the web platform tests to include the registered types 😄  
https://github.com/web-platform-tests/wpt/pull/56627/

index.html
marcoscaceres: Sent some suggestions for this....

index.html
mohamedamir: +1 to "permitted"

index.html
timcappalli: ?

index.html
```diff
@@ -544,15 +539,120 @@ <h2>
...
+            [[[ISO18013-7]]] - Annex C
+          </td>
+          <td>
+            [=digital credential/Exchange Protocol=]
+          </td>
+          <td>
+            `org-iso-mdoc`
+          </td>
+          <td>
+            [[ISO18013-7]] Annex C
```
timcappalli, line 647: @marcoscaceres for consistency, this should also have a deep link. Any idea how to do that?

index.html
marcoscaceres: Just a stray change. Please leave this one as is.

index.html
marcoscaceres: Aren't they PDFs? I don't think we can do much here until ISO decides to publish web friendly formats 😢

index.html
```diff
@@ -2285,7 +2228,20 @@ <h2>
     </section>
     <section class="prefer-full-spec-title" id="index"></section>
     <section id="idl-index"></section>
-    <section id="conformance"></section>
+    <section id="conformance">
+      <h2>
+        Enumerations as DOMString types
+      </h2>
+      <p>
+        Enumeration types are not referenced by other parts of the Web IDL because that
```
marcoscaceres: We can add this to the appropriate algorithm as a note. Or better, if we do use the enums in the algorithms (even if not seen by JavaScript), then we don't need this text at all (i.e., we do a "IDL conversion" to an enum value, which is a perfectly fine thing to do).

index.html
```diff
@@ -113,12 +113,8 @@ <h2 class="informative">
...
     </p>
     <p>
-      The API design is agnostic to both credential [=digital
-      credential/presentation requests|presentation=] [=digital
-      credential/exchange protocols=], credential [=digital credential/issuance
-      request|issuance=] [=digital credential/issuance protocols|protocols=]
-      and credential formats. However, to promote interoperability this
-      document includes a [[[#protocol-registry]]].
+      The API design is agnostic to credential formats and includes support for
+      multiple [=digital credential/exchange protocols=]. See [[[#supported-protocols]]].
```
mohamedamir: Did we establish the convention that exchanges protocols refer to both presentation and issuance?

index.html
```diff
@@ -544,15 +539,107 @@ <h2>
...
-    </dl><!--
+    </dl>
+    <!--
+    // MARK: Supported Protocols
+    -->
+    <h2 id="supported-protocols">
+      Supported protocols
+    </h2>
+    <p>
+      The following presentation and [=digital credential/issuance protocols|issuance=] protocols are supported
```
```suggestion
      The following [=digital credential/exchange protocols|presentation=] and [=digital credential/issuance protocols|issuance=] protocols are supported
```

timcappalli: This will get updated as part of #434

index.html
timcappalli: This will get changed as part of #434

index.html
wseltzer: Would "recognized" work?

index.html
timcappalli: is it really different than any other spec feature though? We don't say in the spec "If you want to add a new feature, create an issue".

index.html
TallTed: Works for me, with shorter lines that make life easier in many ways. I also added a missing `s`. I think that's all... but I encourage review with local DIFF of your choice, since GitHub fails to properly highlight this suggestion.
```suggestion
    </p>
    <p>
       A [=user agent=] MUST implement at least one of the
       [=digital credential/exchange protocols=] listed in the
       [=table of exchange protocols=]. However, it is OPTIONAL for a user agent
       to implement all the [=digital credential/exchange protocols=] listed in
       the [=table of exchange protocols=]. 
    </p>
    <aside class="note" title="Checking which exchange protocols a user agent allows">
      <p>Developers can check which [=digital credential/exchange protocols=]
      a user agent allows by calling the
      `DigitalCredential.`{{DigitalCredential.userAgentAllowsProtocol()}} static
      method. See [[[#checking-if-protocol-is-allowed]]] for a usage example.
      </p>
    </aside>
```

timcappalli: In my opinion, this should be added in a separate PR, as there was not alignment on what this text should say during the 2026-01-12 call.

index.html
RByers: "Recognized" works for me, but yes also to Ted just "Protocols". The specification defines an API surface including possible values for this string, I don't think it's really any different than say the ["CredentialMediationRequirement" enum](https://w3c.github.io/webappsec-credential-management/#enumdef-credentialmediationrequirement), except for the detail of how exactly unrecognized values are handled.

index.html
RByers: FWIW I'm personally fine with Marcos's suggested text here. Also fine leaving it to a future PR if there's not agreement.

index.html
RByers: I think it's pretty unlikely that we're going to see people wanting to add more protocols, I'd hate to complicate our spec with such text. I'd be OK pulling this out to a separate md file in the repo though Marcos if you want to keep it visible beyond an old spec checkpoint?

index.html
RByers: Describing in the algorithm that the value is converted to the enum (with failures silently dropped without throwing) sounds right to me.

index.html
marcoscaceres: Sorry, can you clarify? Isn't the requirement that **at least one** must be implemented?

timcappalli: There is not working group consensus on that and would prefer to not hold up this PR.

marcoscaceres: Happy to send as a follow-up, but not sure I follow. Speaking out loud here, how else do we guarantee that at least one protocol will be available to developers? 

This conformance requirement gives developers the assurance that one of the protocols will be supported, while not putting a requirement on implementers as to which they MUST implement (the reality today). 

We could certainly tighten the requirement for user agents to support all the listed APIs (i.e., "MUST implement 1, SHOULD implement the others", but we can't really force implementers to implement one or the other. Making it a SHOULD is also extensible, in as far as "legacy user agents" would still be conforming even if we update the Protocols in the future. 

Anyway... I'll file a bug and we can follow up.

marcoscaceres: Ok, filed https://github.com/w3c-fedid/digital-credentials/issues/439 ... optionally, we should flag this as an issue?

```suggestion
    </p>
    <aside class="issue" data-number="439">
    </aside>
```

index.html
marcoscaceres: If we get consensus on #439, we can call these "Recommended Protocols" - which are the ones the working group RECOMMENDs ("RFC2119 SHOULD" equivalent, with at least one being mandated but being left to the user agent).

index.html
```diff
@@ -841,10 +949,27 @@ <h4>
       member denotes the [=digital credential/issuance protocol=].
     </p>
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value is be one
-      of the well-defined keys defined in [[[#protocol-registry]]] or any other
-      custom one.
+      The {{DigitalCredentialCreateRequest/protocol}} member's value MUST be one
```
marcoscaceres: Statement of fact: 

```suggestion
      The {{DigitalCredentialCreateRequest/protocol}} member's value is one
```

As it will be algorithmically enforced (and tested) in #420.

index.html
mohamedamir: I don't think "Recommended Protocols" conveys accurately the semantics of this list.
I like simply "Protocols" as Ted suggested.

index.html
mohamedamir: After the last WG discussion, I am leaning more towards "supported" or "recognized"

index.html
timcappalli: For the sake of making progress, let's do "Protocols" (with a note on record that I still believe it should be "Supported Protocols").

index.html
```diff
@@ -804,10 +893,31 @@ <h4>
       denotes the [=digital credential/exchange protocol=].
     </p>
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value can be one
-      of the well-defined protocol identifiers defined in
-      [[[#protocol-registry]]] or a custom protocol identifier.
+      The {{DigitalCredentialCreateRequest/protocol}} member's value MUST be one
```
marcoscaceres: Statement of fact also:

```suggestion
      The {{DigitalCredentialCreateRequest/protocol}} member's value is one
```

index.html
```diff
@@ -552,15 +547,109 @@ <h2>
...
+    <!--
+    // MARK: Supported Protocols
+    -->
+    <h2 id="protocols">
+      Protocols
+    </h2>
+    <p>
+      The following presentation and [=digital credential/issuance protocols|issuance=] protocols are supported
+      by this specification.
+      <!-- TODO: replace with new dfn from 434" -->
```
marcoscaceres: these TODO: are accidentally  side-stepping having consensus on #434. Please use the current terminology and then we can find and replace with the new terminology once we have agreement and a PR that resolves #434. 🙏

index.html
```diff
@@ -552,15 +547,109 @@ <h2>
...
-    </dl><!--
+    </dl>
+    <!--
+    // MARK: Supported Protocols
+    -->
+    <h2 id="protocols">
+      Protocols
+    </h2>
+    <p>
+      The following presentation and [=digital credential/issuance protocols|issuance=] protocols are supported
```
```suggestion
      The following [digital credential/exchange protocol|exchange=] and [=digital credential/issuance protocols|issuance=] protocols are supported
```

index.html
timcappalli: It's not side stepping anything. The TODO is explicitly called out, the text does not reference any new term, and the existing incorrect reference was removed.

marcoscaceres: We don’t need them though. 

When “presentation” is in the spec, it won’t link to anything (readers can’t see the TODO… they are not helpful)

Why not just send a PR quickly to fix #434? 

We land that, we can land this (with the right links).

index.html
```diff
@@ -2224,7 +2169,20 @@ <h2>
...
+      </h2>
+      <p>
+        Enumeration types are not referenced by other parts of the Web IDL because that
+        would preclude other values from being used without updating this specification
+        and its implementations. It is important for backwards compatibility that
+        user agents handle unknown values. Enumerations for this specification exist
+        here for documentation and as a registry. Where the enumerations are represented
+        elsewhere, they are typed as {{DOMString}}s, for example in
+        {{DigitalCredentialGetRequest/protocol}}.
+      </p>
```
marcoscaceres: As also per @RByers' comments. 

```suggestion
```

After this lands, I'll update #420 to use the enum.

index.html
```diff
@@ -804,10 +893,31 @@ <h4>
...
     </p>
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value can be one
-      of the well-defined protocol identifiers defined in
-      [[[#protocol-registry]]] or a custom protocol identifier.
+      The {{DigitalCredentialCreateRequest/protocol}} member's value is one
+      of the protocol identifiers defined in [[[#protocols]]].
     </p>
+    <h5>
+      Presentation protocol enumeration
```
marcoscaceres: As these are Web IDL, these enums should be moved to The Digital Credentials API section under the heading "`DigitalCredentialPresentationProtocol` enum" and so on. 

We will use them in the algorithms of the API. 


```suggestion
      The {{DigitalCredentialPresentationProtocol}} enum
```

index.html
```diff
@@ -878,10 +988,27 @@ <h4>
...
     </p>
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value is be one
-      of the well-defined keys defined in [[[#protocol-registry]]] or any other
-      custom one.
+      The {{DigitalCredentialCreateRequest/protocol}} member's value is one
+      of the protocol identifiers defined in [[[#protocols]]].
     </p>
+    <h5>
+      Issuance protocol enumeration
```
marcoscaceres: As above... 

```suggestion
     The {{DigitalCredentialIssuanceProtocol}} enum
```

And move to the API section.

index.html
```diff
@@ -878,10 +988,27 @@ <h4>
...
+      enum DigitalCredentialIssuanceProtocol {
+        "placeholder-protocol"
+      };
+      </pre>
+      <aside class="note">
+        <p>
+          Note: The {{DigitalCredentialPresentationProtocol}} enumeration is
+          deliberately not referenced, see [[[#enumerations-as-domstring-types]]].
+        </p>
+      </aside>
```
marcoscaceres: Will be referenced in #420 

```suggestion
```

index.html
```diff
@@ -804,10 +893,31 @@ <h4>
...
+        "openid4vp-v1-multisigned",
+        "org-iso-mdoc"
+      };
+      </pre>
+      <aside class="note">
+        <p>
+          Note: The {{DigitalCredentialPresentationProtocol}} enumeration is
+          deliberately not referenced, see [[[#enumerations-as-domstring-types]]].
+        </p>
+      </aside>
```
marcoscaceres: Will be referenced in #420 

```suggestion
```

index.html
```diff
@@ -804,10 +893,31 @@ <h4>
...
+    </h5>
```
marcoscaceres: Stray h5.

index.html
marcoscaceres: If we use some of this text, please add it to the Protocol section instead. It doesn't make sense here. Conformance section is just used for boilerplate.

index.html
timcappalli: this is in the API section?

index.html
MasterKale: How about "Use of the following presentation and issuance protocols are defined by this specification"?

index.html
```diff
@@ -552,15 +547,109 @@ <h2>
...
+      <tbody>
+        <tr>
+          <td>
+            [[[OPENID4VP]]] - Unsigned Requests
+          </td>
+          <td>
+            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+          </td>
+          <td>
+            `openid4vp-v1-unsigned`
```
marcoscaceres: This will make the connection between the `DigitalCredentialPresentationProtocol` explicit. 

```suggestion
            <dfn data-dfn-for="DigitalCredentialPresentationProtocol">openid4vp-v1-unsigned</dfn>
```

index.html
```diff
@@ -552,15 +547,109 @@ <h2>
...
+        </tr>
+        <tr>
+          <td>
+            [[[OPENID4VP]]] - Signed Requests
+          </td>
+          <td>
+            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+          </td>
+          <td>
+            `openid4vp-v1-signed`
```
```suggestion
            <dfn data-dfn-for="DigitalCredentialPresentationProtocol">openid4vp-v1-signed</dfn>
```

index.html
```diff
@@ -552,15 +547,109 @@ <h2>
...
+        </tr>
+        <tr>
+          <td>
+            [[[OPENID4VP]]] - Multi-signed Requests
+          </td>
+          <td>
+            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+          </td>
+          <td>
+            `openid4vp-v1-multisigned`
```
```suggestion
            <dfn data-dfn-for="DigitalCredentialPresentationProtocol">openid4vp-v1-multisigned</dfn>
```

index.html
```diff
@@ -804,10 +893,24 @@ <h4>
...
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value can be one
-      of the well-defined protocol identifiers defined in
-      [[[#protocol-registry]]] or a custom protocol identifier.
+      The {{DigitalCredentialCreateRequest/protocol}} member's value is one
+      of the protocol identifiers defined in [[[#protocols]]].
     </p>
+    <h5>
+      The {{DigitalCredentialPresentationProtocol}} enumeration
+    </h5>
```
marcoscaceres: These should be:

```suggestion
    <h3>
      The {{DigitalCredentialPresentationProtocol}} enumeration
    </h3>
```

And moved to the other section.

index.html
```diff
@@ -878,10 +981,21 @@ <h4>
...
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value is be one
-      of the well-defined keys defined in [[[#protocol-registry]]] or any other
-      custom one.
+      The {{DigitalCredentialCreateRequest/protocol}} member's value is one
+      of the protocol identifiers defined in [[[#protocols]]].
     </p>
+    <h5>
+      The {{DigitalCredentialIssuanceProtocol}} enumeration
+    </h5>
```
```suggestion
    <h3>
      The {{DigitalCredentialIssuanceProtocol}} enumeration
    </h3>
```

index.html
```diff
@@ -878,10 +981,21 @@ <h4>
...
+      <p>
+        This enumeration's values correspond to the supported
+        [=digital credential/issuance protocols=]. 
+      </p>
+      <pre class="idl">
+      enum DigitalCredentialIssuanceProtocol {
+        "placeholder-protocol"
```
marcoscaceres: We can leave this empty. 

```suggestion
```

index.html
```diff
@@ -552,15 +547,109 @@ <h2>
...
+        </tr>
+        <tr>
+          <td>
+            [[[ISO18013-7]]] - Annex C
+          </td>
+          <td>
+            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+          </td>
+          <td>
+            `org-iso-mdoc`
```
```suggestion
            <dfn data-dfn-for="DigitalCredentialPresentationProtocol">org-iso-mdoc</dfn>
```

index.html
marcoscaceres: This should be `"openid4vci"`, right @mohamedamir? 

```suggestion
        "openid4vci"
```

That's what we have in the test suite as the canonical string:
https://github.com/web-platform-tests/wpt/blob/020ff60edf5cb3b2d1fa5d2ac32ba34b38836822/digital-credentials/support/helper.js#L20

marcoscaceres (changes requested):
I tried to implement the spec changes and managed to identify some critical issues with the PR:

* `DigitalCredentialIssuanceProtocol` introduces a fake value. The enum should have a real value (i.e., "openid4vci" ... that's what Chrome seems to use according to the Web Platform Tests). 
* We need to introduce a new `typedef (DigitalCredentialPresentationProtocol or DigitalCredentialIssuanceProtocol) DigitalCredentialsProtocol;` 
* `DigitalCredential`'s `protocol` member's type should be `DigitalCredentialsProtocol`, not `DOMString`.

index.html
timcappalli: No, it should not. That protocol string does not exist yet. I don't see any concern with it being using just for tests temporarily though, to support the development of the protocol.

Bullets 2 and 3 now have an issue: https://github.com/w3c-fedid/digital-credentials/issues/441

index.html
timcappalli: works for me.

index.html
timcappalli: Enums cannot be empty.

index.html
```diff
@@ -975,13 +1061,43 @@ <h4>
...
+    </h4>
+      <p>
+        This enumeration's values correspond to the supported
+        [=digital credential/issuance protocols=] listed in [[[#protocols]]]. 
+      </p>
+      <pre class="idl">
+      enum DigitalCredentialIssuanceProtocol {
+        "placeholder-protocol"
+      };
+      </pre>
```
marcoscaceres: As there's no actual value here, this should be removed. We should only add it when there's a real issuance identifier. 

```suggestion
```

index.html
```diff
@@ -975,13 +1061,43 @@ <h4>
...
       <li>Return `true` if the user agent allows |protocol|, otherwise return
       `false`.
       </li>
     </ol>
+    <h3>
+      Supporting Data Structures
+    </h3>
+    <p>Data structures, such as enumerations, which support {{DigitalCredential}}
+      in this specification.
+    </p>
```
marcoscaceres, line 1077: We don't need this. The enums will be used by `DigitalCredential`'s `protocol` member. They are a normal part of the API. 

```suggestion
```

index.html
marcoscaceres: @mohamedamir? this is somewhat concerning. 

I'm concerned with having fake stuff in the spec. So, consider this my objection to the enum being included until such time we have a standardized string. Enums don't allow empty values for reason. Putting fake values in them to get around that restriction not something we should be doing.  

Put differently: If we can update a fake string, we can just as easily add a real enum if/when such time when an issuance protocol gets standardized.

index.html
timcappalli: While a subsection isn't "needed", I think it is a cleaner approach long term.

index.html
```diff
@@ -552,15 +547,108 @@ <h2>
...
-    </dl><!--
+    </dl>
+    <!--
+    // MARK: Supported Protocols
+    -->
+    <h2 id="protocols">
+      Protocols
+    </h2>
+    <p>
+      Use of the following presentation and issuance protocols are defined by this specification.
```
```suggestion
      Use of the following [=digital credential/exchange protocol=] and [=digital credential/issuance protocols=] are defined by this specification.
```

index.html
```diff
@@ -552,15 +547,108 @@ <h2>
...
+      <!-- TODO: replace with new dfn from 434" -->
```
```suggestion
     Use of the following [=digital credential/exchange protocols=] and [=digital credential/issuance protocols=] are defined by this specification.
```

index.html
```diff
@@ -552,15 +547,108 @@ <h2>
...
+    </p>
+    <aside class="issue" data-number="439">
+    </aside>
+    <table class="data">
+      <caption>
+        <dfn class="export">Table of supported presentation and [=digital credential/issuance protocol|issuance=] protocols</dfn>
+        <!-- TODO: replace with new dfn from 434" -->
```
```suggestion
        <dfn class="export">Table of supported [=digital credential/exchange protocols=] and [=digital credential/issuance protocol|issuance=] protocols</dfn>
```

```suggestion
        <dfn class="export">Table of supported [=digital credential/exchange protocols=] and [=digital credential/issuance protocol|issuance=] protocols</dfn>
```

index.html
```diff
@@ -552,15 +547,108 @@ <h2>
...
+          <td>
+            [[OPENID4VP]] <a data-cite="OPENID4VP#name-jws-json-serialization">Appendix A: 3.2.2</a>
+          </td>
+        </tr>
+        <tr>
+          <td>
+            [[[ISO18013-7]]] - Annex C
+          </td>
+          <td>
+            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
```
```suggestion
            [=digital credential/Exchange protocols=]
```

index.html
```diff
@@ -552,15 +547,108 @@ <h2>
...
       <dt>
         Request coordinator
       </dt>
       <dd>
         See [=credential request coordinator=].
       </dd>
-    </dl><!--
+    </dl>
+    <!--
+    // MARK: Supported Protocols
```
```suggestion
    // MARK: Protocols
```

index.html
```suggestion
        <dfn class="export">Table of supported [=digital credential/exchange protocols=] and [=digital credential/issuance protocol|issuance=] protocols</dfn>
```

index.html
```diff
@@ -552,15 +547,108 @@ <h2>
...
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td>
+            [[[OPENID4VP]]] - Unsigned Requests
+          </td>
+          <td>
+            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
```
```suggestion
            [=digital credential/Exchange protocol=]
```

index.html
marcoscaceres: I strongly disagree. Additional text just bloats the spec and is a maintenance burden and adds limited value. 

Please only add normative prose to the text.

index.html
```diff
@@ -975,13 +1061,43 @@ <h4>
...
+    <h4>
+      The {{DigitalCredentialIssuanceProtocol}} enumeration
+    </h4>
+      <p>
+        This enumeration's values correspond to the supported
+        [=digital credential/issuance protocols=] listed in [[[#protocols]]]. 
+      </p>
+      <pre class="idl">
+      enum DigitalCredentialIssuanceProtocol {
+        "openid4vci-v1"
```
marcoscaceres: Great! But I don't see it in the table? 

I also don't see the string in? 
https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html

What am I missing?

index.html
timcappalli: Well then we can agree to disagree. Let the chairs know if you believe editorial style / choices should block the merging of this PR. Thanks!

index.html
timcappalli: In a [previous comment](https://github.com/w3c-fedid/digital-credentials/pull/401#discussion_r2715261938), you proposed this change.

index.html
marcoscaceres: Yes, because every word matters in a spec. 

Happy to have @mohamedamir break the deadlock as third editor.

timcappalli: The difference between these two approaches is two lines and ~6 words....

index.html
marcoscaceres: Yes, because I was under the impression and trusted that that `"openid4vci"` was a real thing (I don't participate in the OpenID standardization). 

If `openid4vci-v1` is not a real thing, then this enum shouldn't be here. Please remove it until such time as there's a standardized protocol string that can go into the table.

index.html
```diff
@@ -552,15 +547,108 @@ <h2>
...
+          <td>
+            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+          </td>
+          <td>
+            <dfn data-dfn-for="DigitalCredentialPresentationProtocol">`openid4vp-v1-multisigned`</dfn>
+          </td>
+          <td>
+            [[OPENID4VP]] <a data-cite="OPENID4VP#name-jws-json-serialization">Appendix A: 3.2.2</a>
+          </td>
+        </tr>
```
marcoscaceres, line 635: The table is missing OPENID4VCI: 

```suggestion
        </tr>
        <tr>
          <td>
            [[[OPENID4VCI]]]
          </td>
          <td>
           [digital credential/Issuance protocol=] 
          </td>
          <td>
            <dfn data-dfn-for="DigitalCredentialIssuanceProtocol">openid4vci-v1</dfn>
          </td>
          <td></td>
        </tr>
```

index.html
marcoscaceres: It literally adds a sub section to the spec and makes the Table of Contents more unreadable: 

Screenshot 2026-01-23 at 12 42 04 PM

It also brings into question why the other things are not here... what constitutes a "supporting structure". Are enums even structures? etc...

index.html
```diff
@@ -552,15 +547,108 @@ <h2>
...
+        <tr>
+          <td>
+            [[[ISO18013-7]]] - Annex C
+          </td>
+          <td>
+            Presentation Protocol <!-- TODO: replace with new dfn from 434" -->
+          </td>
+          <td>
+            <dfn data-dfn-for="DigitalCredentialPresentationProtocol">`org-iso-mdoc`</dfn>
```
bc-pi, line 644: does ISO18013-7 define `org-iso-mdoc`?

index.html
```diff
@@ -552,15 +547,108 @@ <h2>
...
+          <td>
+            <dfn data-dfn-for="DigitalCredentialPresentationProtocol">`openid4vp-v1-multisigned`</dfn>
+          </td>
+          <td>
+            [[OPENID4VP]] <a data-cite="OPENID4VP#name-jws-json-serialization">Appendix A: 3.2.2</a>
+          </td>
+        </tr>
+        <tr>
+          <td>
+            [[[ISO18013-7]]] - Annex C
```
bc-pi, line 638: I must admit to some ignorance about the tooling stack here but the [preview](https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/401.html#normative-references) seems to indicate that this results in a normative reference to a document that requires payment for access. Whether through registry or direct listing in this spec, and more so even for the latter, I don't find that acceptable. 


[ISO18013-7]
    [ISO/IEC 18013-7:2025 ISO-compliant driving licence, Part 7: Mobile driving licence (mDL) add-on functions](https://www.iso.org/standard/91154.html). ISO/IEC JTC 1/SC 17. International Organization for Standardization. May 2025. URL: https://www.iso.org/standard/91154.html


Image


Image

index.html
marcoscaceres: Yes.i recall it does but can reconfirm.

index.html
marcoscaceres: I heard a while back there was an effort to make it freely available. I’ll see if I can get some status on that. 

Also noting that it’s not uncommon for W3C specs to cite ISO specs even though they are behind a paywall. The W3C Process doesn’t prevent that.

index.html
marcoscaceres: In C.1 General, it stats that:

Screenshot 2026-01-26 at 3 57 57 PM

index.html
```diff
@@ -804,9 +892,8 @@ <h4>
       denotes the [=digital credential/exchange protocol=].
     </p>
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value can be one
-      of the well-defined protocol identifiers defined in
-      [[[#protocol-registry]]] or a custom protocol identifier.
+      The {{DigitalCredentialGetRequest/protocol}} member's value is one
+      of the protocol identifiers defined in {{DigitalCredentialPresentationProtocol}}.
```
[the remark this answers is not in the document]
timcappalli: There was text in the spec to explain this, but was removed due to @marcoscaceres objection.

index.html
```diff
@@ -975,13 +1062,34 @@ <h4>
       algorithm:
     </p>
     <ol class="algorithm">
-      <li>If |protocol| is not a [=digital credential/protocol identifier=],
+      <li>If |protocol| is not a member of {{DigitalCredentialPresentationProtocol}},
       return `false`.
       </li>
```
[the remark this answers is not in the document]
timcappalli: This will get addressed in a future PR.

index.html
```diff
@@ -878,9 +966,8 @@ <h4>
       member denotes the [=digital credential/issuance protocol=].
     </p>
     <p>
-      The {{DigitalCredentialCreateRequest/protocol}} member's value is be one
-      of the well-defined keys defined in [[[#protocol-registry]]] or any other
-      custom one.
+      The {{DigitalCredentialCreateRequest/protocol}} member's value is one
+      of the protocol identifiers defined in [[[#protocols]]].
```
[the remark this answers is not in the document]
timcappalli: This will be addressed in a future PR.

index.html
timcappalli: will be addressed in #434

index.html
timcappalli: will be addressed in #434

index.html
timcappalli: will be addressed in #434

index.html
timcappalli: will be addressed in #434

timcappalli: will be addressed in #434

index.html
timcappalli: will be addressed in #434

index.html
timcappalli: will be addressed in #434

index.html
timcappalli: Noted with thanks.

index.html
bc-pi: I don't find that acceptable.

marcoscaceres: We should probably discuss in https://github.com/w3c-fedid/digital-credentials/issues/214? 

Would it be acceptable to make the table non-normative? There’s no implementation requirement on any user agent to support any one protocol. The protocols just become “Appendix of known protocols”. 

Otherwise we drop this section entirely or wait a few weeks until ISO makes 18013-7 available for free?

index.html
bc-pi: How does a regular person see that?

index.html
bc-pi: > We should probably discuss in #214?

That was about inclusion criteria for the registry that is going away and is marked to be closed by this. 

 
> Would it be acceptable to make the table non-normative? There’s no implementation requirement on any user agent to support any one protocol. The protocols just become “Appendix of known protocols”.

Not really, no. 


> Otherwise we drop this section entirely or wait a few weeks until ISO makes 18013-7 available for free?

I thought today @RByers said it already was. But maybe he misspoke or I misheard. 

Why cannot this PR proceed without referencing a closed pay-for-access document? And when/if 18013-7 is made available for free, then that can be added in a separate PR? @timcappalli (and maybe @lee) didn't seem to like that idea on the call ~2 hours ago but I'm honestly not seeing why that would be problematic.

marcoscaceres: > That was about inclusion criteria for the registry that is going away and is marked to be closed by this.

Yeah :( ... because the presumption was that the ISO spec would be available for free already, which renders that discussion mute.

> I thought today @RByers said it already was. But maybe he misspoke or I misheard.

If it was we wouldn't be in this predicament, right?  An ugly hack would be to just point to their Working Draft PDFs that they put up on GitHub... for example:

https://github.com/ISOWG10/ISO-18013/blob/main/Working%20Documents/

But that's a bit mean... but hey, at least they make something public. 😛 

(The above was a joke... let's not cite that!) 

> I'm honestly not seeing why that would be problematic.

The problem is that it creates developer confusion (if only temporarily)... the spec will list the OpenID ones, but not the ISO one. And because the algorithms and API make use of the enums defined here, things start falling out of sync between implementations and the spec: we want the spec to match reality.

index.html
```diff
@@ -552,15 +548,108 @@ <h2>
...
+    </h2>
+    <p>
+      Use of the following presentation protocols is defined by this specification.
+      <!-- TODO: replace with new dfn from 434 -->
+    </p>
+    <aside class="issue" data-number="439">
+    </aside>
+    <table class="data">
+      <caption>
+        <dfn class="export">Table of supported presentation and [=digital credential/issuance protocol|issuance=] protocols</dfn>
```
marcoscaceres, line 574: I thought we agreed not to use the word "supported"? Can it just be?: 

```suggestion
        <dfn class="export">Table of presentation and [=digital credential/issuance protocol|issuance=] protocols</dfn>
```

index.html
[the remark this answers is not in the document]
marcoscaceres: Co-pilot is (sorta) correct here. We can't just point at the table that doesn't contain an issuance value (it doesn't match the use of an enum for presentation).  

Honest question: can someone please explain why we can't list [[OPENID4VCI]] and `openid4vci-v1`? My understanding was that Chrome implemented that (or the `openid4vci` string, as that's what in the Web Platform Tests). 

This removal of things is making me paranoid that [[OPENID4VCI]] and `openid4vci-v1` is not actually supported by any user agent? If that's the case, that's ok... but it feels like something weird is going on here. 

@leecam or @mohamedamir, can you maybe explain the Chrome situation? In https://github.com/w3c-fedid/digital-credentials/pull/204 it was indicated that issuance had implementation commitments, but does it not support the `openid4vci-v1`? How is issuance implemented and does `openid4vci-v1` work?

In the Web Platform Tests, we explicitly test for `"openid4vci"`... was that not correct? Do we need to remove that from the test suite as it's bad value (or change it to `openid4vci-v1` rather)?

PR #402, opened 2025-12-01
chore: Adds cspell config

timcappalli:


marcoscaceres:
Please don't merge stuff until I've also had a chance to review it. 🙏

We should be running this on the GitHub sever (pre-commit check)... this is not something I have locally and it's not documented how to actually use it in the CONTRIBUTING.md, for instance.

Issue #222, opened 2025-04-20
Handling Multiple Credentials from Different Wallets in DC API Presentation Responses


mohamedamir:
**Summary:**

This issue addresses the challenge of handling scenarios where a verifier requests multiple verifiable credentials (VCs) that reside in potentially different digital wallets within the context of DC API for Presentations. For example, when age proof together with payment authentication are requested in a single DC API call.

When a Verifier sends a presentation request for multiple VCs held across different credential providers, our current architecture encounters difficulties in aggregating and returning these credentials. The primary obstacle lies in the response encryption mechanism, which is not designed to handle data originating from multiple independent wallets. Consequently, we cannot combine the encrypted responses from each wallet into a single, decryptable result for the Verifier.

**Proposed Solutions:**
**1- Returning an Array of VC Objects**
This approach suggests modifying the API to return an array of individual DC objects, where each object represents a credential retrieved from a different wallet.
**Pros:**
- Simpler to implement on the platform/credential selector level.
- Can be easily compared to sending multiple presentation requests and aggregating the responses.

**Cons:**
 - Exposes complexity to the API level that are specific to rare use cases and are only supported by some protocols.
 - Could potentially cause the Android API and the Web API to diverge. Web DC API returns an array, and the Android DC API returns a single VC.
 - Could be confusing since the request contains an array of requests, and if the response also contains an array of responses, it could give the illusion that multiple requests have been satisfied, although one one requests of the requests array is satisfied (it just happened that this satisfied request, asked for multiple VCs) . 

**2- Protocol-Level Modification**
Instead of exposing such complexity on the API-level, this approach suggests keeping that complexity hidden on the protocol level. Openid4vp for example can define its response to be either a single VC or an array of VCs depending on the contents of the request, and it is up to the verifier to inspect the response.

Pros:
 - Keep the Web API simple, and aligned with the platform API. (Unless the platform API will also change)
 - Keeps the API optimized for the mainstream use case of requesting one VC per DC API call.

Cons:
 - Complicates the aggregation of multiple VCs on the platform level.
 - Could result in more complex protocol definitions.

npdoty:
Requesting multiple credentials in a single request seems confusing and complex (to both the site developer and the user) and hostile to the user. Making requests about a single credential makes error handling easier for the site, makes implementation easier, and lends itself to more informed and appropriately friction-ful consent requests to the user.

[...]

PR #419, opened 2025-12-16
Define "present credential requests" algorithm.

marcoscaceres:
Defines the "present the credential request" algorithm — the core user-facing part of the Digital Credentials flow that happens after validation passes.

## What the algorithm does

Once the credential requests have been validated (by the "prepare" algorithm), this algorithm:

1. **Constructs a request context** — bundles the validated requests with the top-level and document origins, so the platform/chooser knows who is asking and from where.
2. **Presents a digital credential chooser** — a UA-provided UI that shows the user available credentials from registered holders. The user can select a credential, cancel, or the operation can be aborted via an `AbortSignal`.
3. **Handles abort** — if the signal fires, the algorithm returns (the abort callback was already wired up by the "prepare" steps).
4. **Handles cancellation** — rejects with `AbortError`.
5. **Maps platform errors** — translates platform-specific failures to web-facing errors: `NotAllowedError` (permission denied), `TypeError` (malformed request data), `InvalidStateError` (concurrent request), or `OperationError` (catch-all).
6. **Processes the response** — parses the holder's response as JSON, validates it's an object, constructs a `DigitalCredential` instance with `data` and `protocol`, resolves the promise, and cleans up coordinator state.

## New terms introduced

- **Digital credential chooser** — the UA-provided UI for credential selection
- **Request context** — a struct carrying `requests`, `top origin`, and `document origin` into the chooser

Closes #419

- [x] Modified Web platform tests (https://github.com/web-platform-tests/wpt/pull/59087)

Implementation commitment:

- [x] [WebKit](https://github.com/WebKit/WebKit/pull/56489)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [x] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/419.html" title="Last updated on Apr 29, 2026, 10:38 PM UTC (001a6df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/419/4c1c0df...001a6df.html" title="Last updated on Apr 29, 2026, 10:38 PM UTC (001a6df)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1087,8 +1087,8 @@
         <ol>
           <li>Let |error| be a newly created {{"AbortError"}} {{DOMException}}.
           </li>
-          <li>[=credential request coordinator/Reject request with=] |error|
-          and |promise|.
+          <li>[=credential request coordinator/reject the credential request
+          with=] |error| and |promise|.
           </li>
           <li>Return.
           </li>
@@ -1124,8 +1124,8 @@
               </dd>
             </dl>
           </li>
-          <li>[=credential request coordinator/Reject request with=] |error|
-          and |promise|.
+          <li>[=credential request coordinator/reject the credential request
+          with=] |error| and |promise|.
           </li>
           <li>Return.
           </li>
@@ -1148,8 +1148,8 @@
           </li>
           <li>If |parsedDataOrError| is an [=exception=]:
             <ol>
-              <li>[=credential request coordinator/Reject request with=] |parsedDataOrError|
-              and |promise|.
+              <li>[=credential request coordinator/reject the credential
+              request with=] |parsedDataOrError| and |promise|.
               </li>
               <li>Return.
               </li>
@@ -1159,8 +1159,8 @@
             <ol>
               <li>Let |error| be a newly created {{TypeError}}.
               </li>
-              <li>[=credential request coordinator/Reject request with=] |error|
-              and |promise|.
+              <li>[=credential request coordinator/reject the credential
+              request with=] |error| and |promise|.
               </li>
               <li>Return.
               </li>
@@ -1183,9 +1183,8 @@
               request coordinator/active promise=] is not |promise|, then
               return.
               </li>
-              <li>Let |abortSignal| be the [=credential request
-              coordinator=]'s [=credential request coordinator/abort
-              signal=].
+              <li>Let |abortSignal| be the [=credential request coordinator=]'s
+              [=credential request coordinator/abort signal=].
               </li>
               <li>Let |abortAlgorithm| be the [=credential request
               coordinator=]'s [=credential request coordinator/abort
--- a/index.html
+++ b/index.html
@@ -1132,6 +1132,9 @@
       </li>
       <li>If a [=digital credential=] was selected by the user:
         <ol>
+          <li>Let |selectedRequest| be the item in |validatedRequests| that was
+          used to invoke the [=holder=] for this exchange.
+          </li>
           <li>Let |responseData| be a [=string=] [=digital
           credential/presentation response=] or [=string=] [=digital
           credential/issuance response=] returned by the [=holder=].
@@ -1166,8 +1169,8 @@
             </ol>
           </li>
           <li>Let |protocol:DigitalCredentialProtocol| be the [=enumeration
-          value=] of the [=digital credential/protocol identifier=] used in the
-          exchange.
+          value=] of |selectedRequest|'s [=digital credential/protocol
+          identifier=].
           </li>
           <li>Let |credential| be a newly created {{DigitalCredential}}
           instance with its {{DigitalCredential/data}} initialized to
--- a/index.html
+++ b/index.html
@@ -1053,14 +1053,14 @@
       |documentOrigin|.
       </li>
       <li>Present a [=digital credential chooser=] with |requestData| and wait
-      for the user to either:
+        for the user to do one of the following:
         <ul>
           <li>Select a [=digital credential=] and [=holder=] that can fulfill
-          the request,
+            the request.
           </li>
-          <li>Cancel the operation,
+          <li>Cancel the operation.
           </li>
-          <li>Abort the operation via |signal|, or
+          <li>Abort the operation via |signal|.
           </li>
           <li>Encounter a platform-specific error.
           </li>
--- a/index.html
+++ b/index.html
@@ -581,9 +581,9 @@
         <dfn>Digital credential chooser</dfn>
       </dt>
       <dd>
-        A user agent-provided user interface that presents one or more
+        A platform-provided user interface that presents one or more
         [=digital credential/credential requests=] to the user, allowing them
-        to select a [=digital credential=] and [=holder=] that can fulfill the
+        to select a [=digital credential=] or [=holder=] that can fulfill the
         request, or cancel the operation.
         <aside class="note" title="Relationship to credential chooser">
           The [=digital credential chooser=] may be invoked as part of a
@@ -1073,12 +1073,10 @@
       <li>Present a [=digital credential chooser=] with |requestData| and wait
         for the user to do one of the following:
         <ul>
-          <li>Select a [=digital credential=] and [=holder=] that can fulfill
+          <li>Select a [=digital credential=] or [=holder=] that can fulfill
             the request.
           </li>
           <li>Cancel the operation.
-          </li>
-          <li>Abort the operation via |signal|.
           </li>
           <li>Encounter a platform-specific error.
           </li>
@@ -1098,7 +1096,7 @@
       </li>
       <li>If the user cancels the operation or no credential was selected:
         <ol>
-          <li>Let |error| be a newly created {{"AbortError"}} {{DOMException}}.
+          <li>Let |error| be a newly created {{"NotAllowedError"}} {{DOMException}}.
           </li>
           <li>[=credential request coordinator/reject the credential request
           with=] |error| and |promise|.
--- a/index.html
+++ b/index.html
@@ -1530,6 +1530,16 @@
       </dt>
       <dd>
         An [=origin=].
+        <p class="note" title=
+        "Why both origins are included in the request context">
+          The [=request context/document origin=] can differ from the [=request
+          context/top origin=] when the API is called from a cross-origin
+          [^iframe^] (e.g., an embedded verification service). Including both
+          origins allows the [=digital credential chooser=] and [=holder=] to
+          make informed trust decisions about which party is actually
+          requesting the credential, not just which top-level site is hosting
+          the request.
+        </p>
       </dd>
     </dl>
     <h4>
--- a/index.html
+++ b/index.html
@@ -1165,9 +1165,9 @@
               </li>
             </ol>
           </li>
-          <li>Let |protocol:DigitalCredentialProtocol| be the
-          {{DigitalCredentialProtocol}} [=enumeration value=] that matches the
-          [=digital credential/protocol identifier=] used in the exchange.
+          <li>Let |protocol:DigitalCredentialProtocol| be the [=enumeration
+          value=] of the [=digital credential/protocol identifier=] used in the
+          exchange.
           </li>
           <li>Let |credential| be a newly created {{DigitalCredential}}
           instance with its {{DigitalCredential/data}} initialized to
--- a/index.html
+++ b/index.html
@@ -1041,13 +1041,13 @@
     </h3>
     <p>
       To <dfn data-dfn-for="credential request coordinator">present the
-      credential request</dfn> given a [=Document=] |document|, a [=sequence=]
-      of validated credential requests |validatedRequests|, a {{Promise}}
+      credential request</dfn> given a [=Document=] |document|, a [=list=] of
+      validated credential requests |validatedRequests|, a {{Promise}}
       |promise:Promise|, and an optional {{AbortSignal}} |signal|:
     </p>
     <ol class="algorithm">
-      <li>Let |topOrigin| be |document|'s [=relevant settings object=]'s
-      [=environment settings object/origin=].
+      <li>Let |topOrigin| be |document|'s [=top-level traversable=]'s
+      [=navigable/active document=]'s [=Document/origin=].
       </li>
       <li>Let |documentOrigin| be |document|'s [=Document/origin=].
       </li>
--- a/index.html
+++ b/index.html
@@ -249,7 +249,7 @@
         used to request a [=digital credential=] from the user agent. If the
         user agent supports [=digital credential/presentation
         requests|presentation=], it allows the user to select a digital
-        credential through a [=credential chooser=]:
+        credential through a [=digital credential chooser=]:
       </p>
       <pre class="example html" title="Requesting a digital credential">
         &lt;button&gt;Verify Identity&lt;/button&gt;
@@ -556,6 +556,20 @@
       <dd>
         See [=credential request coordinator=].
       </dd>
+      <dt>
+        <dfn>Digital credential chooser</dfn>
+      </dt>
+      <dd>
+        A user agent-provided user interface that presents one or more
+        [=digital credential/credential requests=] to the user, allowing them
+        to select a [=digital credential=] and [=holder=] that can fulfill the
+        request, or cancel the operation.
+        <aside class="note" title="Relationship to credential chooser">
+          The [=digital credential chooser=] may be invoked as part of a
+          broader [=credential chooser=], or as a separate platform-provided
+          user interface. The exact relationship is implementation-defined.
+        </aside>
+      </dd>
     </dl><!--
     // MARK: Protocols
     -->
@@ -853,7 +867,7 @@
         </ol>
       </li>
       <li>[=credential request coordinator/Present the credential request=]
-      with |document|, |validatedRequests|, and |signal|.
+      with |document|, |validatedRequests|, |promise|, and |signal|.
       </li>
       <li>If |document| stops being [=Document/fully active=], [=credential
       request coordinator/abort the credential request=] with an
@@ -960,10 +974,10 @@
           coordinator/interaction state=] to "[=credential request
           coordinator/aborting=]".
           </li>
-          <li>Dismiss the [=credential chooser=].
+          <li>Dismiss the [=digital credential chooser=].
             <p class="note">
-              Dismissal can fail (e.g., if the [=credential chooser=] was
-              destroyed due to memory pressure), but the [=coordinator=]
+              Dismissal can fail (e.g., if the [=digital credential chooser=]
+              was destroyed due to memory pressure), but the [=coordinator=]
               proceeds to complete the credential request regardless.
             </p>
           </li>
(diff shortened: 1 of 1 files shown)


index.html
mohamedamir: Can this be removed then?
```suggestion
```

index.html
```diff
@@ -666,6 +666,120 @@ <h3>
...
+      credential request</dfn> given a [=Document=] |document|, a [=sequence=]
+      of validated credential requests |validatedRequests|, and an optional
+      {{AbortSignal}} |signal|:
+    </p>
+    <ol class="algorithm">
+      <li>Let |requestData| be [=struct=] consisting of |validatedRequests|,
+      |document|'s [=relevant settings object=]'s [=environment settings
+      object/origin=], and |document|'s [=Document/origin=].
+      </li>
+      <li>Present a [=credential chooser=] with |requestData| and wait for the
```
mohamedamir: > [=credential chooser=]

The credential choose linked here is the CredMan chooser which is different from the chooser intended by this algorithm I assume.

index.html
```diff
@@ -666,6 +666,120 @@ <h3>
...
+              The algorithm [=AbortSignal/add|added=] to |signal| by the
+              [=credential request coordinator/prepare credential requests=]
+              steps handle tearing down the [=credential chooser=].
+            </p>
+          </li>
+        </ol>
+      </li>
+      <li>If the user cancels the operation or no credential was selected:
+        <ol>
+          <li>Let |error| be a newly created {{"AbortError"}} {{DOMException}}.
```
mohamedamir: Why `AbortError` when no credential was selected or when the user cancels the operation?

index.html
```diff
@@ -666,6 +666,120 @@ <h3>
...
+          </li>
+          <li>[=credential request coordinator/Complete credential request
+          with=] |error|.
+          </li>
+          <li>Return.
+          </li>
+        </ol>
+      </li>
+      <li>If the platform returns a platform-specific error:
```
mohamedamir: This is the first mention of the platform, right?
So far it sounded like the browser is doing all of it.
Am I overlooking something?

index.html
```diff
@@ -666,6 +666,120 @@ <h3>
...
+          <li>[=credential request coordinator/Complete credential request
+          with=] |error|.
+          </li>
+          <li>Return.
+          </li>
+        </ol>
+      </li>
+      <li>If a [=digital credential=] was selected by the user:
+        <ol>
+          <li>Let |responseData| be a [=string=] [=digital
```
mohamedamir: Is this guaranteed to be a string?
should we document this somewhere that this is the expected format?

index.html
```diff
@@ -666,6 +666,120 @@ <h3>
...
+          credential/presentation response=] or [=string=] [=digital
+          credential/issuance response=] returned by the [=holder=].
+          </li>
+          <li>Let |parsedDataOrError| be the result of [=parse a JSON string to
```
mohamedamir: Maybe rename to avoid confusion with the request data?
```suggestion
          <li>Let |parsedResponseDataOrError| be the result of [=parse a JSON string to
```

index.html
```diff
@@ -666,6 +666,120 @@ <h3>
...
+              </li>
+            </ol>
+          </li>
+          <li>If |parsedDataOrError| is not an [=object=]:
+            <ol>
+              <li>Let |error| be a newly created {{TypeError}}.
+              </li>
+              <li>[=credential request coordinator/Complete credential request
+              with=] |error|.
+              </li>
```
mohamedamir: Do you think it will clearer this way?
```suggestion
              <li>[=credential request coordinator/Complete credential request
              with=] a {{TypeError}}.
              </li>
```
I don't strong opinion, I am genuinely asking :-)
If so, we might change others to be equally brief.
WDYT?

mohamedamir (commented):
I like the direction in general, I have added a couple of comments

Thank you, @marcoscaceres

index.html
```diff
@@ -666,6 +666,120 @@ <h3>
...
+      Present the credential request
+    </h3>
+    <p>
+      To <dfn data-dfn-for="credential request coordinator">present the
+      credential request</dfn> given a [=Document=] |document|, a [=sequence=]
+      of validated credential requests |validatedRequests|, and an optional
+      {{AbortSignal}} |signal|:
+    </p>
+    <ol class="algorithm">
+      <li>Let |requestData| be [=struct=] consisting of |validatedRequests|,
```
jschanck:
```suggestion
      <li>Let |requestData| be a [=struct=] consisting of |validatedRequests|,
```

index.html
```diff
@@ -666,6 +666,120 @@ <h3>
...
+      <li>Present a [=credential chooser=] with |requestData| and wait for the
+      user to either:
+        <ul>
+          <li>Select a [=digital credential=] and [=holder=] that can fulfill
+          the request, or
+          </li>
+          <li>Cancel the operation.
+          </li>
+        </ul>
+      </li>
```
jschanck: The text below this makes it clear that we might exit the credential chooser due to the abort controller or an error. Those cases should be mentioned here.

index.html
marcoscaceres: In both cases, "the operation was aborted [by the user]." I don't think we should distinguish them. User cancelling the operation is generally an `AbortError`.

index.html
marcoscaceres: Ok, yeah... this is where the "digital wallet" concept would have helped.... I guess even then, it would still be "show a UI to select a digital wallet with request data" and the digital wallet itself is a [=credential chooser=]. 

Could you help me with the wording here, @mohamedamir ... or we can come up with something together when we next chat?

index.html
marcoscaceres: You are right. However, presenting the UI and so on was presumedly done by the platform too... should clarify that.  

For what it's worth, [in WebKit](https://github.com/WebKit/WebKit/blob/f463a6d642dd4ffc2f67cb369c3eabadd194a1f9/Source/WebKit/UIProcess/DigitalCredentials/WKDigitalCredentialsPicker.mm#L387), we have the following errors coming back potentially from the OS (from the "Identity Document Framework", which is provided by the OS/platform): 

```C++
    switch (error.code) {
    case WKIdentityDocumentPresentmentErrorNotEntitled:
        exceptionData = { ExceptionCode::NotAllowedError, "Not allowed because not entitled."_s };
        break;
    case WKIdentityDocumentPresentmentErrorInvalidRequest:
        exceptionData = { ExceptionCode::TypeError, "Invalid request."_s };
        break;
    case WKIdentityDocumentPresentmentErrorRequestInProgress:
        exceptionData = { ExceptionCode::InvalidStateError, "Request already in progress."_s };
        break;
    case WKIdentityDocumentPresentmentErrorCancelled:
        exceptionData = { ExceptionCode::AbortError, "Request was cancelled."_s };
        break;
```

index.html
marcoscaceres: I think it has to be a string, as we have to:

1. Check if JSON parsable/serializable (otherwise `.toJSON()` would break).
2. Create the JS Object in the right JS realm (which obviously no native app can do for us). 
3. Check it is a JS Object (because the WebIDL requires it for `.data`).

index.html
marcoscaceres: Good point, @jschanck... I'll describe them.

index.html
mohamedamir: Here is one proposal:
What do you think that we define in our spec a `digital credentials picker`
And define as the UI displayed by the underlying platform together with digital wallets to let the user select a digital credentials in response to a digital credential presentation request.

And we can then refer to that one in our spec.
WDYT?

Happy to discuss/brainstorm in our weekly as well

index.html
mohamedamir: I disagree :-)
in webauthn [get-algorithm-step 16](https://www.w3.org/TR/webauthn-2/#sctn-getAssertion)
```
If the user exercises a user agent user-interface option to cancel the process, 
For each authenticator in issuedRequests invoke the authenticatorCancel operation on authenticator 
and remove authenticator from issuedRequests. 
Return a DOMException whose name is "NotAllowedError".
```

and even more explicit in the [level 3](https://w3c.github.io/webauthn/#sctn-get-request-exceptions)
```
AbortError The ceremony was cancelled by an AbortController. 
See § 5.6 Abort Operations with AbortSignal and § 1.3.4 Aborting Authentication Operations.
.
.
.
NotAllowedError A catch-all error covering a wide range of possible reasons, 
including common ones like the user canceling out of the ceremony. 
Some of these causes are documented throughout this spec, while others are client-specific.
```
And Firefox even fixed a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1621377) related to that for webauthn.

[CredMan](https://www.w3.org/TR/credential-management-1/#:~:text=This%20algorithm%20returns%20either%20null%20if%20the%20user%20chose%20not%20to%20share%20a%20credential%20with%20the%20site%2C%20a%20Credential%20object%20if%20the%20user%20chose%20a%20specific%20credential%2C%20or%20a%20Credential%20interface%20object%20if%20the%20user%20chose%20a%20type%20of%20credential.) however, returns `null` when cancelling the credential chooser IIUC.

But I do think we shouldn't mix both, user cancelling and developer cancelling the request

index.html
mohamedamir: Apart from the AbortError being discussed in another comment thread, I think it's always the platform that does the credential selection.
Even in WebKit, IIUC, it is the platform that does it.
Maybe we should make it clearer in the text of the algorithm in earlier steps?
I don't know, it might be surprising to some when first hear about the platform here and not sure about its role.

WDYT?

index.html
marcoscaceres: This spec is not Web Auth and we can do better. Using a catchall is not a good thing. It means developers will rely on error messages. People criticize Cred Man and Web Auth AFAIK for not using a range of exceptions. 

We can elevate this to the TAG if need be, but seems like a not great use of time. 

I’m strongly of the opinion that this should be an AbortError. We have a great range of exception types for a reason - we should use them.

index.html
marcoscaceres: Yeah, that makes sense. Open to suggestions or can try to clarify things.

index.html
```diff
@@ -666,6 +666,118 @@ <h3>
...
+      credential request</dfn> given a [=Document=] |document|, a [=sequence=]
+      of validated credential requests |validatedRequests|, and an optional
+      {{AbortSignal}} |signal|:
+    </p>
+    <ol class="algorithm">
+      <li>Let |requestData| be a [=struct=] consisting of |validatedRequests|,
+      |document|'s [=relevant settings object=]'s [=environment settings
+      object/origin=], and |document|'s [=Document/origin=].
+      </li>
+      <li>Present a [=credential chooser=] with |requestData| and wait for the
```
marcoscaceres: We might need to clarify what happens when there's multiple matching requests or equivalent requests that are valid to present.

index.html
marcoscaceres: Yep, I'm fine with that.

index.html
marcoscaceres: discussed with @mohamedamir ... will switch `WKIdentityDocumentPresentmentErrorNotEntitled` to `NotAllowedError`. 

For default case where the error is unknown, we should default to "OperationError".

I'll rewrite these in a general form and @mohamedamir will check if they align with Android's platform errors.

index.html
```diff
@@ -748,6 +748,121 @@ <h3>
...
+      <li>Let |requestData| be [=struct=] consisting of |validatedRequests|,
+      |document|'s [=relevant settings object=]'s [=environment settings
+      object/origin=], and |document|'s [=Document/origin=].
+      </li>
+      <li>Present a [=credential chooser=] with |requestData| and wait for the
+      user to either:
+        <ul>
+          <li>Select a [=digital credential=] and [=holder=] that can fulfill
+          the request, or
+          </li>
```
marcoscaceres: Discussed with @mohamedamir, we are going to slot in an issue here for "[Presenting with CTAP #456](https://github.com/w3c-fedid/digital-credentials/issues/456)" so it's clear when CTAP comes into play. And we can craft text around the requirements for cross device / cross platform support here of CTAP and potentially the protocols. 

note that this is independent of "Mandate one presentation protocol MUST be supported" #454.

index.html
```diff
@@ -748,6 +748,121 @@ <h3>
...
+      Present the credential request
+    </h3>
+    <p>
+      To <dfn data-dfn-for="credential request coordinator">present the
+      credential request</dfn> given a [=Document=] |document|, a [=sequence=]
+      of validated credential requests |validatedRequests|, and an optional
+      {{AbortSignal}} |signal|:
+    </p>
+    <ol class="algorithm">
+      <li>Let |requestData| be [=struct=] consisting of |validatedRequests|,
```
marcoscaceres: Discussed with @mohamedamir and @simoneonofri... I'm going to split this out into two variables, so it's more clear. 

[In WebKit](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/identity/DigitalCredentialsRequestData.h), this looks like:

```C++
struct DigitalCredentialsRequestData {
    Vector<ValidatedMobileDocumentRequest> requests;
    SecurityOriginData topOrigin;
    SecurityOriginData documentOrigin;
};
```

index.html
```suggestion
To <dfn data-dfn-for="credential request coordinator">reject request with</dfn> a (JavaScript Value) |error| and a {{Promise}} |promise:Promise|:
```

marcoscaceres: Ooops... wrong branch!

TallTed: > Ooops... wrong branch!

You can delete such posts.

index.html
marcoscaceres: The [=digital credential chooser=] is responsible for resolving this. When there are multiple matching or equivalent requests, the chooser presents them to the user and returns exactly one response. By design, the platform API enforces a single response per invocation (e.g., [WebKit's implementation](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/DigitalCredentials/WKDigitalCredentialsPicker.mm)).

Note that the response blob itself (e.g. ISO mdoc, OID4VP) may contain multiple documents — but that is a protocol-level concern, not an API-level one.

The broader question of requesting credentials across multiple independent wallets is tracked in #222.

index.html
```diff
@@ -1015,19 +1029,189 @@ <h3>
...
+          <li>Select a [=digital credential=] and [=holder=] that can fulfill
+          the request,
+          </li>
+          <li>Cancel the operation,
+          </li>
+          <li>Abort the operation via |signal|, or
+          </li>
+          <li>Encounter a platform-specific error.
+          </li>
+        </ul>
```
```suggestion
      <li>Present a [=digital credential chooser=] with |requestData| and wait
        for the user to do one of the following:
        <ul>
          <li>Select a [=digital credential=] and [=holder=] that can fulfill
            the request.
          </li>
          <li>Cancel the operation.
          </li>
          <li>Abort the operation via |signal|.
          </li>
          <li>Encounter a platform-specific error.
          </li>
        </ul>
```

TallTed:
There are many other "following steps" and similar clarity-improving changes that I would ordinarily suggest. Knowing that Copilot and other contributors are going to fight these, I'm refraining until this editorial question is settled, for now or for ever.

hlflanagan:
Discussed [15 April 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-04-15-DCAPI-B-notes.md#define-present-credential-requests-algorithm419). Will bring back to the group when the reviewers approve the PR.

index.html
```diff
@@ -577,6 +577,20 @@ <h2>
       <dd>
         See [=credential request coordinator=].
       </dd>
+      <dt>
+        <dfn>Digital credential chooser</dfn>
+      </dt>
+      <dd>
+        A user agent-provided user interface that presents one or more
```
```suggestion
        A platform-provided user interface that presents one or more
```

index.html
```diff
@@ -577,6 +577,20 @@ <h2>
...
+        [=digital credential/credential requests=] to the user, allowing them
+        to select a [=digital credential=] and [=holder=] that can fulfill the
```
mohamedamir: For issuance, we select a holder not a credential, I would go for or here
```suggestion
        to select a [=digital credential=] or [=holder=] that can fulfill the
```

index.html
```diff
@@ -1033,19 +1047,189 @@ <h3>
           </li>
         </ol>
       </li>
-    </ol>
+    </ol><!--
+    // MARK: Present the credential request
+    -->
     <h3>
       Present the credential request
```
mohamedamir: "Present" is an overladed term here given credential presentation.
Maybe process is less confusing and covers both presentation and issuance nicely?

```suggestion
      Process the credential request
```

mohamedamir: And if you agree, other place in this PR too.

index.html
```diff
@@ -1033,19 +1047,189 @@ <h3>
...
+      </li>
+      <li>Let |requestData| be a new [=request context=] whose [=request
+      context/requests=] is |validatedRequests|, [=request context/top origin=]
+      is |topOrigin|, and [=request context/document origin=] is
+      |documentOrigin|.
+      </li>
+      <li>Present a [=digital credential chooser=] with |requestData| and wait
+        for the user to do one of the following:
+        <ul>
+          <li>Select a [=digital credential=] and [=holder=] that can fulfill
```
```suggestion
          <li>Select a [=digital credential=] or [=holder=] that can fulfill
```

index.html
```diff
@@ -1033,19 +1047,189 @@ <h3>
...
+              The algorithm [=AbortSignal/add|added=] to |signal| by the
+              [=credential request coordinator/prepare credential requests=]
+              steps handle tearing down the [=digital credential chooser=].
+            </p>
+          </li>
+        </ol>
+      </li>
+      <li>If the user cancels the operation or no credential was selected:
+        <ol>
+          <li>Let |error| be a newly created {{"AbortError"}} {{DOMException}}.
```
```suggestion
          <li>Let |error| be a newly created {{"NotAllowedError"}} {{DOMException}}.
```

index.html
```diff
@@ -1033,19 +1047,189 @@ <h3>
...
+          credential/issuance response=] returned by the [=holder=].
+            <aside class="note" title="Why the response is a string">
+              The response is a string because the user agent is responsible
+              for parsing it into a JavaScript object in the correct realm,
+              verifying it is valid JSON, and confirming the parsed value is an
+              object, as required by the {{DigitalCredential/data}} attribute.
+            </aside>
+          </li>
+          <li>Let |parsedDataOrError| be the result of [=parse a JSON string to
+          a JavaScript value=] passing |responseData|.
```
```suggestion
          <li>Let |parsedResponseDataOrError| be the result of [=parse a JSON string to a JavaScript value=] 
          given |responseData| and |document|'s [=relevant realm=].
```

index.html
```diff
@@ -1033,19 +1047,189 @@ <h3>
...
+      </li>
+      <li>Present a [=digital credential chooser=] with |requestData| and wait
+        for the user to do one of the following:
+        <ul>
+          <li>Select a [=digital credential=] and [=holder=] that can fulfill
+            the request.
+          </li>
+          <li>Cancel the operation.
+          </li>
+          <li>Abort the operation via |signal|.
```
mohamedamir: The text before this list says "wait for the user to do one of the following:"
The user will never abort the operation via signal.
I think we can remove this from that list!

index.html
```diff
@@ -1033,19 +1047,189 @@ <h3>
...
+          </li>
+          <li>Return.
+          </li>
+        </ol>
+      </li>
+      <li>If a [=digital credential=] was selected by the user:
+        <ol>
+          <li>Let |selectedRequest| be the item in |validatedRequests| that was
+          used to invoke the [=holder=] for this exchange.
+          </li>
```
mohamedamir: This happens completely in the platform, and the user agent isn't expected to have insights into which request is forwarded to the holder.

mohamedamir: To be more clear, I suggest to call this the protocol returned by the platform without making assumption on what it means.
WDYT?

index.html
```diff
@@ -1033,19 +1047,189 @@ <h3>
...
+          credential/presentation response=] or [=string=] [=digital
+          credential/issuance response=] returned by the [=holder=].
+            <aside class="note" title="Why the response is a string">
+              The response is a string because the user agent is responsible
+              for parsing it into a JavaScript object in the correct realm,
+              verifying it is valid JSON, and confirming the parsed value is an
+              object, as required by the {{DigitalCredential/data}} attribute.
+            </aside>
+          </li>
+          <li>Let |parsedDataOrError| be the result of [=parse a JSON string to
```
mohamedamir: Should we parse the JSON here (off the main thread) or better move this to the inside the queued task later below?

index.html
```diff
@@ -577,6 +577,20 @@ <h2>
       <dd>
         See [=credential request coordinator=].
       </dd>
+      <dt>
+        <dfn>Digital credential chooser</dfn>
+      </dt>
```
mohamedamir, line 582: Thank you for adding this!
Helps removing a lot of confusion!

mohamedamir (changes requested):

index.html
```diff
@@ -1338,6 +1523,43 @@ <h3>
...
+      </dd>
+      <dt>
+        <dfn data-dfn-for="request context">top origin</dfn>
+      </dt>
+      <dd>
+        An [=origin=].
+      </dd>
+      <dt>
+        <dfn data-dfn-for="request context">document origin</dfn>
+      </dt>
```
mohamedamir: The current [CTAP protocol spec](https://fidoalliance.org/specs/fido-v2.2-ps-20250714/fido-client-to-authenticator-protocol-v2.2-ps-20250714.html#sctn-feature-descriptions-jsonBasedMessages-request-props) accepts only one origin.
How are we going to transfer this request context that contains both the top-level origin as well as the document origin in cross-device flows?
This would require CTAP spec change.

@leecam FYI

index.html
TallTed: Makes sense. Can't make that suggestion in this thread because https://github.com/w3c-fedid/digital-credentials/pull/419#discussion_r3110895045 only references the opening `<li>; it doesn't include the closing `</li>`. So I'll make another suggestion.

index.html
```diff
@@ -1033,19 +1047,189 @@ <h3>
...
+      <li>Present a [=digital credential chooser=] with |requestData| and wait
+        for the user to do one of the following:
+        <ul>
+          <li>Select a [=digital credential=] and [=holder=] that can fulfill
+            the request.
+          </li>
+          <li>Cancel the operation.
+          </li>
+          <li>Abort the operation via |signal|.
+          </li>
```
TallTed: As suggested in https://github.com/w3c-fedid/digital-credentials/pull/419/files#r3110895045
```suggestion
```

hlflanagan:
Discussed at the [20 April 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-04-20-DCAPI-A-notes.md#aob) call. Note the idea to create a list of v2 (breaking changes) items for DC API, including calling origin and top level origin. Will bring back to next week's call.

index.html
marcoscaceres: The "prepare" step really does the "processing", this algorithm is really about managing the presentation aspect and the errors that can occur with that. 

Happy to change the other one, or look at them as a whole as a followup as it's an editorial fix up.

marcoscaceres:
@mohamedamir all suggestions applied. 

@hlflanagan, @mohamedamir filed a bug about the iframe's origin #504 to fix in the future.  

I think we are done 🥳 

Confirmed that apart from a few error codes this matches 1:1 both WebKit and Chrome.

index.html
marcoscaceres: Initiate..

marcoscaceres: @copilot can you change "Process the credential request" to "Initiate the credential request"?

timcappalli (approved):
LGTM with the name change (initiate)

mohamedamir (approved):
LGTM after replacing all mentions of "present" with "initiate"

index.html
```diff
@@ -1033,19 +1047,197 @@ <h3>
...
+      [=navigable/active document=]'s [=relevant settings object=]'s
+      [=environment settings object/origin=].
+      </li>
+      <li>Let |requestData| be a new [=request context=] whose [=request
+      context/requests=] is |validatedRequests| and [=request context/top-level
+      origin=] is |topLevelOrigin|.
+      </li>
+      <li>[=In parallel=]:
+        <ol>
+          <li>Present a [=digital credential chooser=] with |requestData| and
```
```suggestion
          <li>Display a [=digital credential chooser=] with |requestData| and
```

mohamedamir (approved):

wseltzer:
[29 April minutes](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-04-29-DCAPI-B-notes.md)

PR #420, opened 2025-12-16
Define "prepare credential requests" algorithm

### LINKED ISSUE #159: Handling ongoing operations / subsequent call to get()
The spec is currently missing a model for dealing with subsequent call to get() (i.e., subsequently calling `.get()`). 

We have two options. A subsequent call to get():

 1.  aborts the current get request (AbortError). New call to get() takes over. 
 2. gets aborted or invalidated (with the respective exceptions). Current call continues until it settles.

Given that we always show UI, the implications of 1 could be annoying to users. It would mean it would that it could be used to serve the same purpose as the abort controller. 

Thus, my preference is that we use 2, as it prevents accidental tear down of the presentment UI. Developers can still tear down the UI with the abort controller whenever they choose. 

Thoughts?

### LINKED ISSUE #207: Issuance: multiple requests
Topic: should the .create call allow multiple request objects, and if so, what constraints should be applied?

Consensus from the [2025-03-10 call](https://github.com/WICG/digital-credentials/wiki/2025-03-10-Meeting-Notes#204-httpsgithubcomwicgdigital-credentialspull204-for-issuance) was to allow multiple request objects with the same protocol identifier, and add some spec text stating that to ensure a reasonable experience for the user, all request objects should represent the same logical result for the user (e.g. the same type of credential (mDL, PID, etc)).

Alternative: only one request for each protocol identifier is allowed per call, enforced by the client (NotAllowedError or ConstraintError). NOTE: this could be added at a later date if needed.

[Discussion from call 📝](https://github.com/WICG/digital-credentials/wiki/2025-03-10-Meeting-Notes#204-httpsgithubcomwicgdigital-credentialspull204-for-issuance)

Tagging @samuelgoto @leecam @marcoscaceres @bvandersloot-mozilla for comments.

### LINKED ISSUE #220: Supporting multiple protocols in a single DC API call

### LINKED ISSUE #432: Update get procedure for when UA does not support a protocol
After #401, we need to update the get procedure / processing rules for when a user agent does not support a protocol which is supported by the spec.

marcoscaceres:
Added the "prepare credential requests" algorithm, detailing validation of each request, handling of unsupported protocols, error propagation, and abort signal integration.

Closes #432
Closes #249
Closes #220
Closes #207
Closes #159

- [x] Modified Web platform tests 

Implementation commitment:

- [x] WebKit 
  - https://bugs.webkit.org/show_bug.cgi?id=309625 
  - https://github.com/WebKit/WebKit/pull/56489
- [x] [Chromium](https://crbug.com/489769231)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/420.html" title="Last updated on Mar 11, 2026, 1:24 AM UTC (1994d34)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/420/af9b6fa...1994d34.html" title="Last updated on Mar 11, 2026, 1:24 AM UTC (1994d34)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -763,8 +763,8 @@
     <p>
       To <dfn data-dfn-for="credential request coordinator">prepare credential
       requests</dfn> given a [=Document=] |document:Document|, a [=sequence=]
-      of {{DigitalCredentialGetRequest}} or a [=sequence=] of
-      {{DigitalCredentialCreateRequest}} |requests|, a {{Promise}}
+      of {{DigitalCredentialGetRequest}} values or a [=sequence=] of
+      {{DigitalCredentialCreateRequest}} values |requests|, a {{Promise}}
       |promise:Promise|, and an optional {{AbortSignal}} |signal:AbortSignal|:
     </p>
     <ol class="algorithm">


index.html
```diff
@@ -658,8 +658,116 @@ <h3>
...
+              <li>Set the [=credential request coordinator=] [=credential
+              request coordinator/interaction state=] to "[=credential request
+              coordinator/aborting=]".
+              </li>
+              <li>Request that the platform dismiss the [=credential chooser=]
+              and await confirmation of dismissal.
+              </li>
+              <li>Let |error| be either |signal|'s [=AbortSignal/abort reason=]
+              if one was given, or a newly created {{"AbortError"}}
+              {{DOMException}} otherwise.
```
jschanck:
```suggestion
              <li>Let |error| be |signal|'s [=AbortSignal/abort reason=].
```
`AbortController.abort()` sets the abort reason to a new AbortError by default.

index.html
```diff
@@ -658,8 +658,116 @@ <h3>
...
+              <li>If the [=credential request coordinator=] is not in the
+              "[=credential request coordinator/requesting=]" [=credential
+              request coordinator/interaction state=], return.
+              </li>
+              <li>Set the [=credential request coordinator=] [=credential
+              request coordinator/interaction state=] to "[=credential request
+              coordinator/aborting=]".
+              </li>
+              <li>Request that the platform dismiss the [=credential chooser=]
+              and await confirmation of dismissal.
```
jschanck: Who is requesting and awaiting here?

Maybe: "Have the request coordinator dismiss the credential chooser and await confirmation of dismissal before it returns to the idle state".

index.html
```diff
@@ -658,8 +658,116 @@ <h3>
...
+      <li>If |validatedRequests| [=list/is empty=]:
+        <ol>
+          <li>[=credential request coordinator/Complete credential request
+          with=] a newly created {{TypeError}}.
+          </li>
+          <li>Return.
+          </li>
+        </ol>
+      </li>
+      <li>If |signal| was passed:
```
jschanck: I can imagine a scenario where validating a request requires fetching some network resource or waiting for user interaction. So I think the abort algorithm should be configured before requests are validated.

index.html
```diff
@@ -658,8 +658,116 @@ <h3>
...
+      <li>Let |validatedRequests| be a new empty [=list=].
+      </li>
+      <li>[=List/For each=] |request| of |requests|:
+        <ol>
+          <li>If the [=user agent=] does not support |request|'s [=digital
+          credential/exchange protocol=], [=iteration/continue=].
+          </li>
+          <li>Let |validatedRequestOrError| be the result of validating
+          |request|'s [=digital credential/request data=] according to
+          |request|'s [=digital credential/exchange protocol=]. Validation
```
jschanck:
```suggestion
          |request|'s [=digital credential/exchange protocol=] or other criteria. Validation
```

User agents may need to perform additional validation beyond what's described in the exchange protocol's specification.

(Is this section normative?)

index.html
```diff
@@ -658,8 +658,116 @@ <h3>
...
+                during validation.
+              </p>
+            </aside>
+          </li>
+          <li>If |validatedRequestOrError| is an [=exception=]:
+            <ol>
+              <li>[=credential request coordinator/Complete credential request
+              with=] |validatedRequestOrError|.
+              </li>
+              <li>Return.
```
jschanck: I think `return` rather than `continue` is the right call here, but we might need more flexibility. We `continue` over unsupported exchange protocols above. What if a validation procedure fails due to the presence of an unknown and unprocessable extension in an exchange protocol? Should that be a fatal error for the entire bundle of requests?

index.html
marcoscaceres: Yes, this section is normative. 

Can you help me understand "other criteria"? I don't mind adding "...or other criteria." but we might need a non-normative note clarifying what that means.

index.html
marcoscaceres: Good catch. You are correct.

index.html
marcoscaceres: good question and point. 

That would be determined by the validation above I think (i.e., the WhatEver Protocol would need to state how extensions are handled). 

I think that ("unknown and unprocessable extension") should happen here:
https://github.com/w3c-fedid/digital-credentials/pull/419/changes#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R715

(present the UI and the wallet might choke on what was passed... so it would tell the platform that there's a validation error... that's what we do in WebKit through "Identity Document Services": the wallet might reject the request because of an unknown and unprocessable extension)

index.html
marcoscaceres: Yes, exactly! ❤️

index.html
```diff
@@ -658,8 +658,114 @@ <h3>
...
+    <ol class="algorithm">
+      <li>Let |global| be |document|'s [=relevant global object=].
+      </li>
+      <li>If the [=credential request coordinator=] is not in the "[=credential
+      request coordinator/idle=]" [=credential request coordinator/interaction
+      state=]:
+        <ol>
+          <li>[=Queue a global task=] on the [=DOM manipulation task source=]
+          given |global| to [=reject=] |promise| with an
+          {{"InvalidStateError"}} {{DOMException}}.
```
mohamedamir: I am sure about this one.
I honestly think it should be `NotAllowedError`
I would keep `InvalidStateError` for things about the global state, similar to how it is done in CredMan which uses the `InvalidStateError` when the document isn't fully active for example.

index.html
```diff
@@ -1038,7 +1145,15 @@ <h3>
           forthcoming.
         </aside>
       </li>
-      <li>Return a {{DigitalCredential}}.
+      <li>Let |promise| be [=a new promise=].
```
mohamedamir: Did we agree to create a new promise here or use the one passed by CredMan?

marcoscaceres: CredMan does not pass us the promise, so we have to create our own. We then return that promise and it gets automatically “unwrapped” upon settlement. 

We might still need to add a note to CredMan, but it’s the cleanest option.

index.html
marcoscaceres: Right, but this is literally about an “invalid state” in a state machine, which is why I chose it. Happy to find other examples of similar uses.

index.html
marcoscaceres: Yes, that could happen. Or possibly even just over IPC.

index.html
marcoscaceres: Sorta related… for the TAG, I’m working on general guidance around when the document gets removed and things become non-fully active: https://github.com/w3ctag/design-principles/pull/597

We are going to recommend a mix of AbortError and InvalidState, depending on situation.

index.html
marcoscaceres: Just to be clear, returning it to the Idle state needs to happen at the turn of the event loop.

index.html
jschanck: I'm imagining that the user agent might define its own criteria. It might be configured to block certain requests, e.g. requests for a particular credential attribute, or requests that are signed with a certain algorithm (thinking of the transition to post-quantum signature schemes here), or requests that are presented with certificates from a certain authority.

hlflanagan:
Discussed in [21 January 2026 Series B call](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-01-21-DCAPI-B-notes.md#define-present-credential-requests-algorithm-420)

index.html
marcoscaceres: @mohamedamir, ok to resolve this?

index.html
marcoscaceres: @jschanck, ok to resolve this? (and we handle it in the "present" branch?)

index.html
marcoscaceres: @mohamedamir, ok to resolve this?

index.html
```diff
@@ -1009,10 +1139,11 @@ <h3>
       non-overridable behavior of the API.
     </p>
     <pre class="idl">
+    typedef DigitalCredentialPresentationProtocol DigitalCredentialProtocol;
     [Exposed=Window, SecureContext]
     interface DigitalCredential : Credential {
       [Default] object toJSON();
-      readonly attribute DOMString protocol;
+      readonly attribute DigitalCredentialProtocol protocol;
```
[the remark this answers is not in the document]
marcoscaceres: Yep, which would be fixed by adding  https://github.com/w3c-fedid/digital-credentials/issues/453. However, as there are currently no specified issuance protocols, this is fine.

index.html
```diff
@@ -752,8 +752,137 @@ <h3>
...
+      <li>If |signal| was passed:
+        <ol>
+          <li>Assert: |signal| is not [=AbortSignal/aborted=].
+          </li>
+          <li>[=AbortSignal/Add=] the following algorithm to |signal|:
+            <ol>
+              <li>If the [=credential request coordinator=] is not in the
+              "[=credential request coordinator/requesting=]" [=credential
+              request coordinator/interaction state=], return.
+              </li>
```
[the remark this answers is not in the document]
marcoscaceres: Oh! nice find @copilot!

index.html
mohamedamir: Ah, right!
Sorry for the noise!

index.html
marcoscaceres: Talked to @mohamedamir, agreed to change to NotAllowedError.

index.html
```diff
@@ -762,24 +762,156 @@ <h3>
...
+    <ol class="algorithm">
+      <li>Let |global| be |document|'s [=relevant global object=].
+      </li>
+      <li>If the [=credential request coordinator=] is not in the "[=credential
+      request coordinator/idle=]" [=credential request coordinator/interaction
+      state=]:
+        <ol>
+          <li>[=Queue a global task=] on the [=DOM manipulation task source=]
+          given |global| to [=reject=] |promise| with an
+          {{"InvalidStateError"}} {{DOMException}}.
```
```suggestion
          {{"NotAllowedError"}} {{DOMException}}.
```

index.html
mohamedamir: Great!
Thank you!

index.html
```diff
@@ -762,24 +762,156 @@ <h3>
...
+          {{DigitalCredentialCreateRequest/protocol}}, if |request| is a
+          {{DigitalCredentialCreateRequest}}.
+          </li>
+          <li>If |protocol| does not equal any [=enumeration value=] in
+          {{DigitalCredentialProtocol}}, [=iteration/continue=].
+          </li>
+          <li>If the [=user agent=] does not allow |protocol|,
+          [=iteration/continue=].
+          </li>
+          <li>Let |validatedRequest| be the result of validating |request|'s
```
```suggestion
          <li>Let |validatedRequestOrException| be the result of validating |request|'s
```

index.html
```diff
@@ -762,24 +762,156 @@ <h3>
...
+                particular credential attributes, (b) uses or requires
+                cryptographic algorithms the [=user agent=] is configured not
+                to accept (e.g., as part of algorithm agility or a transition
+                to post-quantum schemes), or (c) relies on certificates or
+                trust anchors that are not accepted by the [=user agent=]'s
+                configured trust decisions.
+              </p>
+            </aside>
+          </li>
+          <li>If |validatedRequest| is an [=exception=]:
```
```suggestion
          <li>If |validatedRequestOrException| is an [=exception=]:
```

index.html
```diff
@@ -762,24 +762,156 @@ <h3>
...
+            <ol>
+              <li>[=credential request coordinator/Complete credential request
+              with=] |promise| and |validatedRequest|.
```
```suggestion
              with error=] passing |promise| and |validatedRequestOrException|.
```

index.html
```diff
@@ -762,24 +762,156 @@ <h3>
...
+              </li>
+              <li>Return.
+              </li>
+            </ol>
+          </li>
+          <li>[=list/Append=] |validatedRequest| to |validatedRequests|.
```
```suggestion
          <li>[=list/Append=] |validatedRequestOrException| to |validatedRequests|.
```

index.html
```suggestion
To <dfn data-dfn-for="credential request coordinator">reject request with</dfn> a (JavaScript Value) |error| and a {{Promise}} |promise:Promise|:
```

index.html
```diff
@@ -762,24 +762,156 @@ <h3>
...
+      <li>
+        <aside class="issue" title="Coming soon">
+          See <a href=
+          "https://github.com/w3c-fedid/digital-credentials/pull/462">Add
+          "Abort the credential request" algorithm</a>.
+        </aside>
+      </li>
+    </ol>
     <h3>
       Complete credential request with error
```
```suggestion
      Reject request with error
```

index.html
```diff
@@ -762,32 +762,165 @@ <h3>
...
+                might apply additional validation criteria based on local
+                policy, configuration, or evolving security considerations. For
+                example, a [=user agent=] might reject a request that (a) seeks
+                particular credential attributes, (b) uses or requires
+                cryptographic algorithms the [=user agent=] is configured not
+                to accept (e.g., as part of algorithm agility or a transition
+                to post-quantum schemes), or (c) relies on certificates or
+                trust anchors that are not accepted by the [=user agent=]'s
+                configured trust decisions.
+              </p>
```
timcappalli: I think this needs deeper discussion. From a developer ergonomics standpoint, a request that isn't valid from a protocol standpoint is very different from a user agent not liking the details of the request. We need to discuss how to address this.

index.html
```diff
@@ -773,8 +773,8 @@ <h3>
       state=]:
         <ol>
           <li>[=Queue a global task=] on the [=DOM manipulation task source=]
-          given |global| to [=reject=] |promise| with an
-          {{"NotAllowedError"}} {{DOMException}}.
+          given |global| to [=reject=] |promise| with an {{"NotAllowedError"}}
```
```suggestion
          given |global| to [=reject=] |promise| with a {{"NotAllowedError"}}
```

index.html
marcoscaceres: Happy to discuss further. 

Can you raise a separate issue outlining your thinking?

I’m thinking we could move Validation into a new sub-section and discuss all the implications there. 

Happy to show members the level of validation we do in WebKit and across all Apple’s Platforms when handling IOS18013 requests: https://developer.apple.com/videos/play/wwdc2025/232/

timcappalli: The concern is less about protocol validation (the previous paragraph) and more about specific parts of a given request (the highlighted paragraph). 

Can you remove this text from this PR until we've discussed this and reached consensus?

index.html
```diff
@@ -762,32 +762,165 @@ <h3>
...
+                might apply additional validation criteria based on local
+                policy, configuration, or evolving security considerations. For
+                example, a [=user agent=] might reject a request that (a) seeks
+                particular credential attributes, (b) uses or requires
+                cryptographic algorithms the [=user agent=] is configured not
+                to accept (e.g., as part of algorithm agility or a transition
+                to post-quantum schemes), or (c) relies on certificates or
+                trust anchors that are not accepted by the [=user agent=]'s
+                configured trust decisions.
+              </p>
```
```suggestion
              <aside class="issue" data-number="472">
              </aside>
```

index.html
marcoscaceres: Done. Also moved the validation to its own section so it's easier to discuss and add to.

Can you please take a look and see if this satisfies your concern?

index.html
```diff
@@ -754,40 +754,182 @@ <h3>
...
+    -->
     <h3>
       Prepare credential requests
     </h3>
     <p>
       To <dfn data-dfn-for="credential request coordinator">prepare credential
       requests</dfn> given a [=Document=] |document:Document|, a sequence of
       {{DigitalCredentialGetRequest}} or {{DigitalCredentialCreateRequest}}
-      |requests|, a {{Promise}} |promise:Promise|, and an optional
+      objects |requests|, a {{Promise}} |promise:Promise|, and an optional
```
```suggestion
      object |requests|, a {{Promise}} |promise:Promise|, and an optional
```

marcoscaceres: Oops, this should be a sequence, not objects. Will fix.

index.html
```diff
@@ -754,36 +754,171 @@ <h3>
...
+          </li>
+          <li>Validate |request|'s [=digital credential/request data=]
+          according to |request|'s [=digital credential/presentation protocol=]
+          or other criteria. Validation requirements are protocol-specific and
+          are outside the scope of this specification:
+            <aside class="note" title="Validation details outside scope">
+              <p>
+                Validation includes verifying |request|'s [=digital
+                credential/request data=] conforms to the requirements of the
+                specified [=digital credential/presentation protocol=]. Please
```
mohamedamir: This can be also an issuance protocol

index.html
```diff
@@ -754,36 +754,171 @@ <h3>
...
+          </li>
+        </ol>
+      </li>
+      <li>Return |validatedRequests|.
       </li>
       <li>If |document| stops being [=Document/fully active=], [=credential
-      request coordinator/abort the credential request=] with an {{"AbortError"}}
-      {{DOMException}}.
+      request coordinator/abort the credential request=] with an
+      {{"AbortError"}} {{DOMException}}.
```
mohamedamir: I don't think this belongs to the validation algorithm, no?

index.html
marcoscaceres:
```suggestion
                specified [=digital credential/presentation protocol=] or [=digital credential/issuance protocol=]. Please
```

index.html
marcoscaceres: oh oops! good catch! copy pasta (or maybe when I rebased?)... fixing.

index.html
marcoscaceres: Filed https://github.com/w3c-fedid/digital-credentials/issues/472

index.html
```diff
@@ -754,30 +754,164 @@ <h3>
...
+          {{DigitalCredentialCreateRequest}}.
+          </li>
+          <li>If |protocol| does not equal any [=enumeration value=] in
+          {{DigitalCredentialProtocol}}, [=iteration/continue=].
+          </li>
+          <li>If the [=user agent=] does not allow |protocol|,
+          [=iteration/continue=].
+          </li>
+          <li>Validate |request|'s [=digital credential/request data=]
+          according to |request|'s [=digital credential/presentation protocol=]
```
marcoscaceres, line 896:
```suggestion
          according to |request|'s [=digital credential/presentation protocol=]
          or [=digital credential/issuance protocol=]
```

index.html
```diff
@@ -754,30 +754,170 @@ <h3>
...
+              request coordinator/active promise=] is not |promise|, return.
+              </li>
+              <li>[=credential request coordinator/Abort the credential
+              request=] |signal|'s [=AbortSignal/abort reason=].
+              </li>
+            </ol>
+          </li>
+          <li>Set the [=credential request coordinator=]'s [=credential request
+          coordinator/abort algorithm=] to |abortAlgorithm|.
+          </li>
```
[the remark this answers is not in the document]
marcoscaceres: This is incorrect, but for a different reason. A a pre-aborted signal are actually taken care of by the request and create steps:

https://www.w3.org/TR/credential-management/#algorithm-request
https://www.w3.org/TR/credential-management/#algorithm-create

Those algorithms are the ones that eventually call into these ones. 

So, I think I'll just add an Assert with a note.

index.html
```diff
@@ -754,30 +754,170 @@ <h3>
...
+          coordinator/abort signal=] to |signal|.
+          </li>
+          <li>Let |abortAlgorithm| be the following algorithm:
+            <ol>
+              <li>If the [=credential request coordinator=]'s [=credential
+              request coordinator/active promise=] is not |promise|, return.
+              </li>
+              <li>[=credential request coordinator/Abort the credential
+              request=] |signal|'s [=AbortSignal/abort reason=].
+              </li>
```
mohamedamir, line 814:
```suggestion
              <li>[=credential request coordinator/Abort the credential
              request=] with |signal|'s [=AbortSignal/abort reason=].
              </li>
```

mohamedamir: or maybe
```suggestion
              <li>[=credential request coordinator/Abort the credential
              request=] given |signal|'s [=AbortSignal/abort reason=].
              </li>
```

index.html
```diff
@@ -754,30 +754,170 @@ <h3>
...
+          </li>
+          <li>Return.
+          </li>
+        </ol>
+      </li>
+      <li>Assert: the [=credential request coordinator=]'s [=credential request
+      coordinator/active promise=] is `null`.
+      </li>
+      <li>Set the [=credential request coordinator=]'s [=credential request
+      coordinator/active promise=] to |promise|.
```
mohamedamir, line 793: Why do we set the `[=credential request coordinator/active promise=]` that early before changing the state to `requests`
My concern is right before this we have an assert that when it is in the idle state, the active promise is null, and we immediately break that assert here.

I see two options here.
Either bring 
```
Set the [=credential request coordinator=] [=credential request
      coordinator/interaction state=] to "[=credential request
      coordinator/requesting=]".
``` 
right before setting the promise.

or we don't set the active promise here, and then abort method will take the active promise as a param (which will actually be more aligned with reject)

WDYT?

index.html
```diff
@@ -754,30 +754,170 @@ <h3>
...
+          </li>
+          <li>Return.
+          </li>
+        </ol>
+      </li>
+      <li>Set the [=credential request coordinator=] [=credential request
+      coordinator/interaction state=] to "[=credential request
+      coordinator/requesting=]".
+      </li>
+      <li>[=credential request coordinator/Present the credential request=]
```
mohamedamir, line 855: High level comment that can be addressed in a follow-up maybe.
This now says `Present the credential request`. The term present is overloaded for "credential presentation".
And the request might be an issuance request
In which case, this step is effectively "Present the issuance request"

Maybe "Process the credential requests"

index.html
```diff
@@ -754,30 +754,170 @@ <h3>
...
+          or other criteria. Validation requirements are protocol-specific and
+          are outside the scope of this specification:
+            <aside class="note" title="Validation details outside scope">
+              <p>
+                Validation includes verifying |request|'s [=digital
+                credential/request data=] conforms to the requirements of the
+                specified [=digital credential/presentation protocol=] or
+                [=digital credential/issuance protocol=]. Please refer to the
+                specification of the specific [=digital credential/presentation
+                protocol=] for details, including potential reasons for
```
```suggestion
                [=digital credential/issuance protocol=]. Please refer to the
                specification of the specific [=digital credential/presentation
                protocol=] or [=digital credential/issuance protocol=] for details, including potential reasons for
```

index.html
```diff
@@ -810,18 +950,17 @@ <h3>
...
       </li>
-      <li>[=credential request coordinator/Complete credential request with=]
-      |activePromise| and |error|.
+      <li>[=credential request coordinator/reject request with=] |error| and
+      |activePromise|.
       </li>
     </ol>
     <h3>
-      Complete credential request with error
+      Reject request with
```
mohamedamir: For consistency?
```suggestion
      Reject the credential request with
```

mohamedamir (approved):
Thanks Marcos!
Looks great!
I have added some comments!
Feel free to resolve async if you agree on them or discuss on our next weekly!

index.html
marcoscaceres: yeah, let's figure that out in the "present" branch... I don't mind what we call it and can make the adjustment here and there.

index.html
```diff
@@ -754,30 +754,170 @@ <h3>
...
+          <li>Let |protocol| be |request|'s
+          {{DigitalCredentialGetRequest/protocol}}, if |request| is a
+          {{DigitalCredentialGetRequest}}, or |request|'s
+          {{DigitalCredentialCreateRequest/protocol}}, if |request| is a
+          {{DigitalCredentialCreateRequest}}.
+          </li>
+          <li>If |request| is a {{DigitalCredentialGetRequest}} and |protocol|
+          does not equal any [=enumeration value=] in
+          {{DigitalCredentialPresentationProtocol}}, [=iteration/continue=].
+          </li>
```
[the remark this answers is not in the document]
marcoscaceres: Good robot! You gone done good. 

Filed https://github.com/w3c-fedid/digital-credentials/issues/475

anowardear062-svg:
Okay thanks

index.html
```diff
@@ -754,30 +754,170 @@ <h3>
...
       Prepare credential requests
     </h3>
     <p>
       To <dfn data-dfn-for="credential request coordinator">prepare credential
-      requests</dfn> given a [=Document=] |document:Document|, a sequence of
-      {{DigitalCredentialGetRequest}} or {{DigitalCredentialCreateRequest}}
-      |requests|, a {{Promise}} |promise:Promise|, and an optional
-      {{AbortSignal}} |signal:AbortSignal|:
+      requests</dfn> given a [=Document=] |document:Document|, a [=sequence=]
+      of {{DigitalCredentialGetRequest}}s or a [=sequence=] of
```
TallTed: It's generally better if pluralisms are completely shown in the same style, so this line should be changed to —
```suggestion
      of {{DigitalCredentialGetRequests}} or a [=sequence=] of
```
— *if* ReSpec (or other inline tooling) will make the underlying link correctly. If some other "aliasing" (e.g., `{{DigitalCredentialGetRequest|DigitalCredentialGetRequests}}`)is needed to prevent the plumbing from leaking, then that should be used.

marcoscaceres: That wouldn’t be correct as they are concrete types (and aliasing them would be more confusing). They have no plural form 😔… it’s why we say they are a sequence of {{X}}.

TallTed: `a sequence of {{X}}` is not correct when {{X}} is a singular word, which I think is why someone added the `s` sometime back.

Correct would include `a sequence of {{X}} values` or similar. I'm not sure what the type of `{{DigitalCredentialGetRequest}}` is, so I can't make a concrete suggestion for this.

index.html
marcoscaceres: Good catch! You're right... the invariant "idle ⟺ no active promise" was being violated. 🤦 

Let's go with option 1: moving the state change to "requesting" earlier, right before setting the active promise. 

I'll need to patch up WebKit.

index.html
marcoscaceres: Great suggestion... indeed! checked and "values" is right. Let's make that change.

index.html
```diff
@@ -754,30 +754,178 @@ <h3>
...
     <p>
       To <dfn data-dfn-for="credential request coordinator">prepare credential
-      requests</dfn> given a [=Document=] |document:Document|, a sequence of
-      {{DigitalCredentialGetRequest}} or {{DigitalCredentialCreateRequest}}
-      |requests|, a {{Promise}} |promise:Promise|, and an optional
-      {{AbortSignal}} |signal:AbortSignal|:
+      requests</dfn> given a [=Document=] |document:Document|, a [=sequence=]
+      of {{DigitalCredentialGetRequest}} or a [=sequence=] of
+      {{DigitalCredentialCreateRequest}} |requests|, a {{Promise}}
+      |promise:Promise|, and an optional {{AbortSignal}} |signal:AbortSignal|:
```
```suggestion
      of {{DigitalCredentialGetRequest}} values or a [=sequence=] of
      {{DigitalCredentialCreateRequest}} values |requests|, a {{Promise}}
      |promise:Promise|, and an optional {{AbortSignal}} |signal:AbortSignal|:
```

index.html
marcoscaceres: Applied suggestion above.

PR #421, opened 2025-12-16
Define "complete credential request with error"

### LINKED ISSUE #130: Define error handling

marcoscaceres:
Closes #130

- [X] Modified Web platform tests - already tested via, for example the AbortController tests.

Implementation commitment:

- [X] WebKit - Matches implementation behavior. 
- [X] Chromium - Matches implementation behavior. 
- [ ] Gecko

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/421.html" title="Last updated on Jan 13, 2026, 12:38 AM UTC (8370402)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/421/625c5b9...8370402.html" title="Last updated on Jan 13, 2026, 12:38 AM UTC (8370402)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -486,6 +486,14 @@
         [=digital credential/presentation request=] by a [=verifier=].
       </dd>
       <dt>
+        <dfn data-dfn-for="digital credential" data-local-lt=
+        "credential response">Credential response</dfn>
+      </dt>
+      <dd>
+        A [=digital credential/presentation response=] or an [=digital
+        credential/issuance response=].
+      </dd>
+      <dt>
         <dfn data-for="digital credential">Credential request</dfn>
       </dt>
       <dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -655,17 +655,33 @@ <h3>
...
-      To be written.
-    </p><!--
+    <ol class="algorithm">
+      <li>Assert: the [=credential request coordinator=]'s [=credential request
+      coordinator/active promise=] is not `null`.
+      </li>
+      <li>[=Queue a global task=] on the [=DOM manipulation task source=] given
+      |document|'s [=relevant global object=] to perform the following steps:
+        <ol>
+          <li>[=Reject=] the [=credential request coordinator=]'s [=credential
```
marcoscaceres, line 687: I need to check, but there might be a cases where, by the time we get here, the promise might be `null`.

For instance, the user hit "esc" and the abort controller aborted at the same time (e.g., a timeout).... in such a case, we can just check if it's null and just do nothing.

index.html
```diff
@@ -655,11 +655,35 @@ <h3>
...
     </p>
+    <ol class="algorithm">
+      <li>Assert: the [=credential request coordinator=]'s [=credential request
+      coordinator/active promise=] is not `null`.
+      </li>
+      <li>[=Queue a global task=] on the [=DOM manipulation task source=] given
+      |error|'s [=relevant global object=] to perform the following steps:
+        <ol>
+          <li>If the [=credential request coordinator=]'s [=credential request
+          coordinator/active promise=] is `null`, then return.
```
mohamedamir, line 685: Didn't we assert it is not `null` already?

marcoscaceres: By the time the task is run, the promise might have been cleared.

mohamedamir: Alright!
We discussed this in person
Chromium has the same logic but done in a different (deeper) layer
Basically the callback that owns the promise is bound to a weak pointer, and then the whole callback object is checked for nullness before invoking it.

Although this logic doesn't exist directly explicitly in the code, we agreed there is a value in keeping it in the spec for completeness especially it is implicitly implemented!

Issue #133, opened 2024-06-24
Consider requiring mitigation of script injection attacks.


mikewest:
In #109, @RByers noted that mandatory encryption could be a path towards mitigating the risk that unexpected/injected code on a site could enable an attacker to extract details of a user's identity.

I'd suggest a) that that risk is real, and b) that you could mitigate it to some extent by requiring the page on which the API executes to have at least minimal defenses against injection attacks by requiring a sufficiently-strong CSP along the lines of https://mikewest.github.io/injection-mitigated/.

RByers:
This is certainly interesting to me, but the details really matter here. I'm OK imposing some burden on adoption, but it's got to be a bar that many would already find reasonable, understandable and achievable. 

Is the [InjectionMitigated] idea being used for any web platform APIs today? Do you have HTTP Archive (or UMA) data indicating what fraction of the web is already meeting the criteria you've outlined?

If you show us a pragmatic bar which many security-sensitive sites already meet, then I'm open to exploring if we could make it a requirement for this API - even if we're the first such API to do so. But precisely defining that pragmatic bar, keeping track of the fraction of the web already meeting it and doing the evangelism for common infrastructure to try to drive up compatibility with the criteria seems like to me to be an independent (and possibly substantial) effort that would need to be a pre-requisite to any API taking a hard dependency, right?

mikewest:
> Is the [InjectionMitigated] idea being used for any web platform APIs today? 

Chromium has an implementation of the attribute that we're using for `getAllScreensMedia()` (see [`media_devices.idl`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/mediastream/media_devices.idl;drc=3f8f9df137702ca101cd084fd1ccb86628276f3e;l=36)) to lock in some of the security properties of IWAs (strict CSP + Trusted Types enforcement) while that feature is in an IWAless OT.

> Do you have HTTP Archive (or UMA) data indicating what fraction of the web is already meeting the criteria you've outlined?

UMA suggests that ~10% of top-level pages have a strict CSP, and ~4% of top-level pages enforce Trusted Types. I suspect there's strong overlap between the two, given the folks involved.

> But precisely defining that pragmatic bar, keeping track of the fraction of the web already meeting it and doing the evangelism for common infrastructure to try to drive up compatibility with the criteria seems like to me to be an independent (and possibly substantial) effort that would need to be a pre-requisite to any API taking a hard dependency, right?

[...]

PR #426, opened 2026-01-08
Addressing Request Tampering

simoneonofri:
Secure Context
 - Adding the related threat as "in-scope"
 - Adding Secure Context as the mitigation of the threat

For more context, please refer to the [Google Doc](https://docs.google.com/document/d/1BpBBiv7GgkGi1_Y7NvyD3Mkalj0g857Qw-aan3NqYwU/edit?tab=t.dilz864boly) containing the Threat Model (which will be published as separate Notes) and the complete Section.

- [x] [Modified Web platform tests](https://github.com/w3c-fedid/digital-credentials/pull/426)

Implementation commitment:

- [x] WebKit - already implemented.
- [x] Chromium -  already implemented.
- [x] Gecko - - already implemented.

Documentation and checks

- [ ] Affects privacy
- [x] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/426.html" title="Last updated on Mar 20, 2026, 12:07 AM UTC (c42d7f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/426/d8337fa...c42d7f2.html" title="Last updated on Mar 20, 2026, 12:07 AM UTC (c42d7f2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1578,9 +1578,9 @@
           <dfn data-local-lt="tampering">Request Tampering</dfn>
         </dt>
         <dd>
-          A malicious script or compromised page attempts to alter a
-          {{DigitalCredentialGetRequest}} or {{DigitalCredentialCreateRequest}}
-          before processing.
+          A network attacker that can inject or modify page content in an
+          insecure context attempts to alter a {{DigitalCredentialGetRequest}}
+          or {{DigitalCredentialCreateRequest}} before processing.
         </dd>
       </dl>
       <h4>


index.html
```diff
@@ -1316,112 +1316,137 @@ <h2>
...
+      </div>
+      <p>
+        The Security Considerations section describes the API's security
+        properties, the threats in scope, the assumptions on which security
+        depends, and the residual threats that remain after the responses are
+        applied. This specification defines requirements for the User Agent's
+        behavior only when mediating credential presentation and issuance.
+      </p>
+      <p>
+        Other security properties that depend on protocols, Wallet
```
```suggestion
        Other security properties that depend on protocols, wallet
```

index.html
```diff
@@ -1316,112 +1316,137 @@ <h2>
...
         <p>
-          The documents listed below outline initial security considerations
-          for Digital Credentials, both broadly and for presentation on the
-          web. Their contents will be integrated into this document gradually.
+          Further analysis and threats are specified in related Threat Models,
+          such as the Threat Model for Decentralized Credentials (at the
+          ecosystem level for Decentralized Identities), the Threat Model for
+          the Web (at the web ecosystem level), RFC 3552 (at the internet
+          level), and the FIDO Security Reference, since the FIDO's Client to
+          Authenticator Protocol (CTAP) protocol is used for cross-device flow.
```
```suggestion
          Authenticator Protocol (CTAP) is used for cross-device flow.
```

index.html
```diff
@@ -1316,112 +1316,137 @@ <h2>
...
-            <a href=
-            "https://github.com/w3c-fedid/digital-credentials/wiki/Horizontal-reviews#self-review-questionnaire-security-and-privacy">
-            TAG Security and Privacy Considerations Questionnaire (WIP)</a>
+          <li>In-scope threats introduced or addressed by the DC API itself.
           </li>
-          <li>
-            <a href=
-            "https://github.com/w3c-cg/threat-modeling/blob/main/models/decentralized-identities.md">
-            Threat Model for Decentralized Identities</a>
+          <li>Out-of-Scope threats handled by protocols, Wallets, OS platform
```
```suggestion
          <li>Out-of-scope threats handled by protocols, wallets, OS platform
```

index.html
```diff
@@ -1316,112 +1316,137 @@ <h2>
...
+        <section>
+          <h4>
+            In-scope threats or attacks
+          </h4>
+          <p>
+            The following threats or attacks arise directly from the API's
+            mediation role:
+          </p>
+          <ul>
+            <li><strong>T1</strong>: Presentation Request Modification (Tampering) — A malicious script or compromised page attempts to alter a DigitalCredentialGetRequest before processing.</li>
```
```suggestion
            <li><strong>T1</strong>: Presentation Request Modification (Tampering) —
              A malicious script or compromised page attempts to alter a
              DigitalCredentialGetRequest before processing.</li>
```

index.html
```diff
@@ -1316,112 +1316,137 @@ <h2>
...
-            "https://github.com/w3c-fedid/digital-credentials/wiki/Horizontal-reviews#self-review-questionnaire-security-and-privacy">
-            TAG Security and Privacy Considerations Questionnaire (WIP)</a>
+          <li>In-scope threats introduced or addressed by the DC API itself.
           </li>
-          <li>
-            <a href=
-            "https://github.com/w3c-cg/threat-modeling/blob/main/models/decentralized-identities.md">
-            Threat Model for Decentralized Identities</a>
+          <li>Out-of-Scope threats handled by protocols, Wallets, OS platform
+          security, or transport layers. Even if Out-of-Scope, they are
```
```suggestion
          security, or transport layers. Even if out-of-scope, they are
```

index.html
```diff
@@ -1316,112 +1316,137 @@ <h2>
...
-        <p class="issue" title="Work in progress">
-          Explain that cross-device issuance or presentation uses a separate
-          protocol that has its own security characteristics.
+        <p>
+          The following mitigations derive from normative requirements already
+          present in the specification.
         </p>
+        <section>
+          <h4>Secure Context</h4>
+          <p>WebIDL [=interfaces=] of the Digital Credential API are only exposed in a Secure Context, thus reducing [=tampering=].</p>
```
```suggestion
          <p>WebIDL [=interfaces=] of the Digital Credential API are only exposed in a
          Secure Context, thus reducing [=tampering=].</p>
```

kdenhartog:
Seems like localhost should be excluded in this API case for secure contexts. Otherwise, there's a potential for abuse where the verifier relies on localhost calls to conduct exfiltration of PII as a means to bypass this enforcement layer (e.g. localhost proxy running or https://securityonline.info/androids-secret-tracking-meta-yandex-abused-localhost-for-user-data/)

npdoty:
I think maybe we should clarify that Secure Context protects only against certain types of tampering (like the network attacker, not against XSS, malicious scripts, etc.).

hlflanagan:
Discussed [2026-02-18 FedID WG call](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-02-18-DCAPI-B-notes.md#security-considerations-secure-context-426). No further questions.

index.html
```diff
@@ -1369,7 +1371,8 @@ <h3>
         normative requirements in the specification.
       </p>
       <p>
-        To be written...
+        WebIDL [=interfaces=] of the Digital Credential API are only exposed in
+        a [=secure contexts=], reducing the risk of [=tampering=].
```
marcoscaceres: @npdoty, something like?: 

```suggestion
        a [=secure contexts=], reducing the risk of [=tampering=] through an [=secure contexts|insecure context=] (e.g., a malicious script being injected through the network).
```

index.html
```diff
@@ -1369,7 +1371,8 @@ <h3>
...
       </p>
```
marcoscaceres, line 1615: Meeting notes: 

- For XSS protection, suggest that RP sign all requests using either ISO-18013, and OpenID4VP except unsigned. 
- Don't send unsigned requests for "important" things (e.g., driver's license vs club membership)
- Consult your ~~doctor~~ issuer 
- The protection will happen also on the wallet, which shouldn't accept unsigned requests.  
- Security consideration for wallet is to also not accept unsigned requests, and to double check they don't have a "developer mode" or backdoor themselves.

simoneonofri: Additional meeting notes:
 - We are considering "signing" as a way to protect against "tampering" thus integrity control.
 - [Secure Context](https://www.w3.org/TR/secure-contexts/) protects from Network Level Attacks (as Marcos specified in the suggestion).
 
 
 This also can be an example on how we are managing this kind of assumptions (e.g., for the Wallet, in the specific case)

c2bo: I believe we should at least recommend signing the request (on the underlying protocol level)

simoneonofri: @c2bo this is the idea, also we're considering encryption (so confidentiality) for privacy (so another threat, not tampering)

c2bo: I'd like to be as concrete as possible - at least for the discussion. This would be my current understanding:
- signing of request (on the underlying protocol layer via trust frameworks)
- signing of response (via the mechanisms provided in the underlying credential formats like deviceSigned for mDoc)
- encryption of response (on the underlying protocol layer via ephemeral keys provided by the verifier in the protocol request)

index.html
TallTed: One edit that makes @marcoscaceres suggestion internally consistent and grammatical. There are other edits that can achieve similar results.
```suggestion
        [=secure contexts=], eliminating the risk of [=tampering=] through
        [=secure contexts|insecure contexts=] (e.g., a malicious script being
        injected through the network).
```

index.html
dlongley: An alternative to signing requests is for the wallet to fetch the request again over TLS. There are pros and cons to this approach, one pro is the potential elimination of additional trust frameworks vs. just checking origins, one con is the potential need to compare against the original request if fidelity there is a requirement. There are others that could be further detailed if every possible mitigation needs this sort of thing, otherwise this approach could just be listed as an option that a protocol could offer to mitigate the concern.

index.html
marcoscaceres: I'm cautious about making absolute claims as most of the time the best we can do is make it super hard. At least, I'd want @mikewest (editor of Secure Context) to give that language an 👍 before including it.

TallTed: Fair enough. One revision, which might still be too strong, but can easily be loosened further --
```suggestion
        [=secure contexts=], reducing the risk of [=tampering=] through
        [=secure contexts|insecure contexts=] (e.g., a malicious script being
        injected through the network). Please refer to the
        [Security Considerations](https://www.w3.org/TR/secure-contexts/#security-considerations)
        of the [[[Secure Contexts]]] specification for more information.
```

marcoscaceres: Ok, yeah, that sounds good... and Secure Context's Security Considerations speak to the limitations of "Secure Contexts" themselves, and "esoteric mechanisms" that can be potentially abused for infiltration. 

So, we should add something along the lines of. 

> "Please refer to the [Security Considerations](https://www.w3.org/TR/secure-contexts/#security-considerations) of the [[[Secure Contexts]]] specification for more information."

@simoneonofri ☝️

TallTed: Added to [previous suggestion, above](https://github.com/w3c-fedid/digital-credentials/pull/426#discussion_r2908277075).

mikewest: Thanks for looping me in, @marcoscaceres. It looks like y'all ended up with "WebIDL [=interfaces=] of the Digital Credential API are only exposed in a [=secure contexts=], reducing the risk of [=tampering=]."

I have two suggestions:

1. Nit: Drop "a" (or make "secure contexts" singular).

2. I think locking the API to secure contexts only addresses the capability of network attackers to alter the page's scripts. It does not address malicious scripts or compromised pages generally. I'd suggest shifting the framing here to "reducing the capability of both passive and active network attackers." We should probably add exported definitions of those terms to the relevant sections in https://www.w3.org/TR/secure-contexts/#threat-models, but for the moment just linking there would probably be enough.

mikewest: (FWIW, I think something like https://github.com/w3c-fedid/digital-credentials/issues/133 would be more appropriate as a mitigation against maliciously injected scripts, but it's a bigger lift for developers than simply addressing the transport layer.)

simoneonofri: @TallTed thank you, "reducing" instead of "minimizing" is a specific terminology we're using in the [TMG](https://www.w3.org/TR/threat-modeling-guide/#consider-responses). I know it's a subtle difference in meaning, and what is in common language does not help, but at TMG we thought it was important to distinguish between “reduce” and “minimize”

simoneonofri: @mikewest thank you so much for the feedback! I also update the Threat Model in parallel.

TallTed: Revised my [earlier suggestion](https://github.com/w3c-fedid/digital-credentials/pull/426#discussion_r2908277075) to use `reducing`.

index.html
```diff
@@ -1369,7 +1371,12 @@ <h3>
         normative requirements in the specification.
       </p>
       <p>
-        To be written...
+        WebIDL [=interfaces=] of the Digital Credential API are only exposed in
+        [=secure contexts=], reducing the risk of [=tampering=] through
+        [=secure contexts|insecure contexts=] (e.g., a malicious script being
+        injected through the network). Please refer to the
+        [Security Considerations](https://www.w3.org/TR/secure-contexts/#security-considerations)
+        of the [[[Secure Contexts]]] specification for more information.
```
simoneonofri:
```suggestion
        of the [[[secure-contexts]]] specification for more information.
```

index.html
```diff
@@ -1603,7 +1605,12 @@ <h3>
         normative requirements in the specification.
       </p>
       <p>
-        To be written...
+        WebIDL [=interfaces=] of the Digital Credential API are only exposed in
+        [=secure contexts=], reducing the risk of [=tampering=] through
+        [=secure contexts|insecure contexts=] (e.g., a malicious script being
```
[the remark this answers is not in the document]
marcoscaceres: this is not correct... though I wish it was.

PR #463, opened 2026-02-04
Add slots for abort management

marcoscaceres:
Add two new slots maintained by the credential request coordinator: an abort signal and an abort algorithm. This is to manage removal of the algorithm from the appropriate signal.

- [ ] <del>Modified Web platform tests</del> - not testable.

Implementation commitment:

- [X] [WebKit](https://github.com/WebKit/WebKit/pull/57819)
- [ ] Chromium
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/463.html" title="Last updated on Feb 18, 2026, 11:48 PM UTC (def35ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/463/c35a8a8...def35ee.html" title="Last updated on Feb 18, 2026, 11:48 PM UTC (def35ee)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -799,6 +799,24 @@
           <li>If the [=credential request coordinator=]'s [=credential request
           coordinator/active promise=] is not |promise|, then return.
           </li>
+          <li>Let |signal| be the [=credential request coordinator=]'s
+          [=credential request coordinator/abort signal=].
+          </li>
+          <li>Let |abortAlgorithm| be the [=credential request coordinator=]'s
+          [=credential request coordinator/abort algorithm=].
+          </li>
+          <li>If |signal| is not `null` and |abortAlgorithm| is not `null`:
+            <ol>
+              <li>[=AbortSignal/Remove=] |abortAlgorithm| from |signal|.
+              </li>
+            </ol>
+          </li>
+          <li>Set the [=credential request coordinator=]'s [=credential request
+          coordinator/abort signal=] to `null`.
+          </li>
+          <li>Set the [=credential request coordinator=]'s [=credential request
+          coordinator/abort algorithm=] to `null`.
+          </li>
           <li>[=Reject=] |promise| with |error|.
           </li>
           <li>Set the [=credential request coordinator=]'s [=credential request


index.html
```diff
@@ -669,6 +669,16 @@ <h2>
...
+    <p>
+      The [=credential request coordinator=] maintains an <dfn data-dfn-for=
+      "credential request coordinator">abort signal</dfn>, which the user agent
+      initializes as `null`.
+    </p>
+    <p>
+      The [=credential request coordinator=] maintains an <dfn data-dfn-for=
+      "credential request coordinator">abort algorithm</dfn>, which the user
+      agent initializes as `null`.
+    </p>
```
mohamedamir, line 690: I am sorry, but why is this needed?

marcoscaceres: Sorry, I should have provided more context. The reason this is needed is that [[=AbortSignal/add=]](https://dom.spec.whatwg.org/#abortsignal-add) doesn't return a handle or ID for the algorithm that gets added. So, when it come time to remove that algorithm, the only way to [[=AbortSignal/remove=]](https://dom.spec.whatwg.org/#abortsignal-remove) it is to pass it ( the "abort algorithm") in:

> To remove an algorithm |algorithm| from an [AbortSignal](https://dom.spec.whatwg.org/#abortsignal) signal, [remove](https://infra.spec.whatwg.org/#list-remove) algorithm from signal’s [abort algorithms](https://dom.spec.whatwg.org/#abortsignal-abort-algorithms).

Thus, the above basically represents a weak pointer to the abort algorithm, that we can then use to remove it.

marcoscaceres: @mohamedamir, I added the actual removal below, for when "Complete credential request with error" runs. Hopefully that makes more sense. 

See "prepare" for where the algorithm is added.

mohamedamir: I honestly Marcos, I find this an implementation detail that doesn't add value to the spec, but rather makes the algorithm harder to follow.
Maybe we should discuss in our next 1:1?

marcoscaceres: Sorry if it not clear. This is definitely not an implementation detail though. It defines what happens if the abort signal is called after the promise resolves. 

Per the (DOM) abort controller spec, one needs to define this to use an abort controller. Otherwise signals can stomp on each other. Have a read of:
https://dom.spec.whatwg.org/#abortcontroller-api-integration

Should hopefully help clarify this.

marcoscaceres: Just also want to note that I'm not just documenting what WebKit does here... it was the specification writing process that found a real bug. 

The promise-identity guard ensures correctness (an abort from an earlier signal can’t affect a later operation). The reason we still want to specify removal is lifecycle: the spec explicitly adds an abort algorithm on a potentially long-lived signal, and without defining *removal* it leads to accumulating abort steps and running them later even if they no-op.

In other words, finished operations shouldn’t leave callbacks registered forever and why AbortSignal has both **add** and **remove** steps defined: thus, as we add the abort algorithm during setup, we remove it during completion.

Hope that helps!

mohamedamir: Thanks a lot @marcoscaceres  for the explanation!
TIL

timcappalli (approved):

PR #454, opened 2026-01-28
Require support for all presentation protocols

marcoscaceres:
Closes #439
Closes #474

Requires a user agent to support every presentation protocol listed in the table, rather than a subset, so a verifier can rely on the API whichever credential format or request mode it uses. Issuance protocol support stays recommended.

The following tasks have been completed:

- [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/57557)

Implementation commitment:

- [x] WebKit: https://bugs.webkit.org/show_bug.cgi?id=317545
- [ ] Chromium (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [x] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer
- [ ] Updated digitalcredentials.dev

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/454.html" title="Last updated on Aug 19, 2026, 10:53 PM UTC (194fe45)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/454/ae8abce...194fe45.html" title="Last updated on Aug 19, 2026, 10:53 PM UTC (194fe45)">Diff</a>

PR #474, opened 2026-03-09
All requests must go to client platform for cross-device

timcappalli:
Updates PR #454 to mandate sending requests to the client platform to support cross-device

[protocol-filtering-openid4vp.https.html](https://github.com/web-platform-tests/wpt/blob/master/digital-credentials/protocol-filtering-openid4vp.https.html) and [protocol-filtering-mdoc.https.html](https://github.com/web-platform-tests/wpt/blob/master/digital-credentials/protocol-filtering-mdoc.https.html) currently assert that a request whose protocol the user agent does not allow is filtered out, and that the call rejects with a `TypeError` when none survive. Those tests would need updating to match this requirement.

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

WebKit and Chromium both drop requests with unsupported protocols before dispatching to the platform, so this would be a behavior change in both.

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/474.html" title="Last updated on Aug 19, 2026, 3:08 AM UTC (af949d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/474/836e32e...af949d4.html" title="Last updated on Aug 19, 2026, 3:08 AM UTC (af949d4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -574,12 +574,12 @@
       supported presentation and issuance protocols=].
     </p>
     <p>
-      If a [=user agent=] does not support any {{DigitalCredentialGetRequest}}
-      or {{DigitalCredentialCreateRequest}} objects within the
-      {{DigitalCredentialRequestOptions/requests}} member, it MUST pass
-      the entire DigitalCredentialRequestOptions value to the
-      underlying client platform, or appropriate internal mechanism,
-      for cross-device handling.
+      [=User agents=] MUST pass the entire DigitalCredentialRequestOptions
+      value to the underlying client platform, or appropriate internal mechanism,
+      for cross-device handling, regardless of whether the user agent itself
+      supports the protocol(s) specified in the request options. This allows
+      the platform to determine whether the request can be handled on a different
+      device, even if the current user agent does not support the protocol(s).
     </p>
     <aside class="note" title="Checking which protocols a user agent allows">
       <p>
--- a/index.html
+++ b/index.html
@@ -574,8 +574,8 @@
       supported presentation and issuance protocols=].
     </p>
     <p>
-      [=User agents=] MUST pass the entire DigitalCredentialRequestOptions
-      value to the underlying client platform, or appropriate internal mechanism,
+      [=User agents=] MUST pass the entire `DigitalCredentialRequestOptions`
+      value to the underlying client platform, or appropriate internal mechanism
       for cross-device handling, regardless of whether the user agent itself
       supports the protocol(s) specified in the request options. This allows
       the platform to determine whether the request can be handled on a different
--- a/index.html
+++ b/index.html
@@ -567,11 +567,19 @@
       defined by this specification.
     </p>
     <p>
-      A [=user agent=] MUST implement at least one of the [=digital
+      A [=user agent=] MUST directly implement at least one of the [=digital
       credential/presentation protocols=] &mdash; and it is RECOMMENDED
       that they implement <em>all</em> [=digital credential/presentation
       and issuance protocols=] &mdash; listed in the [=table of
       supported presentation and issuance protocols=].
+    </p>
+    <p>
+      If a [=user agent=] does not support any {{DigitalCredentialGetRequest}}
+      or {{DigitalCredentialCreateRequest}} objects within the
+      {{DigitalCredentialRequestOptions/requests}} member, it MUST pass
+      the entire DigitalCredentialRequestOptions value to the
+      underlying client platform, or appropriate internal mechanism,
+      for cross-device handling.
     </p>
     <aside class="note" title="Checking which protocols a user agent allows">
       <p>
--- a/index.html
+++ b/index.html
@@ -567,7 +567,7 @@
       defined by this specification.
     </p>
     <p>
-      A [=user agent=] MUST directly implement at least one of the [=digital
+      A [=user agent=] MUST implement at least one of the [=digital
       credential/presentation protocols=] &mdash; and it is RECOMMENDED
       that they implement <em>all</em> [=digital credential/presentation
       and issuance protocols=] &mdash; listed in the [=table of


index.html
```diff
@@ -567,12 +567,20 @@ <h2 id="protocols">
...
       and issuance protocols=] &mdash; listed in the [=table of
       supported presentation and issuance protocols=].
     </p>
+    <p>
+      If a [=user agent=] does not support any {{DigitalCredentialGetRequest}}
+      or {{DigitalCredentialCreateRequest}} objects within the
+      {{DigitalCredentialRequestOptions/requests}} member, it MUST pass
+      the entire DigitalCredentialRequestOptions value to the
+      underlying client platform, or appropriate internal mechanism,
+      for cross-device handling.
```
c2bo: I agree that it makes sense to require forwarding if the user agent does not support the request, but shouldn't this always be an option provided to the user? Would something in this direction make sense?

```suggestion
      If a [=user agent=] does not support any {{DigitalCredentialGetRequest}}
      or {{DigitalCredentialCreateRequest}} objects within the
      {{DigitalCredentialRequestOptions/requests}} member, or upon request by
      the user, it MUST pass the entire DigitalCredentialRequestOptions value
      to the underlying client platform, or appropriate internal mechanism,
      for cross-device handling.
```

index.html
```diff
@@ -573,6 +573,14 @@ <h2 id="protocols">
       and issuance protocols=] &mdash; listed in the [=table of
       supported presentation and issuance protocols=].
     </p>
+    <p>
+      [=User agents=] MUST pass the entire DigitalCredentialRequestOptions
+      value to the underlying client platform, or appropriate internal mechanism,
+      for cross-device handling, regardless of whether the user agent itself
+      supports the protocol(s) specified in the request options. This allows
+      the platform to determine whether the request can be handled on a different
+      device, even if the current user agent does not support the protocol(s).
```
```suggestion
      [=User agents=] MUST pass the entire `DigitalCredentialRequestOptions`
      value to the underlying client platform, or appropriate internal mechanism
      for cross-device handling, regardless of whether the user agent itself
      supports the protocol(s) specified in the request options. This allows
      the platform to determine whether the request can be handled on a different
      device, even if the current user agent does not support the protocol(s).
```

index.html
```diff
@@ -573,6 +573,14 @@ <h2 id="protocols">
...
       supported presentation and issuance protocols=].
     </p>
+    <p>
+      [=User agents=] MUST pass the entire `DigitalCredentialRequestOptions`
+      value to the underlying client platform, or appropriate internal mechanism
+      for cross-device handling, regardless of whether the user agent itself
+      supports the protocol(s) specified in the request options. This allows
+      the platform to determine whether the request can be handled on a different
+      device, even if the current user agent does not support the protocol(s).
+    </p>
```
mohamedamir, line 583: IMHO, this paragraph makes sense, but I think it should replace the one above. Having both is confusing and doesn't add much I think!

dlongley: It seems to me that the paragraphs should be combined somehow to cover the case that the user agent might want to handle the request for protocols it supports, but for those it does not want to handle itself, it must pass on.

index.html
```diff
@@ -573,6 +573,14 @@ <h2 id="protocols">
       and issuance protocols=] &mdash; listed in the [=table of
       supported presentation and issuance protocols=].
     </p>
+    <p>
+      [=User agents=] MUST pass the entire `DigitalCredentialRequestOptions`
```
marcoscaceres, line 577:
```suggestion
      [=User agents=] MUST pass the entire {{DigitalCredentialRequestOptions}}
```

index.html
marcoscaceres: Right. This needs to go into the one of the algorithms. Particularly in "prepare" one has the |validatedRequests|, which filters out the unknown ones, so this statement somewhat conflates what the algorithm already does (it passes everything that is valid and supported already).

@timcappalli, can you take a look at the appropriate steps in "prepare" and then how the |validatedRequests| get sent to the "present"? I think needs to hook into there. 

Note however that if there's no platform support, a request can't be passed to the platform. So, even if the browser engine tries to passed it to the platform, the platform will just drop it on the floor (e.g., the [Identity Document Services Framework](https://developer.apple.com/documentation/IdentityDocumentServices) only supports [`ISO18013MobileDocumentRequest`](https://developer.apple.com/documentation/identitydocumentservices/iso18013mobiledocumentrequest)  so it's technically impossible for WebKit to meet this requirement until that changes). 

The spec technically says this is [out of scope](https://www.w3.org/TR/digital-credentials/#scope): 

> Functionality outside the [user agent](https://infra.spec.whatwg.org/#user-agent) is out of scope, including **platform-specific frameworks, native operating system APIs**, standalone applications, and hardware components that store, manage, or process [digital credentials](https://www.w3.org/TR/digital-credentials/#dfn-digital-credential). This includes the user interface and user interactions for selecting a specific credential and obtaining the user's permission to forward a request to the user-selected wallet.

index.html
marcoscaceres: The unknown requests get dropped on the floor: just want to be clear that what needs to be said is “give mDoc and OpenID”… as written, it suggests anything goes.

If the purpose is to try to say an implementation MUST pass OpenID or 18013 requests to the platform in the hope it will pass requests cross-platform, it’s problematic because the platform doesn’t know what to do with such requests. 

That’s the situation today in Safari going through the OS-provide framework. The OS will just reject the request because it doesn’t know what to do or how to send the unknown request cross platform. That’s outside the scope of the spec though.

c2bo: We had a bit of a discussion on this yesterday. My current understanding is that the browser is not in control of the platform, but it should imho, upon the request of the user or if it cannot answer that request type provide the request to underlying layer. If that underlying layer cannot deal with the request, then that is how it is - but at least the expected behaviour for the browser is defined.

Currently we have a too strong coupling of layers imho. Adding support on OS level is another discussion to be had, but that seems to be clearly outside of the scope for this document. With the current text the expectations towards Browsers are at least somewhat clear - that is to forward under certain conditions (user choice & can't answer on its own).

marcoscaceres: It's difficult because, in WebKit, I could hypothetically parse and process an entire OpenID4VP request, but then it would get to where it gets handed off to the OS, but it just gets dropped on the floor (actually, results in a `NotSupportedError`). 

That was partially the original code that I had in WebKit... but because it was dead code (the OS couldn't do anything with it... and there is no way to pass it along!), I removed it. See https://github.com/WebKit/WebKit/pull/55468 ([this in particular](https://github.com/WebKit/WebKit/pull/55468/changes#diff-f927bc80699c5ef8498ec8a1f50a03836ccb860e9b4c69136b3890974f67efe7L380-L382))... we don't want to be shipping dead code in WebKit (or in any software really). 

So, mandating it (as a MUST) doesn't buy anyone anything because if the OS/Platform doesn't support taking arbitrary/OpenID4VP requests, there's not much WebKit can do about it.

c2bo: The expectations towards a browser implementation should still be clearly stated in this document. For me, a requirement to forward the request to OS / cross-device upon user request or if unsupported sounds like a very reasonable requirement.

Whether that requirement is currently fulfilled by the different implementations is another discussion.

marcoscaceres: Right, but that’s literally impossible to do today: There is no means to represent and pass an OpendID4VP request to the OS on Apple’s platforms.

That makes the ask unreasonable and not implementable, no matter how reasonable the ask in the spec might be. It would mean WebKit having to implement the entire CTAP stack, which is not a requirement for any format nor of this specification. 

The only way to reasonably do it would be for the OS to support accepting Openid4VP requests; but mandating conformance requirements on the OS is outside the scope of this specification.

Thus inclusion of such a requirement would yield a Formal Objection on the grounds that it’s not implementable, and that the requirement falls outside the scope of the specification.

index.html
marcoscaceres: Ah, this is not in the right place. This needs to be part of one of the algorithms as to exactly when to pass this.

PR #486, opened 2026-03-25
Relax transient activation requirement for DC API requests

### LINKED ISSUE #478: Proposal: Relax transient user activation requirement for all DC API requests
(This is related to issue #377 )

**Problem**
The strict transient user activation requirement in the DC API breaks complex credential flows. While this is highly visible in Digital Payment Credentials (DPCs) that rely on redirects between merchants, PSPs, and issuers, it inherently affects any DC API request utilizing similar redirect or cross-origin iframe architectures. During these transitions, the browser loses the user activation state, causing legitimate `navigator.credentials.get()` calls to fail with a `NotAllowedError`. 

**Precedent**
The Payment Request API faced this exact same issue. In [PR #1009](https://github.com/w3c/payment-request/pull/1009/files), the spec changed the strict activation requirement to a `MAY`. It allows one "freebie" request without user activation to support redirect flows, and then enforces it for subsequent calls to prevent spam.

**Proposal**
Adopt a similar "one freebie request" approach for the DC API.  This will ensure that all legitimate credential flows that depend on redirects/iframes can function reliably without friction, while still preventing spam on subsequent calls.
Since user mediation is always required, this shouldn't negatively impact user privacy or security.

mohamedamir:
Closes #478

To support redirect flows in Digital Payment Credentials and other complex credential interactions, browsers should allow a "freebie" request without transient user activation. Transient activation is typically lost during transitions like merchant-to-issuer redirects, breaking legitimate user flows.

This change introduces a "digital credential user activation allowance" counter (initially 1) for each relevant global object. If a request is made without transient activation, the counter is checked and decremented. User mediation remains mandatory in all cases, ensuring security is not compromised.

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/486.html" title="Last updated on Mar 26, 2026, 8:08 AM UTC (401a454)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/486/d8337fa...401a454.html" title="Last updated on Mar 26, 2026, 8:08 AM UTC (401a454)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -137,11 +137,13 @@
       issuers, verifiers, and holders to control where potentially sensitive
       personally identifiable information is exposed.
       </li>
-      <li>Require [=transient activation=] to perform [=digital
+      <li>Encourage [=transient activation=] to perform [=digital
       credential/presentation requests=] or [=digital credential/issuance
-      requests=], ensuring that sites cannot silently query for nor issue
-      digital credentials, nor communicate with wallet providers, without the
-      user's active participation and confirmation of each action.
+      requests=]. To support redirect flows, user agents allow a limited number
+      of requests to be made without user activation. In all cases, user
+      mediation ensures that sites cannot silently query for nor issue digital
+      credentials, nor communicate with wallet providers, without the user's
+      active participation and confirmation of each action.
       </li>
       <li>Enable platform-provided credential selection UX when multiple wallet
       applications have credentials that match a [=digital
@@ -1337,6 +1339,23 @@
         "openid4vci-v1",
       };
     </pre>
+    <section id="infrastructure">
+      <h2>
+        Infrastructure
+      </h2>
+      <p>
+        This section defines additional infrastructure and state required for
+        the Digital Credentials API.
+      </p>
+      <h3>
+        Digital Credential User Activation Allowance
+      </h3>
+      <p>
+        Each [=relevant global object=] has an associated <dfn>digital
+        credential user activation allowance</dfn>, which is an integer,
+        initially 1.
+      </p>
+    </section>
     <h2 id="credential-management-integration">
       Integration with [[[credential-management]]]
     </h2><!--
@@ -1387,10 +1406,22 @@
           </li>
         </ol>
       </li>
-      <li>If |global| does not have [=transient activation=], return [=a
-      promise rejected with=] a {{"NotAllowedError"}} {{DOMException}}.
+      <li>If |global| does not have [=transient activation=]:
+        <ol>
+          <li>If |global|'s [=digital credential user activation allowance=] is
+          0, return [=a promise rejected with=] a {{"NotAllowedError"}}
+          {{DOMException}}.
+          </li>
+          <li>Decrement |global|'s [=digital credential user activation
+          allowance=] by 1.
+          </li>
+        </ol>
       </li>
-      <li>[=Consume user activation=] of |global|.
+      <li>Else:
+        <ol>
+          <li>[=Consume user activation=] of |global|.
+          </li>
+        </ol>
       </li>
       <li>Let |promise| be [=a new promise=] in [=this=]'s [=relevant realm=].
       </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1337,6 +1339,23 @@ <h4>
...
+      </h2>
+      <p>
+        This section defines additional infrastructure and state required for
+        the Digital Credentials API.
+      </p>
+      <h3>
+        Digital Credential User Activation Allowance
+      </h3>
+      <p>
+        Each [=relevant global object=] has an associated <dfn>digital
```
marcoscaceres, line 1354: It would be better to attach to the coordinator or the environment settings object instead. 

My preference is just to add a flag to the coordinator.

marcoscaceres: This might have to be on the environment settings object, because the coordinator gets destroyed on navigation.

index.html
```diff
@@ -137,11 +137,13 @@ <h2 class="informative">
       issuers, verifiers, and holders to control where potentially sensitive
       personally identifiable information is exposed.
       </li>
-      <li>Require [=transient activation=] to perform [=digital
+      <li>Encourage [=transient activation=] to perform [=digital
       credential/presentation requests=] or [=digital credential/issuance
-      requests=], ensuring that sites cannot silently query for nor issue
-      digital credentials, nor communicate with wallet providers, without the
-      user's active participation and confirmation of each action.
+      requests=]. To support redirect flows, user agents allow a limited number
```
marcoscaceres, line 142: Explicitly only allow one. 

We need to time box it / cooling period  too... so a web page can't navigate back and forth and keep getting the freeby. 

It's an origin bound.... so it has to wait ???? time.

index.html
marcoscaceres: We will need to figure out where to put this ... there must be something in HTML maybe.

index.html
```diff
@@ -137,11 +137,13 @@ <h2 class="informative">
...
+      of requests to be made without user activation. In all cases, user
+      mediation ensures that sites cannot silently query for nor issue digital
+      credentials, nor communicate with wallet providers, without the user's
+      active participation and confirmation of each action.
```
marcoscaceres, line 146: we need to hook this into the "present" algorithm. We might make this into a note or move this to the Privacy or Security section.

npdoty:
Do we actually want transient activation to persist across site navigations? The user clicks a button on a site they know, trust and understand, that causes a navigation to a new origin and before they see anything else, they immediately get a permission request to transfer sensitive personal data to an origin that they don't know and haven't interacted with? That clearly affects privacy and security.

Does that mean every website can trigger a credential request immediately on pageload as long as the user clicked something on the previous site? Maybe they can't spam endless requests, but the immediate request (based on the reputation of the previous site) is exactly the request that is most likely to succeed in stealing the user's data.

Explicit capability delegation would at least limit the abuse, because the previous site at least intended it and it wasn't just a malicious ad redirect or something.

index.html
TallTed, line 146:
```suggestion
      requests=]. To support redirect flows, user agents allow one such
      request to be made without user activation. In all cases, user
      mediation ensures that sites cannot silently query for nor issue digital
      credentials, nor communicate with wallet providers, without the user's
      active participation and confirmation of each sensitive action.
```

npdoty:
It sounded like the driving use case was not for document navigation, but only for an iframe that navigates (although I still don't understand that case myself). But given the security issues related to top-level navigation and that they aren't necessary, then this proposal should be updated to not apply to that, only apply to iframes that themselves have the `allow="digital-credentials-get"` property.

hlflanagan:
Discussed [15 April 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-04-15-DCAPI-B-notes.md#relax-transient-activation-requirement-for-dc-api-requests486). Looking for more discussion for other possible solutions.

hlflanagan:
Discussed on the [1 June Series A call](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-06-01-DCAPI-A-notes.md#relax-transient-activation-requirement-for-dc-api-requests486).

PR #491, opened 2026-04-08
Simplify and clarify core algorithms

marcoscaceres:
Closes #N/A

Reorders `prepare credential requests` to validate requests before
registering the abort signal handler, removing the now-unreachable
post-validation abort check. Extracts the steps of
`userAgentAllowsProtocol()` into a new internal abstract operation,
`user agent allows protocol`, so both the static method and
`validate credential requests` invoke it directly — avoiding a public
JS method in the internal algorithm call graph. Fixes missing realm
qualifier on the new promise in `[[Create]]`.

- [x] ~~Modified Web platform tests~~ — no observable behaviour change

Implementation commitment:

- [x] WebKit — already implemented; `CredentialRequestCoordinator::prepareCredentialRequest()` validates before registering abort signal handler, `userAgentAllowsProtocol()` is separate from validation path
- [x] Chromium — already implemented; `DiscoverDigitalIdentityCredentialFromExternalSource()` and `CreateDigitalIdentityCredentialInExternalSource()` both validate before `signal->AddAlgorithm()`, `userAgentAllowsProtocol()` is separate from validation path
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/491.html" title="Last updated on Apr 14, 2026, 1:18 AM UTC (b545abe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/491/f6f59ce...b545abe.html" title="Last updated on Apr 14, 2026, 1:18 AM UTC (b545abe)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -793,7 +793,7 @@
       </li>
       <li>Let |validatedRequests| be the result of [=credential request
       coordinator/validate credential requests=] given |requests|. If that
-      [=exception/throws=] an [=exception=] |error|:
+      [=exception/throws=] an [=exception=] |error|, take the following steps:
         <ol>
           <li>[=credential request coordinator/Reject the credential request
           with=] |error| and |promise|.
@@ -802,7 +802,7 @@
           </li>
         </ol>
       </li>
-      <li>If |validatedRequests| [=list/is empty=]:
+      <li>If |validatedRequests| [=list/is empty=], take the following steps:
         <ol>
           <li>[=credential request coordinator/Reject the credential request
           with=] a newly created {{TypeError}} and |promise|.
@@ -811,7 +811,7 @@
           </li>
         </ol>
       </li>
-      <li>If |signal| was passed:
+      <li>If |signal| was passed, take the following steps:
         <ol>
           <li>Assert: |signal| is not [=AbortSignal/aborted=].
             <p class="note">
@@ -1291,8 +1291,8 @@
       underlying platform or provider.
     </p>
     <p>
-      To <dfn>user agent allows protocol</dfn> given a {{DOMString}}
-      |protocol|:
+      To check whether a <dfn>user agent allows protocol</dfn> given a {{DOMString}}
+      |protocol|, take the following steps:
     </p>
     <ol class="algorithm">
       <li>If |protocol| is not an [=enumeration value=] of
@@ -1439,7 +1439,7 @@
       </li>
       <li>[=Consume user activation=] of |global|.
       </li>
-      <li>Let |promise| be [=a new promise=] in [=this=]'s [=relevant realm=].
+      <li>Let |promise| be [=a new promise=] in the [=relevant realm=] of [=this=].
       </li>
       <li>[=credential request coordinator/Prepare credential requests=] with
       |document|, |requests|, |promise|, and |signal|.
--- a/index.html
+++ b/index.html
@@ -786,7 +786,7 @@
       </li>
       <li>Let |validatedRequests| be the result of [=credential request
       coordinator/validate credential requests=] given |requests|. If that
-      [=exception/throws=] an [=exception=] |error|, take the following steps:
+      [=exception/throws=] an [=exception=] |error|, then:
         <ol>
           <li>[=credential request coordinator/Reject the credential request
           with=] |error| and |promise|.
@@ -795,7 +795,7 @@
           </li>
         </ol>
       </li>
-      <li>If |validatedRequests| [=list/is empty=], take the following steps:
+      <li>If |validatedRequests| [=list/is empty=], then:
         <ol>
           <li>[=credential request coordinator/Reject the credential request
           with=] a newly created {{TypeError}} and |promise|.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -791,6 +791,26 @@ <h3>
       <li>Set the [=credential request coordinator=]'s [=credential request
       coordinator/active promise=] to |promise|.
       </li>
+      <li>Let |validatedRequests| be the result of [=credential request
+      coordinator/validate credential requests=] given |requests|. If that
+      [=exception/throws=] an [=exception=] |error|:
```
TallTed: Colons follow a full sentence, not a sentence fragment.
```suggestion
      [=exception/throws=] an [=exception=] |error|, take the following steps:
```

index.html
```diff
@@ -791,6 +791,26 @@ <h3>
...
+        <ol>
+          <li>[=credential request coordinator/Reject the credential request
+          with=] |error| and |promise|.
+          </li>
+          <li>Return.
+          </li>
+        </ol>
+      </li>
+      <li>If |validatedRequests| [=list/is empty=]:
```
```suggestion
      <li>If |validatedRequests| [=list/is empty=], take the following steps:
```

index.html
```diff
@@ -791,6 +791,26 @@ <h3>
...
+        <ol>
+          <li>[=credential request coordinator/Reject the credential request
+          with=] a newly created {{TypeError}} and |promise|.
+          </li>
+          <li>Return.
+          </li>
+        </ol>
+      </li>
       <li>If |signal| was passed:
```
```suggestion
      <li>If |signal| was passed, take the following steps:
```

index.html
```diff
@@ -1302,8 +1291,8 @@ <h4>
       underlying platform or provider.
     </p>
     <p>
-      When this method is invoked, the user agent MUST execute the following
-      algorithm:
+      To <dfn>user agent allows protocol</dfn> given a {{DOMString}}
+      |protocol|:
```
TallTed: The above is not really English. I've thought hard about how to fix it, without finding a solution that keeps the existing `<dfn>` text. The various link backs (like `[= ... =]`) will also need revision.
```suggestion
      To check whether a <dfn>user agent allows protocol</dfn> given a {{DOMString}}
      |protocol|, take the following steps:
```

index.html
```diff
@@ -1446,7 +1439,7 @@ <h3>
       </li>
       <li>[=Consume user activation=] of |global|.
       </li>
-      <li>Let |promise| be [=a new promise=].
+      <li>Let |promise| be [=a new promise=] in [=this=]'s [=relevant realm=].
```
```suggestion
      <li>Let |promise| be [=a new promise=] in the [=relevant realm=] of [=this=].
```

timcappalli (approved):
Looks good after @TallTed's feedback

marcoscaceres:
@copilot can you fix the “user agent allows protocol” references? We renamed the definition so there might be broken internal links.

marcoscaceres:
@TallTed, is it ok we use the “If…, then:” pattern? We use that more consistently throughout W3C/WHATWG specs.

mohamedamir (approved):
Looks good!
Thank you!

marcoscaceres:
I’ll wait for 👍 from @TallTed on the “if.., then:” editorial changes.

TallTed:
@marcoscaceres 

> Modern W3C/WHATWG specs (Fetch, HTML, WebAuthn, Credential Management)
> use "then:" for conditional branches rather than the more verbose "take
> the following steps:". Align with that house style for the three
> conditional list items in prepare credential requests; the top-level
> named algorithm definition retains its existing preamble.

I have no argument with `then:` in *code*, but it doesn't belong in English prose, such as this (and the other listed) specifications.

Copilot (and the authors/approvers of the cited "modern" specs [a not-so-subtle disparagement of other recent W3C specifications!]) doesn't seem to know all the rules of English and its punctuation. Basically, the colon that ends such a paragraph SHOULD (and MUST according to most style guides I've ever worked under) only be placed where a period would be valid.

Even if we accept that the "following" formulations are optional (which I don't), I have found it better to include most such optional constructs as they significantly ease comprehension for readers for whom English is a second (or later!) language.

Regrettably, the [W3C Manual of Style](https://www.w3.org/guide/manual-of-style/) does not say anything about this, and uses both formulations itself.

marcoscaceres:
@TallTed, the "modern" phrasing was Copilot's auto-generated commit message, not my words. Those commit messages won't be in the final merge anyway, but I get why it read badly.

On the substance: the DC API extends Credential Management, and Credential Management already uses the "If X, then:" pattern. I'd rather keep our spec consistent with the one we directly depend on than introduce a different convention. It's a bad practice we picked up from HTML, I guess... which uses it over 300 times!

The dfn preamble rewrite is genuinely better, and we really value your close editorial scrutiny on every pull request. Going to merge with "then:" for Cred Man consistency.

PR #493, opened 2026-04-09
Add Request/Respose Tampering Security Considerations

### LINKED ISSUE #479: Adding Signature as Response for Tampering and consider Encryption
As specified in the [PR](https://github.com/w3c-fedid/digital-credentials/pull/426#pullrequestreview-3971263107), this issue is to add a signature as a mitigation to  tampering as well:

> - signing of request (on the underlying protocol layer via trust frameworks)
> - signing of response (via the mechanisms provided in the underlying credential formats like deviceSigned for mDoc)

Consider also (for a different threat):
> - encryption of response (on the underlying protocol layer via ephemeral keys provided by the verifier in the protocol request)

simoneonofri:
Closes #479

Adds request and response tampering to the threat model and establishes normative integrity requirements for protocols used with the Digital Credentials API.

## Why this matters

The DC API mediates access to digital credentials, including government-issued identity documents. It defines the user agent's mediation behavior, but does not by itself guarantee the end-to-end security of credential exchanges. That depends on the underlying protocols.

Without normative integrity requirements, the spec is silent on whether a protocol must protect requests and responses from modification. This gap leaves users exposed to tampering attacks at the protocol level, which the user agent cannot detect or prevent after handoff.

[RFC 7258](https://www.rfc-editor.org/rfc/rfc7258) (BCP 188) established the IETF consensus that known attacks must be mitigated in protocol design: "pervasive monitoring is a technical attack that should be mitigated in the design of IETF protocols, where possible." The same design principle applies to integrity attacks on credential exchanges. The W3C [Ethical Web Principles](https://www.w3.org/TR/ethical-web-principles/#expression) (§2.6) cite RFC 7258 directly: "Our work should not enable state censorship, surveillance or other practices that seek to limit this freedom."

This PR applies that principle: a user agent MUST NOT support a protocol that lacks integrity mechanisms for requests and responses.

## Changes

- Define Request Tampering and Response Tampering as out-of-scope threats (addressed by protocols, not by this API).
- Add normative requirement: a user agent MUST NOT support a protocol unless it defines mechanisms for request and response integrity.
- Rename "Tampering requests through the networks" to "Network Tampering" to distinguish the in-scope network-level threat from protocol-level tampering.
- Cross-reference the [Protocols](https://w3c-fedid.github.io/digital-credentials/#protocols) section to the new security requirements.

- [x] ~~Modified Web platform tests~~ — no observable behavior change

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [x] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

Follow-up from: #487

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/493.html" title="Last updated on Aug 27, 2026, 6:57 AM UTC (585b91b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/493/0aea14a...585b91b.html" title="Last updated on Aug 27, 2026, 6:57 AM UTC (585b91b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1528,15 +1528,13 @@
       <p>
         The sections that follow describe the API's security properties, the
         threats in scope, the assumptions on which security depends, and
-        residual threats that remain after the mitigations are applied. This
-        specification defines requirements for the user agent's behavior only
-        when mediating [=digital credential/credential responses=].
+        residual threats that remain after the mitigations are applied.
       </p>
       <p class="note">
-        Other security considerations that depend on protocols, wallet
-        implementations, operating systems, or transport security are described
-        as expectations or preconditions, but these are not normatively
-        required by this specification unless already normatively specified.
+        Security considerations that depend on wallet implementations,
+        operating systems, or transport security are described as expectations
+        or preconditions, but these are not normatively required by this
+        specification.
       </p>
       <h3 class="informative">
         Threat Model
@@ -1559,14 +1557,12 @@
       </p>
       <dl>
         <dt>
-          <dfn data-local-lt="tampering">Tampering requests through the
-          networks</dfn>
+          <dfn data-local-lt="tampering">Network Tampering</dfn>
         </dt>
         <dd>
-          An attacker attempts to alter attacker that can inject or modify page
-          content in an insecure context attempts to alter a
-          {{DigitalCredentialGetRequest}} or {{DigitalCredentialCreateRequest}}
-          on the wire.
+          A network attacker injects or modifies page content in an insecure
+          context, altering a {{DigitalCredentialGetRequest}} or
+          {{DigitalCredentialCreateRequest}} in transit.
         </dd>
         <dt>
           <dfn>Unauthorized Cross-Origin Access</dfn>
@@ -1593,8 +1589,8 @@
           <dfn>Request Tampering</dfn>
         </dt>
         <dd>
-          Modification of a [=digital credential/presentation protocols=] or
-          [=digital credential/issuance protocols=] request after it is
+          Modification of a [=digital credential/presentation protocol=] or
+          [=digital credential/issuance protocol=] request after it is
           created. This threat is expected to be addressed by the underlying
           exchange protocol; for example, through signing and validating
           requests.
@@ -1603,11 +1599,11 @@
           <dfn>Response Tampering</dfn>
         </dt>
         <dd>
-          Modification of a [=digital credential/presentation protocols=] or
-          [=digital credential/issuance protocols=] response after it is
+          Modification of a [=digital credential/presentation protocol=] or
+          [=digital credential/issuance protocol=] response after it is
           created. This threat is expected to be addressed by the underlying
           exchange protocol; for example, through signing and validating
-          requests.
+          responses.
         </dd>
       </dl>
       <h3 class="informative">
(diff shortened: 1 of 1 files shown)


marcoscaceres:
When reviewing the changes, please use https://github.com/w3c-fedid/digital-credentials/pull/493/changes?w=1 (hide whitespace)

martinthomson (commented):
I find these changes to be a little imprecise about what they are addressing.

This changes "Request Tampering" to "Network Tampering", which seems like it might be narrowing the scope unnecessarily.  Attacks from the network are generally addressed using HTTPS, making the new scope uninteresting in the extreme.  However, the original scope might have been read to include browsers, CDN operators, and others who had access to the messages.

Perhaps the intent here is to split those aspects out into the new "Data Tampering" section, but that is not particularly clear about which entities might be able to make those modifications.

As for mitigations, the use of digital signatures to counteract tampering is weird.  It's not sufficient to say "just sign it".  You need to have expectations about who is going to sign and with what key.

As for the new requirements on protocols, I can see how some people might have a threat model in which the proposed requirements become necessary, but the request/response integrity arrangement is not one that is consistent with my understanding of the web platform.  A website that asks for information and is given that information is sufficient to address a need.  That suffices.

That suggests that the decision to insist on tamper resistance extends beyond what is normal for the web platform.  That means that stronger justification than what is provided for the requirement is needed.

Again, "just sign it" is an insufficient means of dealing with this problem as well.

marcoscaceres (commented):
Your point that "just sign it" is an insufficient means of dealing with this problem is correct, and the gap you identified is real.

The intent behind the MUST is that the UA can only gate on whether a protocol *defines* integrity mechanisms — it cannot inspect runtime crypto. For government-issued identity documents, that gate matters: the issuer, wallet, user agent, and verifier form a chain of trust that doesn't exist in ordinary web content. But the spec currently requires only that mechanisms *exist*, not that they be adequate. A protocol could define signing with an ephemeral self-generated key and technically comply. That's not the bar we mean to set.

On the naming split: "Network Tampering" (in-transit, TLS handles it) and "Data Tampering" (post-creation at protocol level, TLS does not handle it) are intentionally separate. The rationale for the split needs to be in the text.

Your "who is going to sign and with what key" is the precise question the spec needs to answer. We'd welcome your help specifying what makes a mechanism adequate — the trust relationship requirements are exactly where the text needs more precision.

timcappalli:
> Your "who is going to sign and with what key" is the precise question the spec needs to answer. We'd welcome your help specifying what makes a mechanism adequate — the trust relationship requirements are exactly where the text needs more precision.

I don't agree with this being in scope of the specification. Request signing and trust frameworks are very complex and vary drastically by region and use case.

TallTed:
"Trust" is a dangerous word to use, and is inherently vague. I think there are many ways "trust" might be argued as beyond our charter.

"Cryptographically sign", "cryptographically verify", and the like, and all their details, are much less vague and harder to argue as beyond our charter.

TallTed:
Perhaps correct the PR title from `Request/Respose` to `Request/Response`...

mafalamanero:
> Your point that "just sign it" is an insufficient means of dealing with this problem is correct, and the gap you identified is real.
> 
> The intent behind the MUST is that the UA can only gate on whether a protocol *defines* integrity mechanisms — it cannot inspect runtime crypto. For government-issued identity documents, that gate matters: the issuer, wallet, user agent, and verifier form a chain of trust that doesn't exist in ordinary web content. But the spec currently requires only that mechanisms *exist*, not that they be adequate. A protocol could define signing with an ephemeral self-generated key and technically comply. That's not the bar we mean to set.
> 
> On the naming split: "Network Tampering" (in-transit, TLS handles it) and "Data Tampering" (post-creation at protocol level, TLS does not handle it) are intentionally separate. The rationale for the split needs to be in the text.
> 
> Your "who is going to sign and with what key" is the precise question the spec needs to answer. We'd welcome your help specifying what makes a mechanism adequate — the trust relationship requirements are exactly where the text needs more precision.

marcoscaceres:
@copilot resolve the merge conflicts in this pull request

index.html
```diff
@@ -1986,6 +1991,21 @@ <h4>
         end-to-end security of credential presentation and issuance. The
         following defines the [=out-of-scope threats=] for this specification:
       </p>
+      <dl>
```
mohamedamir, line 1999: This PR adds a `<dl>` for Data Tampering immediately followed by a `<ul>` for Broken cryptography.

Maybe we should unify all to `<dl>`?

index.html
```diff
@@ -1948,12 +1950,15 @@ <h4>
       </p>
       <dl>
         <dt>
-          <dfn data-local-lt="tampering">Request Tampering</dfn>
+          <dfn data-local-lt="tampering">Network Tampering</dfn>
```
mohamedamir: `data-local-lt="tampering"` teampering is overloaded in this document

index.html
```diff
@@ -2072,6 +2092,43 @@ <h3>
...
+      </h4>
+      <p>
+        A user agent MUST NOT support a [=digital credential/presentation
+        protocol=] or [=digital credential/issuance protocol=] unless the
+        protocol defines mechanisms that allow all parties to determine that a
+        request or response has not been modified after creation. In general,
+        this is achieved by signing requests and responses at the protocol
+        level.
+      </p>
+      <p>
```
mohamedamir: Sorry, did we agree on mandating signing?

index.html
```diff
@@ -1907,7 +1911,7 @@ <h2>
         </dd>
       </dl>
     </section>
-    <section class="informative">
+    <section>
```
```suggestion
    <section class="informative">
```

index.html
marcoscaceres: I think this might be redundant with there other PRs that are merged... let's me update and see what copilot says.

index.html
simoneonofri: Ok, I removed data-local-lt="tampering" and renamed the defined term to Network Tampering. The cross-reference now uses [=network tampering=], while protocol-level modification is covered by Protocol and Format Vulnerabilities.

index.html
simoneonofri: Agreed, I kept that structure and removed the separate Data Tampering block, then we're going to change the structure

index.html
simoneonofri: Yes, I removed the normative requirement.

PR #506, opened 2026-04-29
Add security considerations on forged credential chooser requests

### LINKED ISSUE #472: User agent request validation and errors
The prepare steps originally stated that: 

> In addition to protocol-defined requirements, a [=user agent=] might apply additional validation criteria based on local policy, configuration, or evolving security considerations. For example, a [=user agent=] might reject a request that (a) seeks particular credential attributes, (b) uses or requires cryptographic algorithms the [=user agent=] is configured not to accept (e.g., as part of algorithm agility or a transition to post-quantum schemes), or (c) relies on certificates or trust anchors that are not accepted by the [=user agent=]'s configured trust decisions.

It would be good to expand on these.

marcoscaceres:
Related to #472, addresses the gesture verification aspect

Describes how a user agent that runs web content in a separate process from the one handling input can be made to present the digital credential chooser with no user interaction, and where such a user agent should evaluate the existing transient activation check so that a compromised content process cannot forge the answer. This is informative only: the normative check and its consumption already exist in the shared prologue, so no algorithm step is added.

- [ ] <s>Modified Web platform tests</s> - not testable

Implementation commitment:

- [x] WebKit ([Bug 312458](https://bugs.webkit.org/show_bug.cgi?id=312458))
- [ ] Chromium - checks transient activation in the renderer only
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [x] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/506.html" title="Last updated on Aug 27, 2026, 1:41 PM UTC (59d8cb2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/506/1a75475...59d8cb2.html" title="Last updated on Aug 27, 2026, 1:41 PM UTC (59d8cb2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1068,6 +1068,22 @@
       context/requests=] is |validatedRequests| and [=request context/top-level
       origin=] is |topLevelOrigin|.
       </li>
+      <li>The user agent SHOULD independently verify that the request to
+      present the [=digital credential chooser=] was initiated by a
+      user-activated event, using a mechanism that does not solely rely on the
+      [=navigable=] responsible for running the content of the [=top-level
+      traversable=].
+        <aside class="note" title="Defense in depth">
+          In multi-process architectures, the [=consume user activation=] check
+          occurs in the process responsible for running web content (variously
+          called the "content process" or "renderer process"), which could be
+          compromised. By correlating the request with a user interaction event
+          dispatched by a trusted process (e.g., the user agent's own
+          event-handling process), the user agent can provide defense in depth
+          against presenting the [=digital credential chooser=] without a
+          genuine user gesture.
+        </aside>
+      </li>
       <li>[=In parallel=]:
         <ol>
           <li>Display a [=digital credential chooser=] with |requestData| and
--- a/index.html
+++ b/index.html
@@ -1245,36 +1245,6 @@
       context/requests=] is |validatedRequests| and [=request context/top-level
       origin=] is |topLevelOrigin|.
       </li>
-      <li>The user agent SHOULD independently verify that the request to
-      present the [=digital credential chooser=] was initiated by a
-      user-activated event, using a mechanism that does not solely rely on the
-      [=navigable=] responsible for running the content of the [=top-level
-      traversable=].
-        <ol>
-          <li>If the verification fails:
-            <ol>
-              <li>Let |error| be a newly created {{"NotAllowedError"}}
-              {{DOMException}}.
-              </li>
-              <li>[=credential request coordinator/Reject the credential
-              request with=] |error| and |promise|.
-              </li>
-              <li>Return.
-              </li>
-            </ol>
-          </li>
-        </ol>
-        <aside class="note" title="Defense in depth">
-          In multi-process architectures, the [=consume user activation=] check
-          occurs in the process responsible for running web content (variously
-          called the "content process" or "renderer process"), which could be
-          compromised. By correlating the request with a user interaction event
-          dispatched by a trusted process (e.g., the user agent's own
-          event-handling process), the user agent can provide defense in depth
-          against presenting the [=digital credential chooser=] without a
-          genuine user gesture.
-        </aside>
-      </li>
       <li>[=In parallel=]:
         <ol>
           <li>Display a [=digital credential chooser=] with |requestData| and
@@ -2013,6 +1983,17 @@
           underlying operating system or [=credential manager=] by passing
           carefully crafted or malformed protocol requests through the API.
         </dd>
+        <dt>
+          <dfn data-local-lt=
+          "forged requests|content process compromise">Forged Requests from a
+          Compromised Content Process</dfn>
+        </dt>
+        <dd>
+          An attacker who has taken control of the process that runs a page's
+          code attempts to make the [=digital credential chooser=] appear
+          without the user having interacted with the page, by skipping the
+          [=transient activation=] check that runs in that same process.
+        </dd>
       </dl>
       <h4>
         Out of Scope Threats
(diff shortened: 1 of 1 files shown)


timcappalli (approved):

marcoscaceres:
@mohamedamir ping 🙏

index.html
```diff
@@ -1165,6 +1165,22 @@ <h3>
       context/requests=] is |validatedRequests| and [=request context/top-level
       origin=] is |topLevelOrigin|.
       </li>
+      <li>The user agent SHOULD independently verify that the request to
+      present the [=digital credential chooser=] was initiated by a
+      user-activated event, using a mechanism that does not solely rely on the
+      [=navigable=] responsible for running the content of the [=top-level
+      traversable=].
```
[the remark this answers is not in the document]
marcoscaceres: Ok, true. These do need to be rejected gracefully. I’ll check what we do in WebKit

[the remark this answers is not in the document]
marcoscaceres: NotAllowedError https://github.com/WebKit/WebKit/pull/68592/changes

index.html
```diff
@@ -1165,6 +1165,36 @@ <h3>
       context/requests=] is |validatedRequests| and [=request context/top-level
       origin=] is |topLevelOrigin|.
       </li>
+      <li>The user agent SHOULD independently verify that the request to
```
marcoscaceres: Wonder if we should drop the SHOULD. Both Chromium and WebKit do this check.

mohamedamir: Quick clarification on Chromium's status: **Chromium does not currently perform this check in the browser/UI process.**

We previously had a browser-side check, but had to remove it ([crrev.com/c/7487180](https://chromium-review.googlesource.com/c/chromium/src/+/7487180)) due to an IPC race condition:

1. **Synchronous Consumption:** Per spec, the renderer consumes user activation synchronously upon the API call (so JS immediately sees `isActive === false`).
2. **IPC Ordering:** Consuming activation sends an async IPC to clear browser-side activation state *before* (or alongside) the credential request IPC.
3. **False Rejection:** Because the activation is already cleared by the time the browser processes the credential request, a browser-side check fails for legitimate requests.

I am honestly not what is the best way forward from here.

I even checked other APIs, and some check it in the renderer (a.k.a content process) (e.g. WebShare, PaymentsRequest, WebAuthn) and some in the browser (a.k.a UI process) (e.g. FontAccess, FileSystemAccess)
But no other API checks both!

How is the situation on webkit?
Is it different from Chromium?
Maybe it is worth doing that exercise on Webkit to align and then see what we can do!

marcoscaceres: WebKit differs in how the UI process knows a gesture happened, and that is what avoids the race.

The UI process handles input events itself, so it records the time of the last gesture first-hand rather than being told about it by the WebContent process. Nothing the WebContent process sends can change that record, so the consume has no message to race against. It separately records which gesture it has already used for a chooser, so one gesture cannot open two.

That only works because the process presenting the chooser is the same one handling input, which the spec cannot require.

index.html
```diff
@@ -1981,6 +1993,55 @@ <h3>
...
+        <!--
+        // MARK: Forged Requests from a Compromised Content Process
+        -->
+        <h3>
+          Forged Requests from a Compromised Content Process
+        </h3>
+        <p>
+          Some user agents run a page's code in one process and handle
+          keyboard, mouse, and touch input in another, more trusted, process.
+          In those user agents both the [=transient activation=] check and the
```
```suggestion
          In those user agents, both the [=transient activation=] check and the
```

index.html
```diff
@@ -1981,6 +1993,55 @@ <h3>
...
+        </p>
+        <p>
+          A multi-process user agent is not asked to do anything beyond what
+          this specification already requires. What matters is where the
+          requirement is evaluated. Such a user agent is encouraged to evaluate
+          the [=transient activation=] check against a record of input held by
+          the process that presents the chooser, rather than one the process
+          running the page can write to, and to record the consumption that
+          follows it in that same trusted process. Having done so, it rejects a
+          request that fails the check, as the [=transient activation=] check
```
mohamedamir: Should we document the specific exception?
```suggestion
          request that fails the check with a {{"NotAllowedError"}}
{{DOMException}}, as the [=transient activation=] check
```

mohamedamir (approved):
LGTM

Thank you!

index.html
mohamedamir: Thanks Marcos for updating the PR.
LGTM now!

marcoscaceres:
@wseltzer, @hlflanagan, editors think this is ready to merge if the WG is ok with it.

index.html
```diff
@@ -2138,6 +2149,55 @@ <h3>
...
+          Forged Requests from a Compromised Content Process
+        </h3>
+        <p>
+          Some user agents run a page's code in one process and handle
+          keyboard, mouse, and touch input in another, more trusted, process.
+          In those user agents, both the [=transient activation=] check and the
+          step that [=Consume user activation|consumes that activation=] are
+          evaluated in the process that runs the page, which is the process an
+          attacker is most likely to have taken over. An attacker in that
+          position can skip both and ask for the [=digital credential chooser=]
```
TallTed, line 2166: It's unclear to me what "both" means. I suggest a parenthetical to be specific, a la —
```suggestion
          position can skip both (A and B) and ask for the [=digital credential chooser=]
```

TallTed: Maybe this is a good guess?
```suggestion
          position can skip both (transient activation and its consuming 
          step) and ask for the [=digital credential chooser=]
```

index.html
```diff
@@ -2138,6 +2149,55 @@ <h3>
...
+          anyway.
+        </p>
+        <p>
+          [=Transient activation=] lasts only a few seconds so that the user
+          can perceive the link between interacting with a page and the page
+          calling an interface gated on that interaction. [=Forged requests=]
+          break exactly that link. What the attacker gains is a credential
+          prompt drawn by the user agent or by the operating system, at a
```
TallTed, line 2174:
```suggestion
          prompt drawn by the user agent or the operating system, at a
```

index.html
```diff
@@ -2138,6 +2149,55 @@ <h3>
...
+          moment of the attacker's choosing, which the user cannot tell apart
+          from a legitimate one. That makes it useful for phishing, and for the
```
TallTed, line 2176:
```suggestion
          from a legitimate prompt. That makes it useful for phishing, and for the
```

hlflanagan:
Discussed [16 September 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-09-16-DCAPI-B-notes.md#add-security-considerations-on-forged-credential-chooser-requests--506). Will bring to the next Series A call for further discussion.

PR #509, opened 2026-05-07
Remove "origin bound" declaration from DigitalCredential

### LINKED ISSUE #508: Remove "origin bound" declaration from DigitalCredential
## Problem

The spec declares that `DigitalCredential` instances are [origin bound](https://www.w3.org/TR/credential-management-1/#credential-origin-bound), meaning they have an `[[origin]]` internal slot. However, this declaration is inert because:

1. **`[[origin]]` is never set during object construction.** The algorithm that creates a `DigitalCredential` (in the "initiate the credential request" steps) only sets `data` and `protocol`.

2. **`[[CollectFromCredentialStore]]` is not implemented.** DC credentials are always discovered from an external source (`[[discovery]]` is `"remote"`). The `[[origin]]` slot is only useful as a same-origin filter in `[[CollectFromCredentialStore]]`, which DC does not override.

3. **`[[Store]]` throws `NotSupportedError`.** DC delegates to `Credential`'s default `[[Store]]`, which throws `NotSupportedError`. Credentials are never persisted in the credential store.

4. **DC credentials are ephemeral.** They're created, resolved to a promise, consumed by the RP, then garbage collected. There's no retrieval path where `[[origin]]` would be checked.

## Security model

The actual origin-binding for Digital Credentials happens at the **protocol layer**, not the Credential Management layer:

- The wallet binds the response to the requesting origin (e.g., OpenID4VP `client_id` = origin)
- Permissions Policy controls which frames can call the API
- Transient activation + user consent via the chooser UI prevent unauthorized requests
- `sameOriginWithAncestors` prevents iframe abuse

The `[[origin]]` internal slot adds no security on top of these mechanisms.

## Proposal

Remove the paragraph:

> `DigitalCredential` instances are origin bound.

This aligns with `PublicKeyCredential` (WebAuthn), which similarly does NOT declare itself "origin bound" because its origin-scoping happens through its own mechanisms (rpId + CollectedClientData).

If DC ever gains real credential store semantics in the future, that would be the appropriate time to add origin-binding along with the `[[CollectFromCredentialStore]]` logic that makes it meaningful.

## Related

- w3c/webappsec-credential-management#295 (making `[[origin]]` universal but nullable on `Credential`)

marcoscaceres:
Closes #508

The `[[origin]]` internal slot provides no security benefit for `DigitalCredential` because:

- DC credentials are **ephemeral** (never stored in the credential store)
- `[[CollectFromCredentialStore]]` is not implemented (`[[discovery]]` is `"remote"`)
- `[[Store]]` delegates to the default implementation, which throws `NotSupportedError` (credentials are never persisted)
- The construction algorithm never sets `[[origin]]` on the returned object

The actual origin-binding for DC happens at the **protocol layer** (wallet binds response to the requesting origin) and via Permissions Policy, transient activation, and user consent.

This aligns with `PublicKeyCredential` (WebAuthn), which also does not declare itself "origin bound" for similar reasons.

See also: w3c/webappsec-credential-management#295 (making `[[origin]]` universal but nullable on `Credential`).

- [ ] Modified Web platform tests (N/A - no behavioral change)

Implementation commitment:

- [ ] WebKit (N/A - no behavioral change)
- [ ] Chromium (N/A - no behavioral change)
- [ ] Gecko (N/A - no behavioral change)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/509.html" title="Last updated on May 12, 2026, 10:18 PM UTC (2cbc3ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/509/8dc4712...2cbc3ab.html" title="Last updated on May 12, 2026, 10:18 PM UTC (2cbc3ab)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1462,9 +1462,6 @@
       static boolean userAgentAllowsProtocol(DOMString protocol);
     };
     </pre>
-    <p>
-      {{DigitalCredential}} instances are [=Credential/origin bound=].
-    </p>
     <h4>
       The `protocol` member
     </h4>


mohamedamir (approved):
LGTM

Thank you!

timcappalli (approved):

marcoscaceres:
Margin as this has no observable effect... was kinda accidentally added, but it's not used anywhere and we confirmed across UAs that one implements it.

Issue #369, opened 2025-09-01
Add details about the Cross-device experience for the Digital Credentials API


mohamedamir:
The current Digital Credentials Web API specification only briefly mentions cross-device functionality, stating it "supports implementing cross-device experience with proximity checks." This is too vague for developers.

This issue proposes adding a new section to the specification that details the cross-device experience, specifically:

- How a user's **smartphone can act as an authenticator for a second device**, like a laptop or desktop computer.

- That this flow leverages the **Client to Authenticator Protocol (CTAP)**, which is the same protocol used in FIDO2.

- The **step-by-step authentication flow**, including how proximity checks and biometric authorization on the mobile device are used to securely authorize the transaction.

- The corresponding **Security Considerations**: the importance of secure, short-range communication channels (like Bluetooth Low Energy or NFC) for the proximity check, as well as the use of cryptographic keys to secure the communication between the devices.

timcappalli:
I think this is better suited for developer docs than the spec (e.g. digitalcredentials.dev).

We could just remove:
> "supports implementing cross-device experience with proximity checks."

marcoscaceres:
Yeah, @timcappalli might be right... however, we could still make some non-normative mention of this in the spec's security considerations section as it's a really common case, right? 

People reading the spec might rightfully ask "how does the cross-device stuff work securely?", to which we could say something like .. "Although cross-device protocol issuance and request protocols/formats are out-of-scope, data exchange is often handled by standards like [CTAP]. Please see [FIDO Security Reference](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-security-ref-v2.0-id-20180227.html) for how it works securely." or something.

marcoscaceres:
See also https://github.com/w3c-fedid/digital-credentials/pull/515/changes#r3549946247

hlflanagan:
Discussed [13 July 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-07-13-DCAPI-A-notes.md#add-details-about-the-cross-device-experience-for-the-digital-credentials-api-369-and-add-a-signal-of-whether-presentation-occurred-locally-or-cross-device-527). OK to close when #538 is merged.

PR #515, opened 2026-05-13
Classify validation failures into specific exception types

### LINKED ISSUE #472: User agent request validation and errors

marcoscaceres:
Closes #472.

Replaces "throw an appropriate exception" in `Validate credential requests` with a switch that maps each class of validation failure to a concrete exception type, matching the style of the platform-error switch in `Initiate the credential request`:

- User agent does not permit the request (local policy, configuration, or the user's choices) → `"NotAllowedError"` `DOMException`, deliberately indistinguishable from user cancellation so a site cannot tell whether the user or the user agent refused, nor infer anything about the user's configuration or credentials
- Request not permitted for security reasons unrelated to the user (context or transport) → `"SecurityError"` `DOMException`
- Malformed/invalid request data → `TypeError`
- Otherwise → `"OperationError"` `DOMException`

The first two cover Tim's point in #472 about distinguishing a user-agent-restricted request from an actual protocol problem.

Also keeps a short note on the additional validation criteria a user agent might apply (local policy, configuration, or the user’s choices, such as rejecting a request that seeks particular credential attributes), and separates what counts as a validation failure (protocol-specific) from how such a failure is classified into an exception type (in-scope). The cryptographic algorithm and certificate examples are gone per review; broken cryptography is recorded as an out-of-scope threat instead, which also fills the empty list there.

Without this, implementations have no shared rule and WPT can only assert "some exception."

cc @w3c-fedid/dc-editors

- [ ] Modified Web platform tests (follow-up — WPT tests exercising this classification are in flight on the WPT side)

Implementation commitment:

- [x] WebKit (follow-up): https://bugs.webkit.org/show_bug.cgi?id=317300
- [ ] Chromium (follow-up)
- [ ] Gecko (follow-up)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/515.html" title="Last updated on Aug 20, 2026, 12:52 AM UTC (957494c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/515/3cdd740...957494c.html" title="Last updated on Aug 20, 2026, 12:52 AM UTC (957494c)">Diff</a>


mohamedamir (approved):
@marcoscaceres as per our discussion, LGTM

index.html
```diff
@@ -994,10 +994,28 @@ <h3>
...
             </aside>
-            <aside class="issue" data-number="472"></aside>
             <ol>
-              <li>If validation fails, [=exception/throw=] an appropriate
-              [=exception=].
+              <li>If validation fails:
+                <ol>
+                  <li>Let |error| be determined as follows:
+                    <dl class="switch">
+                      <dt>
```
marcoscaceres, line 1035: Related to #369 

Do we need, "Request rejected by because it was unsigned" or is it just an operation error? Otherwise, is it SecurityError?

index.html
```diff
@@ -976,10 +976,24 @@ <h3>
...
           data=] or [=digital credential/issuance request data=] according to
           |request|'s [=digital credential/presentation protocol=] or [=digital
-          credential/issuance protocol=] or other criteria. Validation
-          requirements are protocol-specific and are outside the scope of this
-          specification:
-            <aside class="note" title="Validation details outside scope">
+          credential/issuance protocol=] or other criteria.
+            <p>
+              In addition to protocol-defined requirements, a [=user agent=]
+              might apply additional validation criteria based on local policy,
```
TallTed: redundant
```suggestion
              might apply validation criteria based on local policy,
```

index.html
```diff
@@ -994,10 +1008,50 @@ <h3>
...
+              <li>If validation fails:
+                <ol>
+                  <li>Let |error| be determined as follows:
+                    <dl class="switch">
+                      <dt>
+                        The [=user agent=] does not permit the request based on
+                        local policy, configuration, or the user's choices:
+                      </dt>
+                      <dd>
+                        A newly created {{"NotAllowedError"}} {{DOMException}}.
```
MasterKale: Is there something significant about use of the verbiage, "a newly created"? Does this simply mean "a new instance of the `NotAllowedError` DOMException"?

hlflanagan:
Discussed on [27 July 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-07-27-DCAPI-A-notes.md#classify-validation-failures-into-specific-exception-types515) call.

index.html
marcoscaceres: It’s following what WebIDL suggests, but we should link [=created=]. https://webidl.spec.whatwg.org/#dfn-create-exception

marcoscaceres: @copilot can you handle linking [=created=] as per above?

index.html
```diff
@@ -977,10 +977,24 @@ <h3>
...
-            <aside class="note" title="Validation details outside scope">
+          credential/issuance protocol=] or other criteria.
+            <p>
+              In addition to protocol-defined requirements, a [=user agent=]
+              might apply validation criteria based on local policy,
+              configuration, or evolving security considerations. For example, a
+              [=user agent=] might reject a request that (a) seeks particular
+              credential attributes, (b) uses or requires cryptographic
+              algorithms the [=user agent=] is configured not to accept (e.g.,
+              as part of algorithm agility or a transition to post-quantum
```
timcappalli: There should be no case where a browser's crypto support affects a protocol request. I think this should be removed.

index.html
```diff
@@ -977,10 +977,24 @@ <h3>
...
+              schemes), or (c) relies on certificates or trust anchors that are
+              not accepted by the [=user agent=]'s configured trust decisions.
```
timcappalli: I have some concerns about this statement. There are no protocols supported by this spec which bring the browser actively into the cryptographic trust architecture. This should be removed.

timcappalli (changes requested):

index.html
marcoscaceres: So, it's not so much the browser's crypto, but also the OS... when the request goes into the browser, in WebKit, we hand it to the OS to check. That can fail immediately in the sandboxed content process (which is running inside the "user agent"). Does that make sense? I can change to platform tho.

timcappalli: Even then, there should be zero dependency on any cryptographic support at that layer. All crypto operations are handled by the Credential Manager. I think this text should be removed.

marcoscaceres: Ok, let me drop it and clarify some things in the security considerations. Will send an update soon.

marcoscaceres: Removed, platform variant included.

Broken cryptography is now recorded as an out-of-scope threat instead, which also fills that empty list.

On the premise though, since it will come up again: you are right that no cryptographic operations happen in the user agent, and I checked, there is no trust evaluation anywhere in WebKit's DC path. But there is already a dependency on cryptographic structure. Chromium parses the payload of a signed OpenID4VP request without verifying it, to decide whether to show an interstitial. WebKit parses the mdoc deviceRequest and encryptionInfo and builds certificate chain objects out of the reader authentication certificates. And #520 has the user agent turn down a request whose response_mode is not dc_api.jwt. So the algorithm is visible to a user agent today without it doing any crypto, which is what a follow-up issue will rest on.

index.html
marcoscaceres: Removed.

timcappalli (approved):

index.html
MasterKale: I don't think such an exception is needed so long as we continue to have an unsigned protocol listed in the [Table of supported presentation and issuance protocols](https://www.w3.org/TR/digital-credentials/#dfn-table-of-supported-presentation-and-issuance-protocols). User agents should support whatever protocols are in that list. Until someone gets the WG to reach consensus that DC API should not allow unsigned requests, and so long as that table continues to include unsigned OID4VP, then why throw an error when a verifier chooses to use it?

MasterKale (approved):
This is probably the best we can do in terms of exceptions that DC API itself can raise. The fact that the majority of errors that occur once the request is made are communicated back via a protocol-specific manner limits opportunities to get more descriptive.

I'm mostly happy that we're not making everything a `NotAllowedError`.

hlflanagan:
Discussed again on [10 August 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-08-10-DCAPI-A-notes.md#classify-validation-failures-into-specific-exception-types-515) call. OK to merge.

index.html
marcoscaceres: Leaving in for completeness. We can sharpen this once we get more implementation experience.

PR #520, opened 2026-05-18
Specify protocol validation + enforce encryption

### LINKED ISSUE #519: Spec doesn’t enforce encrypted response requirement
I was re-reading the Digital Credentials API's design, privacy, and security considerations, and noticed what looks like a possible mismatch with `openid4vp-v1-unsigned`.

The spec says, in §11.1:

> The Digital Credentials API offers the user agent the ability to intermediate on behalf of the user ... It also enforces certain minimum requirements on supported protocols, such as response encryption.

And in §11.3.1.7:

> To prevent exposure of user information to other parties in "transit", for example browser extensions loaded on verifier pages, and to encourage secure storage of user credentials by the verifier, protocols are required to support and mandate encrypted responses in a credential exchange.

However, OpenID4VP over the Digital Credentials API appears to allow `openid4vp-v1-unsigned` requests to use an unencrypted response.

In particular, OpenID4VP says:

> The value of the `response_mode` parameter MUST be `dc_api` when the response is not encrypted and `dc_api.jwt` when the response is encrypted as defined in Section 8.3.

It also gives a non-normative example of an unsigned OpenID4VP request using:

```js
{
  response_type: "vp_token",
  response_mode: "dc_api",
  nonce: "n-0S6_WzA2Mj",
  client_metadata: {...},
  dcql_query: {...}
}
```

My reading is that `unsigned` only describes the request, not the response. So `openid4vp-v1-unsigned` is not inherently a problem.

The problem is that `openid4vp-v1-unsigned` appears to be usable with `response_mode: "dc_api"`, and OpenID4VP defines that as the unencrypted response mode.

If that reading is correct, then `openid4vp-v1-unsigned` as currently referenced by the Digital Credentials API seems incompatible with the Digital Credentials API's stated requirement that supported protocols mandate encrypted responses.

Questions:

1. Is it correct that `openid4vp-v1-unsigned` can be used with `response_mode: "dc_api"` and therefore produce an unencrypted response?
2. If so, should the Digital Credentials API require OpenID4VP requests to use `response_mode: "dc_api.jwt"`?
3. Should the Digital Credentials API reject `openid4vp-v1-unsigned` requests that specify `response_mode: "dc_api"`?
4. Should the Digital Credentials API define protocol-specific request validation hooks so that a user agent can reject requests that do not meet the API's baseline privacy/security requirements?
5. Should the OpenID4VP DC API profile also make this requirement explicit, perhaps by saying that when OpenID4VP is used through the W3C Digital Credentials API, `response_mode` MUST be `dc_api.jwt`?

This may not require dropping `openid4vp-v1-unsigned`. An unsigned request with an encrypted response may still be acceptable. The concern is specifically allowing `response_mode: "dc_api"` through the Digital Credentials API.

marcoscaceres:
Closes #519 

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/520.html" title="Last updated on Sep 7, 2026, 11:43 PM UTC (ff448d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/520/a3b699e...ff448d9.html" title="Last updated on Sep 7, 2026, 11:43 PM UTC (ff448d9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -690,6 +690,125 @@
         </tr>
       </tbody>
     </table><!--
+    // MARK: Convert request protocol
+    -->
+    <h3>
+      Convert request protocol
+    </h3>
+    <p>
+      To <dfn>convert request protocol</dfn> given a
+      {{DigitalCredentialGetRequest}} or {{DigitalCredentialCreateRequest}}
+      |request|:
+    </p>
+    <ol class="algorithm">
+      <li>If |request| is:
+        <dl class="switch">
+          <dt>
+            A {{DigitalCredentialGetRequest}}:
+          </dt>
+          <dd>
+            <ol>
+              <li>Let |protocolString| be |request|'s
+              {{DigitalCredentialGetRequest/protocol}}.
+              </li>
+              <li>If |protocolString| does not equal any [=enumeration value=]
+              in {{DigitalCredentialPresentationProtocol}}, return failure.
+              </li>
+              <li>Return the {{DigitalCredentialPresentationProtocol}}
+              [=enumeration value=] whose value is |protocolString|.
+              </li>
+            </ol>
+          </dd>
+          <dt>
+            A {{DigitalCredentialCreateRequest}}:
+          </dt>
+          <dd>
+            <ol>
+              <li>Let |protocolString| be |request|'s
+              {{DigitalCredentialCreateRequest/protocol}}.
+              </li>
+              <li>If |protocolString| does not equal any [=enumeration value=]
+              in {{DigitalCredentialIssuanceProtocol}}, return failure.
+              </li>
+              <li>Return the {{DigitalCredentialIssuanceProtocol}}
+              [=enumeration value=] whose value is |protocolString|.
+              </li>
+            </ol>
+          </dd>
+        </dl>
+      </li>
+    </ol><!--
+    // MARK: Validate ISO18013-7 Annex C request data
+    -->
+    <h3>
+      Validate ISO18013-7 Annex C request data
+    </h3>
+    <p>
+      To <dfn>validate ISO18013-7 Annex C request data</dfn> given a JavaScript
+      value |data|:
+    </p>
+    <ol class="algorithm">
+      <li>If |data| is not an [=object=], return failure.
+      </li>
+      <li>If |data| does not have an own property "`deviceRequest`", return
+      failure.
+      </li>
+      <li>If |data| does not have an own property "`encryptionInfo`", return
+      failure.
+      </li>
+      <li>If |data|["`deviceRequest`"] is not a [=string=], return failure.
+      </li>
+      <li>If |data|["`encryptionInfo`"] is not a [=string=], return failure.
+      </li>
+      <li>Let |deviceRequest| be the result of parsing
+      |data|["`deviceRequest`"] as a `Base64DeviceRequest` value according to
+      [[[ISO18013-7]]] Annex C.
+      </li>
+      <li>If |deviceRequest| is failure, return failure.
+      </li>
+      <li>Let |encryptionInfo| be the result of parsing
+      |data|["`encryptionInfo`"] as a `Base64EncryptionInfo` value according to
+      [[[ISO18013-7]]] Annex C.
+      </li>
+      <li>If |encryptionInfo| is failure, return failure.
+      </li>
+      <li>If |encryptionInfo| does not identify the encryption protocol
+      "`dcapi`", return failure.
+      </li>
+      <li>If |encryptionInfo| does not contain encryption parameters with a
+      nonce and recipient public key, return failure.
+      </li>
+      <li>Return success.
+      </li>
+    </ol><!--
+      // MARK: Validate OpenID4VP request data
+    -->
+    <h3>
+      Validate OpenID4VP request data
+    </h3>
+    <p>
+      To <dfn>validate OpenID4VP request data</dfn> given a [=digital
+      credential/protocol identifier=] |protocol| and a JavaScript value
+      |data|:
+    </p>
+    <ol class="algorithm">
+      <li>If |data| is not an [=object=], return failure.
+      </li>
+      <li>Let |authorizationRequest| be the result of processing |data| as an
+      OpenID4VP Authorization Request according to |protocol|'s request type.
+      </li>
+      <li>If |authorizationRequest| is failure, return failure.
+      </li>
+      <li>If |authorizationRequest|'s `response_mode` value is not
+      "`dc_api.jwt`", return failure.
+      </li>
+      <li>If |authorizationRequest| does not contain the parameters necessary
+      for the [=holder=] to encrypt the Authorization Response according to
+      [[[OPENID4VP]]], return failure.
+      </li>
+      <li>Return success.
+      </li>
+    </ol><!--
     // MARK: Credential Request Coordinator
     -->
     <h2>
@@ -907,19 +1026,13 @@
       </li>
       <li>[=List/For each=] |request| of |requests|:
         <ol>
-          <li>Let |protocol| be |request|'s
-          {{DigitalCredentialGetRequest/protocol}}, if |request| is a
-          {{DigitalCredentialGetRequest}}, or |request|'s
-          {{DigitalCredentialCreateRequest/protocol}}, if |request| is a
-          {{DigitalCredentialCreateRequest}}.
-          </li>
-          <li>If |request| is a {{DigitalCredentialGetRequest}} and |protocol|
-          does not equal any [=enumeration value=] in
-          {{DigitalCredentialPresentationProtocol}}, [=iteration/continue=].
-          </li>
-          <li>If |request| is a {{DigitalCredentialCreateRequest}} and
-          |protocol| does not equal any [=enumeration value=] in
-          {{DigitalCredentialIssuanceProtocol}}, [=iteration/continue=].
+          <li>[=Convert request protocol=] given |request|:
+            <ol>
+              <li>If conversion results in failure, [=iteration/continue=].
+              </li>
+              <li>Otherwise, let |protocol| be the result of the conversion.
+              </li>
+            </ol>
           </li>
           <li>If [=user agent allows protocol=] given |protocol| returns
           `false`, [=iteration/continue=].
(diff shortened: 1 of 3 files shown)


marcoscaceres:
@copilot, can you move the "Convert request protocol" algorithm and where it gets used into its own PR?

index.html
```diff
@@ -935,31 +1036,10 @@ <h3>
...
-                credential/presentation protocol=] or [=digital
-                credential/issuance protocol=] for details, including potential
-                reasons for validation failure, and any security and privacy
-                considerations that need to be considered by implementers
-                during validation.
-              </p>
-            </aside>
-            <aside class="issue" data-number="472"></aside>
+          <li>[=credential request coordinator/Validate request data=] given
+          |protocol| and |data|:
```
mohamedamir, line 1040: We don't pass over the request here, but we assert the request type

hlflanagan:
Discussed on the [10 June 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-06-10-DCAPI-B-notes.md#spec-doesnt-enforce-encrypted-response-requirement-issue519--specify-protocol-validation--enforce-encryption-pr520) call. Chairs will be taking this to a formal vote.

index.html
```diff
@@ -1095,21 +1095,96 @@ <h3>
...
+        {{DigitalCredentialPresentationProtocol/"openid4vp-v1-multisigned"}} are
+        distinct [=digital credential/presentation protocols=], this is enforced
+        for each individually (i.e., per operating mode). For OpenID4VP the
+        `response_mode` is "`dc_api.jwt`"; for the ISO mdoc protocol,
+        `encryptionInfo` is present.
+      </p>
+    </aside>
+    <aside class="note" title="Issuance responses">
+      <p>
+        A [=digital credential/issuance response=] does not return credential
```
TallTed, line 1119:
```suggestion
        An [=digital credential/issuance response=] does not return credential
```

wseltzer:
Note the WG poll deciding to mandate response encryption: https://www.w3.org/wbs/154550/fediddcapienc/results/

marcoscaceres:
@copilot resolve the merge conflicts in this pull request

index.html
```diff
@@ -737,6 +737,107 @@ <h3>
...
+      </li>
+      <li>If |data|["`deviceRequest`"] is not a [=string=], return failure.
+      </li>
+      <li>If |data|["`encryptionInfo`"] is not a [=string=], return failure.
+      </li>
+      <li>Let |deviceRequest| be the result of parsing
+      |data|["`deviceRequest`"] as a `Base64DeviceRequest` value according to
+      [[[ISO18013-7]]] Annex C.
+      </li>
+      <li>If |deviceRequest| is failure, return failure.
```
mohamedamir: Under the principle that we want the **bare minimum** request validations to enforce response encryption:

Parsing `deviceRequest` as a `Base64DeviceRequest` according to Annex C is unnecessary. `deviceRequest` carries the requested credential attributes intended for the wallet/holder and has no impact on whether the response is encrypted. Requiring user agents to parse it forces browser engines to embed a full CBOR and ISO 18013-5/7 parser, which re-introduces the heavy request parsing burden that the WG previously decided against.

To enforce response encryption, validating `encryptionInfo` (steps 8–11) is what is needed. Steps 6 and 7 should be removed.

index.html
```diff
@@ -737,6 +737,107 @@ <h3>
...
+    <p>
+      To <dfn>validate OpenID4VP request data</dfn> given a [=digital
+      credential/protocol identifier=] |protocol| and a JavaScript value
+      |data|:
+    </p>
+    <ol class="algorithm">
+      <li>If |data| is not an [=object=], return failure.
+      </li>
+      <li>Let |authorizationRequest| be the result of processing |data| as an
+      OpenID4VP Authorization Request according to |protocol|'s request type.
```
mohamedamir, line 807: Under the bare minimum request validation rule:

1. "processing `|data|` as an OpenID4VP Authorization Request according to `|protocol|`'s request type" implies full authorization request parsing and cryptographic signature verification for `openid4vp-v1-signed` and `openid4vp-v1-multisigned`. Browsers do not hold trust anchors for verifiers and cannot verify those signatures (that is the wallet's responsibility).
2. Checking for encryption parameters (step 5) is unworkable synchronously because OpenID4VP allows encryption metadata/keys to be referenced via `client_metadata_uri` or `jwks_uri` (which require external network fetches that the browser cannot perform during `get()`).

The bare minimum check required to guarantee response encryption in OpenID4VP is verifying that `response_mode` is `"dc_api.jwt"` (which mandates JWE Compact Serialization under OpenID4VP §8.3). The wallet/holder is responsible for validating keys/client metadata and rejecting if missing.

index.html
```diff
@@ -737,6 +737,107 @@ <h3>
...
+      for the [=holder=] to encrypt the Authorization Response according to
+      [[[OPENID4VP]]], return failure.
+      </li>
+      <li>Return success.
+      </li>
+    </ol><!--
+    // MARK: Validate OpenID4VCI request data
+    -->
+    <h3>
+      Validate OpenID4VCI request data
```
mohamedamir: OpenID4VCI is an issuance protocol (`create()`). As noted in lines 1178–1184 and lines 1236–1239, issuance responses do not return credential data to the calling website, and response encryption is strictly a concern for presentation protocols (`get()`).

Additionally, OpenID4VCI integration with the Digital Credentials API is still pending standardization (`[[[OPENID4VCI]]] § Coming Soon`).

If the goal of this PR is to enforce response encryption with the bare minimum request validations, OpenID4VCI request validation is out of scope and should be removed from this PR.

index.html
```diff
@@ -1031,6 +1111,148 @@ <h3>
...
+    -->
+    <h3>
+      Validate request data
+    </h3>
+    <p>
+      To <dfn data-dfn-for="credential request coordinator">validate request
+      data</dfn> given a {{DigitalCredentialProtocol}}
+      |protocol:DigitalCredentialProtocol|, a JavaScript value
+      |data|, and a {{DigitalCredentialGetRequest}} or
+      {{DigitalCredentialCreateRequest}} |request|:
```
```suggestion
      data</dfn> given a {{DigitalCredentialProtocol}}
      |protocol:DigitalCredentialProtocol| and a JavaScript value |data|:
```
`request` was only passed here to pass it down to `validate OpenID4VCI request data` to assert its type. If OpenID4VCI request validation is removed (or since `data` is already extracted in `validate credential requests`), `request` is redundant and can be dropped from this algorithm signature and its caller.

index.html
```diff
@@ -1031,6 +1111,148 @@ <h3>
...
+          </dt>
+          <dd>
+            Return the result of [=validate OpenID4VP request data=] given
+            |protocol| and |data|.
+          </dd>
+          <dt>
+            {{DigitalCredentialIssuanceProtocol/"openid4vci-v1"}}:
+          </dt>
+          <dd>
+            Return the result of [=validate OpenID4VCI request data=] given
```
mohamedamir: If OpenID4VCI request validation is removed as out of scope for response encryption, this branch can be dropped as well.

mohamedamir (commented):
Thanks for putting this together, Marcos! The move to enforce response encryption is a great step forward for security and privacy.

To make sure this lands smoothly and keeps user agent complexity manageable, I suggest we scope request validation strictly to the **bare minimum** checks required to guarantee that the response is encrypted, rather than introducing full request payload parsing into browser engines. Left a few inline comments with suggestions.

index.html
simoneonofri: Step 2 still says to process the request according to the request type, without defining the exact minimal decoding and member-extraction steps or their failure conditions. 

Could the algorithm state, for each of the unsigned, signed, and multisigned protocol identifiers, where response_mode is read from and how decoding or JSON failures return failure?

index.html
```diff
@@ -1048,95 +1128,226 @@ <h3>
...
+      </p>
+    </aside>
+    <p>
+      A [=digital credential/presentation response=] is always encrypted for
+      the requesting party. [=credential request coordinator/Validate request
+      data=] establishes this, per [=digital credential/presentation
+      protocol=], by rejecting a request that does not compel an encrypted
+      response. [=credential request coordinator/Validate response data=] then
+      rejects a response that is plainly unencrypted, which catches a
+      [=holder=] that ignored the request rather than one that set out to leak.
```
simoneonofri, line 1262: This can be excluded (at least in OpenID4VP) if the response is an error. Do we always want to have encryption in this case as well?

index.html
```diff
@@ -774,37 +774,22 @@ <h3>
...
-    </ol><!--
+    </ol>
+    <aside class="note" title="Why only encryptionInfo">
+      <p>
+        [[[ISO18013-7]]] Annex C requires "`encryptionInfo`" in every request
+        and defines only one response structure, an encrypted one. A request
+        carrying "`encryptionInfo`" therefore compels an encrypted [=digital
+        credential/presentation response=], which is all this algorithm
+        establishes. Whether the rest of the request conforms to Annex C is for
+        the [=holder=] to decide, not the [=user agent=].
```
TallTed, line 790: So each user, each holder, must purchase the ISO spec? For how else are they to decide whether a request conforms to Annex C? This makes no sense to me!

index.html
```diff
@@ -826,41 +811,32 @@ <h3>
...
       </li>
     </ol>
-    <aside class="issue" title="OpenID4VCI validation requirements">
+    <aside class="note" title="Reading response_mode from a signed request">
+      <p>
+        For {{DigitalCredentialPresentationProtocol/"openid4vp-v1-unsigned"}},
+        `response_mode` is a member of |data| itself. For
+        {{DigitalCredentialPresentationProtocol/"openid4vp-v1-signed"}} and
+        {{DigitalCredentialPresentationProtocol/"openid4vp-v1-multisigned"}} it
+        sits inside the signed payload, so the [=user agent=] decodes that
```
TallTed, line 823:
```suggestion
        {{DigitalCredentialPresentationProtocol/"openid4vp-v1-multisigned"}},
        it sits inside the signed payload, so the [=user agent=] decodes that
```

index.html
marcoscaceres: The ISO spec is free now. You don't need to purchase it. You just need to download it.

Issue #382, opened 2025-09-28
Wallet Selection Binding during Issuance


leecam:
There is an ask from government Issuers to provide a strong signal of which wallet the user selected during a .create() call.

The attack they are looking to mitigate assumes a user with a trusted browser and OS, but has installed a malicious wallet application. If the user selects this malicious wallet application it may able to forward the request to a wallet on an attacker controlled phone, where the digital credential would get provisioned. The Issuer can't easily detect this attack today. This is issue to look at API that maybe required to address this issue.

The attack would go as follows:
1) User has a phone in a good state. Neither the browser or OS is compromised
2) But they have installed a malicious wallet app
3) The issuer website/app makes a DC create call offering to issue a credential. This request contains the pre-auth code.
4) Malicious wallet is selected by the user and it gets the request.
5) It forwards it to an attacker controlled phone (which could be rooted) where the request is injected into a legitimate wallet app, e.g Google Wallet etc.. This instance of the wallet app is full attacker controlled.
6) The legitimate wallet creates the credential. All the attestations and such would be valid
7) The Digital Credential is now available in a legitimate and issuer-approved wallet, but on the attacker controlled phone.

There are a number of ways we could solve this:
1) The DC Create call could return the wallet that was selected by the user.
2) The Issuer could provide an allow-list of acceptable wallet package name when making the create call
3) The Issuer could provide a binding key when making the create call. Here the platform would pass the wallet hmac(binding_key, selected_wallet_identifier)
4) Something where the Issuer could provide an allow-list of wallet attestations using some new to-be-standardized PKI

Of these option 3 or 4 are my preferred options.

Option 1 is awkward as the credential issuance flow is normally complete before the call returns. It also exposes the user's wallet app in every call and this has poor privacy properties. Doesn't really work cross-device.

Option 2 exposes OS internals to the web, requiring a web API to contain OS specific app package names. Also just don't much like sending allow lists of package names, as you could imagine an issuer allowing any certified wallet, without knowing the package names upfront. It could lead to a lot of fragmentation. If this is signed over it could work for hybrid though.

[...]

PR #546, opened 2026-06-25
Add DigitalCredential.issue() for credential issuance

### LINKED ISSUE #356: Add DigitalCredentialIssuanceResponse partial
_Originally posted by @mohamedamir in https://github.com/w3c-fedid/digital-credentials/issues/343#issuecomment-3200827143_ 

> I find the most confusing part is the return of `DigitalCredential` interface, while the members make sense `protocol` and `data`, the fact that a `DigitalCredential` is returned is confusing.
Maybe we should create another interface e.g. `DigitalCredentialIssuanceResponse` that also inherits `Credential`?
This is only slightly better though because if the `Credential` inheritance.

marcoscaceres:
Closes #356

Moves digital credential issuance off `navigator.credentials.create()` onto a dedicated static `DigitalCredential.issue()` that resolves with a new `DigitalCredentialIssuanceResponse` instead of a `Credential` (the credential is issued into the holder's wallet, so the page only gets a protocol response). Removes the `create()` issuance path, renames the request/options types, and renames `digital-credentials-create` to `digital-credentials-issue`; presentation via `get()` is unchanged.

- [ ] Modified Web platform tests (link) — follow-up PR; the `create*` issuance tests and the `interfaces/digital-credentials.idl` mirror migrate to `issue()` / `digital-credentials-issue`.

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/546.html" title="Last updated on Jun 25, 2026, 11:59 PM UTC (e6f468d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/546/8ff9184...e6f468d.html" title="Last updated on Jun 25, 2026, 11:59 PM UTC (e6f468d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1577,9 +1577,12 @@
       </li>
       <li>Let |document| be |global|'s [=associated `Document`=].
       </li>
+      <li>If |document| is not [=Document/fully active=], return [=a promise
+      rejected with=] a {{"InvalidStateError"}} {{DOMException}}.
+      </li>
       <li>If |document| is not a [=Document/fully active descendant of a
       top-level traversable with user attention=], return [=a promise rejected
-      with=] a {{"InvalidStateError"}} {{DOMException}}.
+      with=] a {{"NotAllowedError"}} {{DOMException}}.
       </li>
       <li>If |document| is not [=allowed to use=] the
       <a>"digital-credentials-issue"</a> [=policy-controlled feature=], return
--- a/index.html
+++ b/index.html
@@ -332,10 +332,9 @@
       <p>
         The following example shows how to request the issuance of a digital
         credential using the [[[#DC-API|Digital Credentials API]]]. To issue a
-        digital credential, a site calls the
-        `navigator.credentials.`{{CredentialsContainer/create()}} method,
-        which, if the user agent supports issuance, would initiate the issuance
-        flow:
+        digital credential, a site calls the {{DigitalCredential/issue()}}
+        static method, which, if the user agent supports issuance, would
+        initiate the issuance flow:
       </p>
       <pre class="example html" title=
       "Requesting issuance of a digital credential">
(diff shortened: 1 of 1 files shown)


timcappalli:
I don't agree with this change.

It makes things ergonomically weird for web and app developers seemingly for the sake of it feeling ergonomically better for browser developers.

This also breaks cross-device issuance with PXP.

marcoscaceres:
I’m not following. Can you please explain how it does either of those things?

timcappalli:
Users of the Digital Credentials API also typically use the Web Authentication API and Federated Credential Management APIs. Having different ergonomics for the same operation, for the purpose of browser engine purity does not make sense, especially when `.create` is already being used.

PXP (CTAP 2.3 hybrid) expects the `get` or `create` JSON serialized output from CredMan.

c2bo:
This feels like a very late change to the API imho - is this really necessary?

Whatever we do, we probably want to have a bit more text on the expectation how issuance works - right now DC API for Issuance is basically just an invocation mechanism where the real protocol steps do not flow through DC API and I don't believe that is captured in the text right now.

hlflanagan:
Discussed on [29 June 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-06-29-DCAPI-A-notes.md) call. Further discussion will be in the related issue #382

marcoscaceres:
Reading the minutes, concluded the current solution is fine even though the semantics are a bit wonky.

PR #550, opened 2026-07-08
Issuance request forwarding: Documenting the threat

simoneonofri:
This PR adds the issuance request forwarding threat to the Security Considerations threat model.

The threat captures the case discussed in #382 where a malicious holder's credential manager, such as a digital wallet, is selected by the user during issuance and forwards the issuance request data to another credential manager on an attacker-controlled device.

Addressing this threat remains under discussion in #382.

Related to #382.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/550.html" title="Last updated on Jul 8, 2026, 8:44 PM UTC (953517e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/550/1b5612a...953517e.html" title="Last updated on Jul 8, 2026, 8:44 PM UTC (953517e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1862,6 +1862,17 @@
           explicit permission from the embedding site, potentially enabling
           credential harvesting or unauthorized access to sensitive user data.
         </dd>
+        <dt>
+          <strong>Issuance Request Forwarding</strong>
+        </dt>
+        <dd>
+          A malicious [=holder's=] [=credential manager=], such as a digital
+          wallet, selected by the user for a [=digital credential/issuance
+          request=], forwards the [=digital credential/issuance request data=]
+          to another [=credential manager=] on a device under attacker control,
+          causing issuance to complete with that other credential manager
+          rather than with the credential manager selected by the user.
+        </dd>
       </dl>
       <h4>
         Out of Scope Threats
--- a/index.html
+++ b/index.html
@@ -2003,7 +2003,7 @@
           carefully crafted or malformed protocol requests through the API.
         </dd>
         <dt>
-          <strong>Issuance Request Forwarding</strong>
+          <dfn>Issuance Request Forwarding</dfn>
         </dt>
         <dd>
           A malicious [=holder's=] [=credential manager=], such as a digital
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1862,6 +1862,17 @@ <h4>
...
+        <dt>
+          <strong>Issuance Request Forwarding</strong>
+        </dt>
+        <dd>
+          A malicious [=holder's=] [=credential manager=], such as a digital
+          wallet, selected by the user for a [=digital credential/issuance
+          request=], forwards the [=digital credential/issuance request data=]
+          to another [=credential manager=] on a device under attacker control,
+          causing issuance to complete with that other credential manager
+          rather than with the credential manager selected by the user.
```
c2bo, line 1874: We should probably add a note that authorization steps in the protocol might mitigate this? So the result doesn't have to be the issuance to complete.

Something like this?

```suggestion
          causing issuance to complete with that other credential manager
          rather than with the credential manager selected by the user. Note
          that the success of such an attack also depends on possible
          authorization steps performed by the underlying
           [=digital credential/issuance protocols=].
```

marcoscaceres: The "the success of such an attack" reads a little weird. 

Projecting forward to the actual mitigation text, what mitigations are in place, for instance, in Openid4vci? 

With an example of a mitigation, we might be able to reframe the note a bit.

index.html
```diff
@@ -1862,6 +1862,17 @@ <h4>
           explicit permission from the embedding site, potentially enabling
           credential harvesting or unauthorized access to sensitive user data.
         </dd>
+        <dt>
+          <strong>Issuance Request Forwarding</strong>
+        </dt>
+        <dd>
+          A malicious [=holder's=] [=credential manager=], such as a digital
```
TallTed, line 1869: Is it the holder that is malicious, or is it the credential manager? A little rephrasing should be done to make this clear.

PR #562, opened 2026-07-20
Private Browsing mode

### LINKED ISSUE #416: Define DC API behavior in Private Browsing Mode
During the TAG review of the Digital Credentials (DC) API, it was noted that the current specification does not explicitly address the behavior of the API when the User Agent (UA) is in "Private" or "Incognito" browsing mode.

The exact feedback is

_"The document does not address the behaviour of the proposed solution in private browsing mode.As per [web platforms design principles (section 2.9)](https://www.w3.org/TR/design-principles/#do-not-expose-use-of-private-browsing-mode), the UA should not exhibit different behaviour in this mode. However, there are specific use cases, such as proof of age, where certain disclosures are mandated by regulations even when the user is browsing in private mode. This could inadvertently reveal unnecessary information to entities other than the UA (e.g., the wallet, website, or verifier) about whether the user is browsing in private or normal mode."_

We need to discuss and document the behavior for the DC API in private browsing mode.

timcappalli:
Closes #416 

~~[ ] Modified Web platform tests (link)~~

Implementation commitment:

~~[ ] WebKit (link to issue)~~
~~[ ] Chromium (link to issue)~~
~~[ ] Gecko (link to issue)~~

Documentation and checks

- [x] Affects privacy

~~[ ] Affects security~~
~~[ ] Pinged MDN~~
~~[ ] Updated Explainer~~
~~[ ] Updated digitalcredentials.dev~~

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/562.html" title="Last updated on Jul 20, 2026, 8:45 PM UTC (635a8e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/562/518f026...635a8e1.html" title="Last updated on Jul 20, 2026, 8:45 PM UTC (635a8e1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2814,6 +2814,18 @@
           we expect [=user agents=] and this specification to leverage it to
           improve the user permission experience as well.
         </p>
+        <h4>
+          Private Browsing Mode
+        </h4>
+        <p>
+          The Digital Credentials API interacts with components outside of the
+          user agent, such as the OS platform and [=credential managers=]. User
+          agents should present the user with a clear and understandable
+          explanation that information from their private browsing session is
+          being shared with these components, and that the user should not
+          expect that their private browsing session is fully isolated from
+          these components.
+        </p>
       </section>
     </section>
     <section>


index.html
```diff
@@ -2814,6 +2814,18 @@ <h4>
           we expect [=user agents=] and this specification to leverage it to
           improve the user permission experience as well.
         </p>
+        <h4>
+          Private Browsing Mode
+        </h4>
+        <p>
+          The Digital Credentials API interacts with components outside of the
+          user agent, such as the OS platform and [=credential managers=]. User
+          agents should present the user with a clear and understandable
```
marcoscaceres, line 2823: This is good advice, but it applies generally and not specifically to private browsing. 

Neither Chrome nor Safari do anything different in Private browsing that we wouldn’t do in regular browsing, so this doesn’t really make sense as “private browsing” advice. 

What’s more useful is conveying what is being shared with who, particularly when it involves cross-site requests (irrespective of private browsing mode). In such cases, credential managers (third-party wallets) cannot be fully trusted to convey the relationship and the UA should intervene.

PR #569, opened 2026-07-29
Add security considerations on signing presentation requests

### LINKED ISSUE #568: Unsigned requests can defeat the required response encryption
Even though we now require response encryption (#520), that protection is not super helpful when the request is made unsigned. The key used to encrypt the response is sent inside the request. When the request is unsigned, nothing stops it from being changed on the way to the wallet. A bad script on the page, say one injected by a malicious browser extension, can swap in its own key, and can just as easily widen what is being asked for, turning a simple age check into a demand for full identity details. The wallet then encrypts the answer to the attacker instead of the site, so the attacker can read everything the user just shared: their name, birth date, photo, document number.

If the request were signed, a wallet would notice it had been changed and reject the request. With an unsigned request there is nothing to check, so neither the wallet nor the site can tell it happened.

Our threat model calls this out. It lists Request Tampering as an in-scope threat, and the only defense we offer is that the API is exposed in secure contexts only, which "reduc[es] the risk of tampering through insecure contexts (e.g., a malicious script being injected through the network)". That covers network attackers, which TLS already handles. However, TLS does not help here, because this attacker is already inside the page. A signature would be the second line of defense. Unsigned requests do not have one. (See also #539, which deals with the same kind of attacker.)

OpenID4VC HAIP is explicit that unsigned requests lean on exactly this:

> Note that unsigned requests depend on the origin information provided by the platform and the web PKI for request integrity protection and to authenticate the Verifier. Signed requests introduce a separate layer for request integrity protection and Verifier authentication that can be validated by the Wallet.

That holds up against someone on the network. It does not hold up against a script inside the page, which has a perfectly good origin and perfectly good web PKI. So the protection unsigned requests rely on is the one thing this attacker already has.

This is not just theoretical. The EU already bans the unsigned mode for its digital wallet. ETSI TS 119 472-2, which EU law (Commission Implementing Regulation (EU) 2026/1731, http://data.europa.eu/eli/reg_impl/2026/1731/oj) makes mandatory, says:

> OIDFVP-HAIP-API-03: The EUDI Wallet shall not support Authorization Requests with exchange protocol value set to "openid4vp-1-unsigned".

One caution. If we fix this by requiring signed requests, it may mean only pre-approved sites can ask for credentials. The [TAG warns](https://www.w3.org/2001/tag/doc/prevent-credential-abuse/) that this can put too much power in the hands of a few issuing authorities. So whatever we do, it should stay open: lots of credential providers, and no issuing authority getting special treatment.

marcoscaceres:
Closes #568. 

Adds a short security considerations subsection strongly encouraging relying parties to sign their presentation requests, explains the in-page tampering it protects against, and is explicit that a signature the wallet cannot check against a recognized key does not help. Resolves the concern raised in #568 without mandating signed requests.

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks:

- [ ] Affects privacy
- [x] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/569.html" title="Last updated on Aug 21, 2026, 12:49 AM UTC (1aa2fdb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/569/02bd8ca...1aa2fdb.html" title="Last updated on Aug 21, 2026, 12:49 AM UTC (1aa2fdb)">Diff</a>


timcappalli (approved):

index.html
```diff
@@ -1958,6 +1958,43 @@ <h3>
...
+          extension. That script can change what is being asked for, turning a
+          narrow request into a demand for much more, and it can replace the
+          parameters used to encrypt the response so that the response is
+          encrypted to the attacker rather than to the verifier. Requiring the
+          API to be used in a [=secure context=] does not prevent this, because
+          the attacker is already inside one. A signed request gives the
+          [=credential manager=] something to check, so the change can be
+          detected instead of passing unnoticed. With an unsigned request, the
+          [=credential manager=] has to trust that the verifier has kept its own
+          page free of such scripts.
```
TallTed: Better to close a paragraph with the desired action and positive effect.
```suggestion
          the attacker is already inside one. With an unsigned request, the
          [=credential manager=] has to trust that the verifier has kept its own
          page free of such scripts. A signed request, on the other hand, gives
          the [=credential manager=] something to check, so the change can be
          detected instead of passing unnoticed. 
```

index.html
```diff
@@ -1958,6 +1958,43 @@ <h3>
...
+        <p>
+          Signing only helps if the [=credential manager=] can tell that the
+          signature belongs to the verifier making the request. A script inside
+          the page can re-sign an altered request with a key of its own, so a
+          signature that cannot be checked against a signing key already
+          recognized as the verifier's offers no protection against this
+          attacker. Which signing keys to accept, and how they are established,
+          is decided by the ecosystem a credential belongs to rather than by
+          this specification, and the protection a signed request actually
+          offers depends on that decision.
```
```suggestion
          Even at that, signing only helps if the [=credential manager=] can tell
          that the signature belongs to the verifier making the request. A script
          inside the page can re-sign an altered request with a key of its own,
          so a signature offers no protection against this attacker unless it can
          be checked against a signing key that is already recognized as the
          verifier's. Which signing keys to accept, and how they are established,
          is decided by the ecosystem a credential belongs to rather than by this
          specification, and the protection a signed request actually offers
          depends on that decision.
```

index.html
```diff
@@ -1958,6 +1958,43 @@ <h3>
...
+        -->
+        <h3>
+          Signing presentation requests
+        </h3>
+        <p>
+          Where a [=digital credential/presentation protocol=] offers a way to
+          sign a request, a [=verifier=] is strongly encouraged to use it.
+        </p>
+        <p>
+          An unsigned request can be altered by a script running in the
```
mohamedamir: Should we update the request tempering sub-section [here](https://www.w3.org/TR/digital-credentials/#in-scope-threats) too?

index.html
```diff
@@ -1958,6 +1958,43 @@ <h3>
...
+          verifier's own page, such as one injected by a malicious browser
+          extension. That script can change what is being asked for, turning a
+          narrow request into a demand for much more, and it can replace the
+          parameters used to encrypt the response so that the response is
```
```suggestion
          parameters used to [[[#encrypting-credential-responses|encrypt the response]]]  so that the response is
```

index.html
```diff
@@ -1958,6 +1958,43 @@ <h3>
...
+      <section>
+        <!--
+        // MARK: Signing presentation requests
+        -->
+        <h3>
+          Signing presentation requests
+        </h3>
+        <p>
+          Where a [=digital credential/presentation protocol=] offers a way to
+          sign a request, a [=verifier=] is strongly encouraged to use it.
```
mohamedamir: our of curiosity, why not SHOULD instead of "is strongly encouraged"?

index.html
mohamedamir: What do you think about cross-referencing this with the cross-device section, which also advocated for request sign-ing.
I do know, I feel both are related, but also don't belong to the same section.

I don't feel strongly though!

```suggestion
          offers depends on that decision. Beyond protecting against same-device
         in-page tampering, request signing also provides critical
         defense-in-depth when presenting credentials across multiple devices
         (see [[[#cross-device-security-and-proximity]]]).
```

mohamedamir (approved):
Looks good.
I have added minor suggestions!

Thanks Marcos!

index.html
marcoscaceres: The parent `<section>` is currently marked as informative, so no RFC2119 keywords take affect here. 

We should fix that eventually and make this section normative.

index.html
marcoscaceres: I don't mind. Otherwise, followup?

index.html
marcoscaceres: Makes sense... do you want to send this ☝️ again?

index.html
mohamedamir: Gotcha!

index.html
mohamedamir: A follow-up SGTM as well!

index.html
```diff
@@ -2072,6 +2072,43 @@ <h3>
...
+        <p>
+          Even at that, signing only helps if the [=credential manager=] can tell
+          that the signature belongs to the verifier making the request. A script
+          inside the page can re-sign an altered request with a key of its own,
+          so a signature offers no protection against this attacker unless it can
+          be checked against a signing key that is already recognized as the
+          verifier's. Which signing keys to accept, and how they are established,
+          is decided by the ecosystem a credential belongs to rather than by this
+          specification, and the protection a signed request actually offers
+          depends on that decision.
```
```suggestion
          depends on that decision. Beyond protecting against same-device
         in-page tampering, request signing also provides critical
         defense-in-depth when presenting credentials across multiple devices
         (see [[[#cross-device-security-and-proximity]]]).
```

index.html
mohamedamir: Done!

PR #539, opened 2026-06-16
Add verifier hardening guidance for script injection

marcoscaceres:
Addresses #133.

This adds an informative subsection, "Verifier hardening against script injection," to the Security Considerations section. It is relying-party-facing guidance. Because the API is exposed in any secure context, a script injected into a verifier's page (through cross-site scripting, a compromised dependency, or a tampered script) can read credential responses the verifier has decrypted and reflected back to client-side JavaScript, and can invoke the API from the verifier's origin to make its own requests. The guidance encourages verifiers to harden their pages with a strong Content Security Policy, complemented by Trusted Types and Subresource Integrity, as defense-in-depth.

It also makes explicit that mandated response encryption does not address this case, because the response is in cleartext once it reaches the page, and that the credential chooser limits silent misuse but does not stop an injected script from reading data the verifier has already exposed. These points reflect the discussion in the issue thread (encryption not protecting reflected data and the risk of a fresh same-origin request, and Subresource Integrity for the supply-chain angle).

The text is non-normative and uses no RFC 2119 keywords, since it is guidance for relying parties rather than a user-agent conformance requirement. It is filed as Addresses, not Closes, because the issue's open question, whether to recommend or to require injection mitigation, is left for the working group; this PR adds the recommendation, not a requirement.

The following tasks have been completed:

- [ ] Modified Web platform tests (link) (not applicable: non-normative guidance, no new user-agent behavior to test)

Implementation commitment:

- [ ] WebKit (link to issue) (not applicable: non-normative)
- [ ] Chromium (link to issue) (not applicable: non-normative)
- [ ] Gecko (link to issue) (not applicable: non-normative)

Documentation and checks

- [ ] Affects privacy
- [x] Affects security
- [ ] Pinged MDN (not applicable: no API surface change)
- [ ] Updated Explainer (not applicable)
- [ ] Updated digitalcredentials.dev (not applicable)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/539.html" title="Last updated on Aug 25, 2026, 12:31 PM UTC (7a0bd14)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/539/2e9b687...7a0bd14.html" title="Last updated on Aug 25, 2026, 12:31 PM UTC (7a0bd14)">Diff</a>

PR #512, opened 2026-05-09
Add document origin to request context

marcoscaceres:
Fixes a spec bug where Digital Credentials requests from cross-origin iframes use the wrong origin. Credential Management already passes the calling document's origin to `[[DiscoverFromExternalSource]]`, but the spec was re-deriving the top-level origin instead. This threads `documentOrigin` through `prepare credential requests` and `initiate the credential request`, storing it in the `request context` struct as `document origin` (replacing `top-level origin`).

Closes #333
Closes #504
Closes #332

- [ ] Modified Web platform tests (link)

Implementation commitment:

- [ ] [WebKit](https://github.com/WebKit/WebKit/pull/64596)
- [ ] Chromium (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks:

- [ ] Affects privacy
- [ ] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer
- [ ] Updated digitalcredentials.dev

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/512.html" title="Last updated on Aug 20, 2026, 1:40 PM UTC (2b71d78)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/512/02bd8ca...2b71d78.html" title="Last updated on Aug 20, 2026, 1:40 PM UTC (2b71d78)">Diff</a>

PR #572, opened 2026-08-12
chore: add CI check for committed CRLF line endings

marcoscaceres:
GitHub's server-side commit paths (the batch "Apply suggestions from code review" button, and the web file editor) can commit CRLF, and `.gitattributes` does not prevent it because it is not applied to those commits; the result is a whole-file diff that hides the real change, as recently happened on #454, #539 and #512.

This adds a check that fails a pull request when any committed file has CRLF or mixed line endings, and prints the one-command fix (`git add --renormalize .`).

### DIFF
--- a/.github/workflows/line-endings.yml
+++ b/.github/workflows/line-endings.yml
@@ -6,6 +6,7 @@ name: Line Endings
 # small edit into a whole-file diff, so catch it here instead.
 
 on:
+  workflow_dispatch: {}
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


TallTed:
@marcoscaceres — Note that the "one-command fix" is only available to command-line git users. That does not include me, at least. I do 99+% of my GitHub work through the default web interface. I have *a very few times* made use of the `.` launched VScode interface to rename and/or move files and/or directories. That's it. I'm sure there are other contributors here and elsewhere who will have no idea how to use that "one-command fix". So I think this is still a partial solution/workaround.

marcoscaceres:
@TallTed, that’s fine. This just catches the strays earlier. 

I’ll see if there’s a way to just always normalize if the bad line endings are detected.

marcoscaceres:
@TallTed, ah, so the semi-automated fix is, the checker hits the error, then there's a "Tell Copilot to fix it" button. That's probably good enough for now.

PR #575, opened 2026-08-19
Apply friction when a request carries no signature to check

### LINKED ISSUE #571: Should we warn the user when a request is not properly protected?
An unsigned request lets a script on the page swap in its own encryption key. The wallet then encrypts the user's identity documents to the attacker, and neither the wallet nor the site can tell, because there is no signature to check. `openid4vp-v1-unsigned` allows this today (#568).

Browsers already warn about self signed certificates for the same reason: the maths is fine, but nothing vouches for it.

Should a user agent do the same here? Something like:

> When a user agent determines that a request's protection cannot be relied upon, it SHOULD apply additional friction and strongly discourage the user from continuing.

If the user declines, it fails as `NotAllowedError` like any other refusal, so nothing new is exposed and sites need no new handling.

marcoscaceres:
Closes #571

Adds a step to the initiate the credential request algorithm: when a request carries no signature a credential manager can verify, the user agent applies additional friction as it shows the chooser and steers the user away from continuing. It sits immediately before the chooser is displayed because that is the moment the user decides.

- [x] ~~Modified Web platform tests~~ — user-facing friction is not web-observable

Implementation commitment:

- [ ] WebKit (link to issue)
- [ ] Gecko (link to issue)

Documentation and checks

- [ ] Affects privacy
- [x] Affects security
- [ ] Pinged MDN
- [ ] Updated Explainer

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/575.html" title="Last updated on Aug 21, 2026, 2:11 AM UTC (a4cc0f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/575/2e9b687...a4cc0f9.html" title="Last updated on Aug 21, 2026, 2:11 AM UTC (a4cc0f9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1114,6 +1114,21 @@
       context/requests=] is |validatedRequests| and [=request context/top-level
       origin=] is |topLevelOrigin|.
       </li>
+      <li>If a [=digital credential/credential request=] in |validatedRequests|
+      carries no signature that a [=credential manager=] can verify, the [=user
+      agent=] SHOULD apply additional friction when displaying the [=digital
+      credential chooser=], and strongly discourage the user from continuing.
+        <p class="note" title="Why friction here">
+          An unsigned request can be altered by a script running in the
+          [=verifier=]'s own page, including the parameters used to encrypt the
+          response, and neither the [=credential manager=] nor the [=verifier=]
+          can detect that it happened. Requests are transmitted unencrypted so
+          that a [=user agent=] can inspect them for exactly this kind of risk
+          analysis. A user who declines is handled by the steps below, which
+          [=credential request coordinator/reject the credential request with=]
+          a {{"NotAllowedError"}} {{DOMException}}.
+        </p>
+      </li>
       <li>[=In parallel=]:
         <ol>
           <li>Display a [=digital credential chooser=] with |requestData| and


index.html
```diff
@@ -1136,6 +1114,21 @@ <h3>
       context/requests=] is |validatedRequests| and [=request context/top-level
       origin=] is |topLevelOrigin|.
       </li>
+      <li>If a [=digital credential/credential request=] in |validatedRequests|
+      carries no signature that a [=credential manager=] can verify, the [=user
```
MasterKale, line 1249: How would the user agent know that a signature couldn't be verified by a downstream CM if the user agent doesn't also perform the signature verification? Put another way, the only thing the user agent can really know about whether the request is signed is A) maybe the protocol identifier communicates that the request is unsigned (e.g. `"openid4vp-v1-unsigned"`) or B) the user agent attempts to perform the protocol-specific signature verification itself. 

**A** has the issue of, if a protocol identifier within this spec doesn't clearly indicate that the request is unsigned then the user agent can't have anything to key off of to show this friction. And **B** just won't work at any reasonable scale.

So I think right off the bat we need to be careful about how the user agent _knows_ a request is signed. And I'm not sure it can be based on any ability to know that a CM receiving the request could verify the signature.

PR #586, opened 2026-09-16
Platform-bound wallet attestation for issuance (#382)

### LINKED ISSUE #382: Wallet Selection Binding during Issuance

mohamedamir:
This PR addresses issue #382 by adding support for platform-bound wallet attestation during credential issuance.

When issuing credentials via `navigator.credentials.create()`, an issuer cannot currently verify which wallet processed the request, or whether an untrusted wallet intercepted the issuance offer and forwarded it to another device. By providing an `issuerAttestationNonce` in `DigitalCredentialCreationOptions`, the issuer supplies an opaque challenge that the user agent routes directly to the platform attestation subsystem rather than including it in the issuance request data delivered to the wallet application.

Fixes #382

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/pull/586.html" title="Last updated on Sep 16, 2026, 7:58 AM UTC (e6b2a7c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/digital-credentials/586/b31343b...e6b2a7c.html" title="Last updated on Sep 16, 2026, 7:58 AM UTC (e6b2a7c)">Diff</a>


npdoty:
The description says that this binds the issuance request to the user-selected wallet, but then later it seems that this also reveals to the issuer the platform-attested name of the wallet instance, so that the issuer knows exactly what software the user is running on their device, and can exclude specific wallets if it doesn't like that software. Is that right? That's both a privacy risk and a risk for coercion by the issuer, with the help of the platform, to ensure that the user only uses government-approved software.

index.html
```diff
@@ -77,6 +77,12 @@
           date: "2025-02-25",
           publisher: "W3C"
         },
+        "OpenID4VCI": {
+          title: "OpenID for Verifiable Credential Issuance 1.0",
+          href: "https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html",
+          authors: ["Torsten Lodderstedt", "Kristina Yasuda", "Tobias Looker"],
+          publisher: "OpenID Foundation"
+        },
```
marcoscaceres, line 85: This is specref already: https://www.specref.org/?q=OpenID4VCI ... so no need for: 

```suggestion
```

.cspell/names.txt
```diff
@@ -19,4 +19,8 @@ Thibodeau
 Yosef
 Jevon
 Authlete
-Heenan
\ No newline at end of file
+Heenan
+Lodderstedt
+Torsten
+Yasuda
+Kristina
```
marcoscaceres, line 26:
```suggestion
Heenan
```

hlflanagan:
Discussed [16 September 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-09-16-DCAPI-B-notes.md#platform-bound-wallet-attestation-for-issuance-382--586). Will continue the discussion on the next call.

index.html
```diff
@@ -1750,6 +1805,17 @@ <h4>
         An [=environment settings object=]'s [=environment settings
         object/origin=].
       </dd>
+      <dt>
+        <dfn data-dfn-for="request context">issuer attestation nonce</dfn>
+      </dt>
+      <dd>
+        A [=byte sequence=] supplied by an [=issuer=] to bind a [=digital
```
TallTed, line 1812:
```suggestion
        A [=byte sequence=] supplied by an [=issuer=] to bind an [=digital
```

index.html
```diff
@@ -2209,6 +2296,101 @@ <h3>
...
+          When issuing a credential with
+          {{CredentialsContainer/create()}}, an [=issuer=] cannot otherwise
+          tell which wallet processed its request, or on which device. A
+          malicious wallet can therefore forward the issuance offer to another
+          device, where a legitimate wallet completes it and produces a valid
+          platform attestation. Supplying an
+          {{DigitalCredentialCreationOptions/issuerAttestationNonce}} in
+          {{DigitalCredentialCreationOptions}} lets the issuer bind the
+          issuance to the wallet the user actually selected, which closes that
+          gap.
```
TallTed, line 2316:
```suggestion
          {{CredentialsContainer/create()}}, an [=issuer=] cannot tell which
          wallet processed its request, nor on which device. A malicious wallet
          can therefore forward the issuance offer to another device, where it
          is completed by a legitimate wallet that produces a valid platform
          attestation. By supplying an
          {{DigitalCredentialCreationOptions/issuerAttestationNonce}} in
          {{DigitalCredentialCreationOptions}}, the issuer can close that gap
          and bind the issuance to the wallet the user actually selected.
```

index.html
```diff
@@ -2209,6 +2296,101 @@ <h3>
...
+          to convey it instead to the platform attestation subsystem. Where
+          such a subsystem exists, it is expected to combine the nonce with the
+          identity of the selected wallet (such as its platform application
+          identifier), sign the result with a hardware-backed device
+          attestation key, and make the resulting integrity assertion available
+          to the wallet, which conveys it to the [=issuer=] over the [=digital
+          credential/issuance protocol=] (for example, via client attestation
+          headers in [[[OpenID4VCI]]]). None of those steps are performed by
+          the user agent, so none of them are normatively required by this
+          specification.
```
TallTed, line 2333:
```suggestion
          The security of this mechanism relies on the [=request context/issuer
          attestation nonce=] being withheld from the wallet. When [=credential
          request coordinator/initiate the credential request|initiating the
          credential request=], the [=user agent=] has to keep the nonce out of
          the [=digital credential/issuance request data=] it transmits and
          instead convey it to the platform attestation subsystem. Where such a
          subsystem exists, it is expected to combine the nonce with the
          identity of the selected wallet (such as its platform application
          identifier) and sign the result with a hardware-backed device
          attestation key. The subsystem then makes the resulting integrity
          assertion available to the wallet, which conveys it to the [=issuer=]
          over the [=digital credential/issuance protocol=] (for example, via
          client attestation headers in [[[OpenID4VCI]]]). None of those steps
          is performed by the user agent, so none of them is normatively
          required by this specification.
```

index.html
```diff
@@ -2209,6 +2296,101 @@ <h3>
...
+        </p>
+        <h4>
+          Cross-device transports and remote platform non-compliance
+        </h4>
+        <p>
+          The Digital Credentials API permits cross-device issuance flows (such
+          as scanning a QR code with a secondary device) while an
+          {{DigitalCredentialCreationOptions/issuerAttestationNonce}} is
+          present. Preserving the security guarantees of platform-bound wallet
+          attestation across a cross-device hop requires the receiving device
```
TallTed, line 2343:
```suggestion
          attestation over a cross-device hop requires the receiving device to
```

index.html
```diff
@@ -2209,6 +2296,101 @@ <h3>
...
+          to enforce operating-system-level isolation, receiving the nonce at
+          the platform layer and conveying it directly to the hardware
+          attestation subsystem without exposing it to the wallet application
+          process.
```
TallTed, line 2347:
```suggestion
          enforce operating-system-level isolation, receiving the nonce at the
          platform layer and conveying it directly to the hardware attestation
          subsystem without exposing it to the wallet application process.
```

hlflanagan:
Discussed [21 September 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-09-21-DCAPI-A-notes.md#platform-bound-wallet-attestation-for-issuance-382---586). Action item for @leecam and @timcappalli to draft the proposal.

